PR #6, opened 2016-03-15
Add refresh button to Editor

courte:
This works with NYPL-Simplified/circulation#105.

The refresh works - you can see it in the logs - but this should be a short-term interface given that the Metadata Wrangler takes some time to run and be absorbed into a work's metadata.


aslagle:
:+1:

PR #18, opened 2016-03-30
Better navigation

skomputer:
This branch changes a few navigation callbacks to make it compatible with changes to opds-browser's navigation, and gives the app's store an "editor" namespace so that in the future its store can be combined with opds-browser's store or other stores if needed.


aslagle:
Loading the start collection isn't working for me on this branch.

aslagle:
:+1:

PR #19, opened 2016-04-01
Browser tests

### LINKED ISSUE #8: Write tests for back/forward button navigation

skomputer:
This branch sets up Nightwatch.js to run a few basic navigation and routing tests (clicking links and tabs, refreshing, clicking the back button, etc) in Firefox and Chrome. Navigation and routing is probably the most complicated part of this application not covered at all by our Jest tests.

Closes #8.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@ lib/*
 dist/*
 node_modules/*
 typings/*
+tests/browser/globals.js
 *~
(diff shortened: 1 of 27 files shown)


aslagle:
:+1:

We'll need to figure out how we'll actually set up circulation so we can run these, but it's fine for now.

PR #24, opened 2016-04-07
added browser test for next/prev book hot keys

skomputer:
This branch adds a browser test verifying that when a book in a collection is displayed, left and right arrow keys navigate to the previous and next book in the collection, respectively.


aslagle:
:+1:

Issue #22, opened 2016-04-05
move interface declarations


skomputer:
For example, the EditorProps interface should be declared in the Editor component file.

PR #37, opened 2016-04-12
responsive header + interfaces refactor

### LINKED ISSUE #15: fix collapsed header styles
Using bootstrap components instead of just classes should suffice to make the collapsed header look decent.

skomputer:
This branch modifies Header to use react-bootstrap components instead of just bootstrap HTML classes, so that the header automatically collapses and expands with different viewport sizes. It also moves non-props interfaces out of the declaration file and into new file that requires them to be explicitly imported.

Closes #15 and #22.


src/components/BookDetailsContainer.tsx
```diff
@@ -1,5 +1,20 @@
 import * as React from "react";
 import TabContainer from "./TabContainer";
+import { BookDetailsContainerProps } from "opds-browser/lib/components/Root";
```
aslagle: How does it work that you're importing BookDetailsContainerProps and then also defining an interface with the same name?

src/components/Complaints.tsx
```diff
@@ -5,6 +5,21 @@ import ActionCreator from "../actions";
...
 import ComplaintForm from "./ComplaintForm";
 import ButtonForm from "./ButtonForm";
+import { BookData, PostComplaint } from "../interfaces";
+import { FetchErrorData } from "opds-browser/lib/interfaces";
+
+export interface ComplaintsProps {
+  bookUrl: string;
+  book: BookData;
+  complaints?: any;
+  fetchError?: FetchErrorData;
```
aslagle: maybe we should rename fetchError - it's confusing that fetchComplaints is a function but fetchError is an error object

PR #41, opened 2016-04-15
Removed publisher from edit form and added audience.

aslagle:
This connects to #29 
This connects to #33

### DIFF
--- a/src/components/__tests__/EditForm-test.tsx
+++ b/src/components/__tests__/EditForm-test.tsx
@@ -18,7 +18,7 @@ describe("EditableInput", () => {
         name="name"
         disabled={false}
         value="initial value"
-        />
+      />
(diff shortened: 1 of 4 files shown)


src/components/__tests__/EditForm-test.tsx
```diff
@@ -18,7 +18,7 @@ describe("EditableInput", () => {
         name="name"
         disabled={false}
         value="initial value"
-        />
+      />
```
skomputer: this is the only change i'm skeptical of

skomputer:
👍

src/components/__tests__/EditForm-test.tsx
aslagle: fixed

PR #42, opened 2016-04-20
React router

skomputer:
This branch makes a minimum number of changes to achieve compatibility with the latest OPDS Browser routing system (see NYPL-Simplified/opds-browser/pull/127).


aslagle:
top level in the header isn't working for me on this branch

PR #51, opened 2016-05-20
classifications form doesn't update state from new state-unrelated props

skomputer:
This branch fixes a bug that caused the ClassificationForm to display old input values immediately after the submit button was pressed.


aslagle:
:+1:

PR #60, opened 2016-06-08
Update packages

skomputer:
This branch updates all dependencies other than jest. For React 15 to work, we have to upgrade to the latest react-bootstrap, which uses completely new form components. Rather than use these new components, which are likely to change again before react-bootstrap is stable, I opted to write new EditableInput and EditabeRadio components that use bootstrap styles directly. Additionally, the updated Tabs component from react-boostrap was interfering with routing in the catalog, so I replaced it with basic elements that use bootstrap styles.


aslagle:
:+1:

PR #63, opened 2016-06-23
added CirculationEventsDownloadForm

skomputer:
Form visibility requires `showCircEventsDownload: true` config option.


aslagle:
:+1:

PR #70, opened 2016-09-06
Replaced jest with mocha and sinon.

aslagle:
This is the same thing I did in https://github.com/NYPL-Simplified/opds-web-client/pull/162


src/components/__tests__/CatalogHandler-test.tsx
```diff
@@ -14,7 +15,7 @@ describe("CatalogHandler", () => {
   let host = "http://example.com";
 
   beforeEach(() => {
-    document.location.href = host + "/test";
+    (jsdom as any).changeURL(window, host + "/test");
```
aslagle: I have a pr in DefinitelyTyped to fix this

PR #71, opened 2016-09-06
Moved inline styles to scss.

aslagle:


aslagle:
The browser tests need to be updated for some of the renamed classes.

PR #85, opened 2017-06-05
Reduced reducers

aslagle:
I had four tabs that fetch and edit data in very similar ways and I'm about to add another one, so I wrote a function to create a reducer for this pattern.


src/__tests__/actions-test.ts
```diff
@@ -427,9 +427,9 @@ describe("actions", () => {
 
       const data = await actions.fetchLibraries()(dispatch);
       expect(dispatch.callCount).to.equal(3);
-      expect(dispatch.args[0][0].type).to.equal(ActionCreator.LIBRARIES_REQUEST);
-      expect(dispatch.args[1][0].type).to.equal(ActionCreator.LIBRARIES_SUCCESS);
-      expect(dispatch.args[2][0].type).to.equal(ActionCreator.LIBRARIES_LOAD);
```
aslagle: I deleted these constants since they're no longer used outside of tests.

These action tests are also very repetitive and I can probably get rid of them later too.

src/components/LibraryEditForm.tsx
```diff
@@ -81,7 +81,7 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
             value={this.props.item && this.props.item.library_registry_shared_secret}
             />
         }
-        { this.props.data.settings && this.props.data.settings.map(setting =>
+        { this.props.data && this.props.data.settings && this.props.data.settings.map(setting =>
```
aslagle: This is an unrelated bug I found.

PR #90, opened 2017-06-22
Made interfaces for different configuration edit forms more similar.

aslagle:
I made some changes in the circulation manager so that the interfaces for the different edit form components could be more similar. In a future pull request, I hope to share more code between the components as well.


aslagle:
The one thing that's completely new is that a setting can have a list without options, and its value will be a list of strings. That's displayed as text inputs for each string, with a button to add another input.

PR #94, opened 2017-06-29
DRM service editing

aslagle:
This branch adds a DRM configuration tab, and removes the library registry short name and shared secret from the library edit form, since they're now go with the DRM integration.

I also made it so that a random secret can be generated client-side, if the setting has "randomizable" set to true. There is now a "Set to random value" button by the text field, instead of a separate checkbox for setting to a random value that hides the text field.

### DIFF
--- a/src/components/ProtocolFormField.tsx
+++ b/src/components/ProtocolFormField.tsx
@@ -144,7 +144,7 @@ export default class ProtocolFormField extends React.Component<ProtocolFormField
 
   randomize() {
     const element = (this.refs["element"] as any);
-    const chars = "1234567890abcdef";
+    const chars = "1234567890abcdefghijklmnopqrstuvwxyz!@#$%^&*()";
     let random = "";
     for (let i = 0; i < 32; i++) {
       random += chars.charAt(Math.floor(Math.random() * chars.length));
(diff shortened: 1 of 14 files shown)


src/components/ProtocolFormField.tsx
```diff
@@ -127,4 +141,14 @@ export default class ProtocolFormField extends React.Component<ProtocolFormField
     this.setState({ listItems: this.state.listItems.concat(listItem) });
     (this.refs["addListItem"] as any).clear();
   }
+
+  randomize() {
+    const element = (this.refs["element"] as any);
+    const chars = "1234567890abcdef";
```
leonardr: I don't think this will be a problem now that people won't be editing config files manually anymore, but earlier we had a problem caused by someone thinking their secret was a hexidecimal number, not a string. Ever since then I've generated random secrets from the full set of printable characters.

But like I say I don't think it matters anymore.

aslagle: I based this on what the server was doing before - `[random.choice('1234567890abcdef') for x in range(32)]`.

leonardr: That makes sense because nobody will have to type this in to anything.

aslagle: It's easy to change this list either way. Should we go with more just in case?

leonardr: Yeah, just add !@#$%^&*()

leonardr:
I don't know enough to :+1: this but we can go over it together if you want.

PR #95, opened 2017-06-29
Added CDN and search configuration tabs.

aslagle:
This is pretty straightforward. 

For CDNs, the only thing different than other tabs is CDNServices.label, which shows the CDN URL in the list of services.

For search, I added a way to make it so if there's already an integration you won't get a create link, since there can be only one search integration. This should maybe come from the server, but I think it's okay to be here for now.

### DIFF
--- a/src/components/CDNServices.tsx
+++ b/src/components/CDNServices.tsx
@@ -1,0 +1,42 @@
+import EditableConfigList from "./EditableConfigList";
+import { connect } from "react-redux";
+import ActionCreator from "../actions";
+import { CDNServicesData, CDNServiceData } from "../interfaces";
+import ServiceEditForm from "./ServiceEditForm";
+
+export class CDNServices extends EditableConfigList<CDNServicesData, CDNServiceData> {
+  EditForm = ServiceEditForm;
+  listDataKey = "cdn_services";
+  itemTypeName = "CDN service";
+  urlBase = "/admin/web/config/cdn/";
+  identifierKey = "id";
+  labelKey = "protocol";
+
+  label(item): string {
+    return item.settings && item.settings.url;
+  }
+}
+
+function mapStateToProps(state, ownProps) {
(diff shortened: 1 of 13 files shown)


src/components/CDNServices.tsx
```diff
@@ -0,0 +1,42 @@
...
+export class CDNServices extends EditableConfigList<CDNServicesData, CDNServiceData> {
+  EditForm = ServiceEditForm;
+  listDataKey = "cdn_services";
+  itemTypeName = "CDN service";
+  urlBase = "/admin/web/config/cdn/";
+  identifierKey = "id";
+  labelKey = "protocol";
+
+  label(item): string {
+    return item.settings && item.settings.url;
```
aslagle: This is what shows the CDN url in the list. I'm going to revisit these lists later and make all of them use the names better.

src/components/EditableConfigList.tsx
```diff
@@ -33,6 +33,7 @@ export abstract class EditableConfigList<T, U> extends React.Component<EditableC
   abstract urlBase: string;
   abstract identifierKey: string;
   abstract labelKey: string;
+  limitOne = false;
```
aslagle: This is what makes it possible to hide the create link if there's already an integration. It's true for search, false for everything else.

src/components/ConfigTabContainer.tsx
```diff
@@ -111,6 +129,8 @@ export default class ConfigTabContainer extends TabContainer<ConfigTabContainerP
       return "Sitewide Settings";
     } else if (name === "drm") {
       return "DRM";
+    } else if (name === "cdn") {
+      return "CDN";
     } else {
       return super.tabDisplayName(name);
     }
```
leonardr: Should 'search' also be in here?

aslagle: The default (`super.tabDisplayName(name)`) is to capitalize the first letter, which works for "search" but not "cdn".

leonardr:
:+1:

PR #96, opened 2017-07-05
Multi-library dropdown in header

aslagle:
This branch makes the admin interface work better when there are multiple libraries on the circulation manager.

The header has a dropdown with a list of the libraries in the system. When you select a different library, you're sent to that library's catalog page. When you're on a page like configuration that's not specific to a library, the library-related links don't show and the dropdown doesn't show a selected library, but you can select one to get back to the library's catalog.

This still needs work, but at least now you can get to the catalogs for all the libraries instead of being stuck on the default library.

### DIFF
--- a/src/components/CatalogHandler.tsx
+++ b/src/components/CatalogHandler.tsx
@@ -19,25 +19,34 @@ export interface CatalogHandlerProps extends React.Props<CatalogHandler> {
   };
 }
 
-export interface CatalogHandlerContext {
-  homeUrl: string;
-}
-
 export default class CatalogHandler extends React.Component<CatalogHandlerProps, any> {
-  context: CatalogHandlerContext;
-
-  static contextTypes: React.ValidationMap<CatalogHandlerContext> = {
-    homeUrl: React.PropTypes.string.isRequired
-  };
-
   static childContextTypes: React.ValidationMap<any> = {
-    tab: React.PropTypes.string
+    tab: React.PropTypes.string,
+    library: React.PropTypes.string
   };
 
   getChildContext() {
     return {
-      tab: this.props.params.tab
+      tab: this.props.params.tab,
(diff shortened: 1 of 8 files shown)


src/components/CatalogHandler.tsx
```diff
@@ -53,11 +62,16 @@ export default class CatalogHandler extends React.Component<CatalogHandlerProps,
   }
 
   render(): JSX.Element {
+    if (!this.getLibrary()) {
+      return (
+        <Header />
+      );
+    }
```
aslagle: If you haven't selected a library, you get the header but nothing else.

src/components/Header.tsx
```diff
@@ -58,4 +93,54 @@ export default class Header extends React.Component<HeaderProps, any> {
       </Navbar>
     );
   }
-}
\ No newline at end of file
+
+  componentWillMount() {
+    if (this.props.fetchLibraries) {
+      this.props.fetchLibraries();
```
aslagle: This fetches library info from the same endpoint that the config page for editing libraries uses.

src/components/Header.tsx
```diff
@@ -58,4 +93,54 @@ export default class Header extends React.Component<HeaderProps, any> {
...
+    }
+  }
+
+  changeLibrary() {
+    let library = (this.refs["library"] as any).value;
+    if (library) {
+      this.context.router.push("/admin/web/collection/" + library + "%2Fgroups");
```
aslagle: I decided to hard-code this instead of passing down a list of home urls for each library from the server.

aslagle: I used the router here because it was easier to test, but it doesn't quite work outside tests. I'm still looking into it.

aslagle: Fixed this in 5c115a4

PR #97, opened 2017-07-06
Parent dropdown for collections

aslagle:
This branch adds a parent dropdown to the service edit form. It's only used when the protocol has "child_settings", which only Overdrive does for now. It also only shows if there is at least one other Overdrive collection that could be the parent.

When the dropdown shows, it has "None" and all the other Overdrive collections in the system. If "None" is selected, the rest of the form will have the protocol's "settings", and if a parent is selected, the rest of the form will have the protocol's "child_settings".

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.0.30",
+  "version": "0.0.31",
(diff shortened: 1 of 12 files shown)


leonardr:
:+1:

PR #98, opened 2017-07-07
Fixed book url shortening for multi-library.

aslagle:
This fixes the shortened book urls now that the library short name is part of the url.

The long urls look like http://circulation/library/works/abcd, and they're shortened to /book/library/abcd. The old code assumed that /works was the beginning of the url.

This change won't work for default library urls where the short name isn't in the url, but for the admin interface it should be fine to always use the urls with the library short name.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.0.31",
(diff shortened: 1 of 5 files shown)


leonardr:
:+1:

PR #104, opened 2017-07-18
Edit library settings and discovery services

aslagle:
Most of this branch is an update to the section of the form for adding a library and its settings to an integration.

In the list of libraries, I added an edit button before the delete button. When you click the edit button, it shows/hides the settings for that library and you can change them without having to delete the library and add it again. I also changed the dropdown to add a new library so that it doesn't start out with a library selected, and the settings form fields don't show until you select one.

I also added discovery services to the Configuration section. I'll add something for registering libraries in a future pr. 
screen shot 2017-07-18 at 12 42 17 pm

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -1,0 +1,28 @@
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+
+export interface EditableProps {
+  disabled: boolean;
+  onEdit: () => void;
+}
+
+export default class Editable extends React.Component<EditableProps, void> {
+  render(): JSX.Element {
+    return (
+      <div className="editable">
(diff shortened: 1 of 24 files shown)


src/components/Editable.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+
+export interface EditableProps {
+  disabled: boolean;
+  onEdit: () => void;
+}
+
+export default class Editable extends React.Component<EditableProps, void> {
```
aslagle: "Editable" is a terrible name for this component, since there's already "EditableInput" and "EditableConfigList", which are unrelated.

PR #116, opened 2017-10-17
Custom list editing

aslagle:
This branch adds the custom list editor page, based on the design here: https://github.com/ricardoom/SE-Admin-UI

But without the advanced search capabilities yet.


aslagle:
I'll just comment that the state is getting complicated. I've been using the redux store for any request-related state, but keeping UI state in the react component states. However, in fcdcb9a I ended up with a piece of UI state that was needed by two components for displaying the form, not just for saving it. So maybe I should start thinking about moving more of the UI state to redux as well.

src/actions.ts
```diff
@@ -98,19 +102,29 @@ export default class ActionCreator extends BaseActionCreator {
...
           body: data,
           credentials: "same-origin"
         }).then(response => {
           if (response.status === 200 || response.status === 201) {
             dispatch(this.success(type));
+            if (response.text) {
+              response.text().then(text => {
+                dispatch(this.load<string>(type, text));
+              });
+            }
```
aslagle: This is for picking up an id that's sent back in response to a successful POST request.

src/components/CustomListEntriesEditor.tsx
```diff
@@ -0,0 +1,295 @@
+import * as React from "react";
+const dnd = require("react-beautiful-dnd");
+const DragDropContext = dnd.DragDropContext;
+const Droppable = dnd.Droppable;
+const Draggable = dnd.Draggable;
```
aslagle: I _should_ be able to import this like `import { DragDropContext, Droppable, Draggable } from "react-beautiful-dnd"`, but I couldn't get the typings to work and I gave up. It's said because this library has particularly good typings and I'm not using them.

src/index.tsx
```diff
@@ -3,6 +3,7 @@ import * as ReactDOM from "react-dom";
 import { Router, Route, browserHistory } from "react-router";
 import ContextProvider from "./components/ContextProvider";
 import CatalogHandler from "./components/CatalogHandler";
+import CustomListPage from "./components/CustomListPage";
```
aslagle: I need a convention for naming components that take up the whole page...

src/actions.ts
```diff
@@ -355,4 +369,20 @@ export default class ActionCreator extends BaseActionCreator {
...
+  }
+
+  editCustomList(library: string, data: FormData, csrfToken) {
+    const url = "/" + library + "/admin/custom_lists";
+    return this.postForm(ActionCreator.EDIT_CUSTOM_LIST, url, data, csrfToken).bind(this);
+  }
+
+  deleteCustomList(library: string, listId: string, csrfToken: string) {
+    const url = "/" + library + "/admin/custom_list/" + listId;
+    const data = new (window as any).FormData();
```
leonardr: Is this necessary to send an empty entity-body or is it left over from an experiment in sending an entity-body with a DELETE request?

src/components/CustomListEditor.tsx
```diff
@@ -0,0 +1,169 @@
...
+            found = true;
+            break;
+          }
+        }
+        if (!found) {
+          entriesChanged = true;
+          break;
+        }
+      }
+    }
```
leonardr: With the caveat that I'm not sure I'm reading this code correctly:

1. I think there's probably a more efficient way to run this check. It's a double-nested loop over lists that might have hundreds of items (as NYPL's staff picks list does). Just as a small example, short-circuiting if the two lists are different sizes should save time in a common case.
2. entriesChanged won't be set to true in the case where an item is present in `this.state.entries` but not present in `propEntry`. Is that a problem?

src/actions.ts
aslagle: It was left over from an experiment, but I do need to pass in some argument for the body. I made it null instead of an empty FormData.

src/components/__tests__/CustomListEntriesEditor-test.tsx
```diff
@@ -0,0 +1,466 @@
...
+    droppable = entriesContainer.find(Droppable);
+    let entries = droppable.find(Draggable);
+    expect(entries.length).to.equal(1);
+    expect(entries.at(0).text()).to.contain("entry B");
+    expect(onUpdate.callCount).to.equal(1);
+    const expectedEntries = [entriesData[1]];
+    expect(onUpdate.args[0][0]).to.deep.equal(expectedEntries);
+  });
+
+  it("adds a search result to the list", () => {
```
leonardr: Now is not the time to be second-guessing the UI decisions but... I will anyway. If I drag an item X off a list and into the search results, doesn't that make it look as if X is part of the search results, when it probably wasn't? Do people think of the "search results" section as more like a workspace than a consistent representation of what happened when a search was run?

src/components/CustomListEditor.tsx
aslagle: 1. I do short-circuit if the lists are different sizes - that's lines 112-113. You may be right though... Also, I wanted to use Array.includes instead of writing this at all but I wasn't happy with its browser compatibility status.
2. I think this problem is solved by short-circuiting if they're different sizes. If it gets here, they're the same size, and checking if every item in the props is also in the state should ensure that all the items match.

src/components/__tests__/CustomListEntriesEditor-test.tsx
aslagle: When you grab a list item to drag it, the search result box is replaced with a box that says "Drop items here to remove from the list". When you drop, it changes back to the search results. This idea was based on a flickr thing Mauricio showed me.

If the item you dropped is in the results for your current search, it will show up there again after you drop it. If it was from a different search, it will just disappear.

aslagle: And I just found a bug that that message doesn't go away if you drop the item back in the list when there are no search results.

src/components/CustomListEditor.tsx
```diff
@@ -0,0 +1,169 @@
...
+    this.setState({ name, entries: this.state.entries });
+  }
+
+  changeEntries(entries: CustomListEntryData[]) {
+    this.setState({ entries, name: this.state.name });
+  }
+
+  save() {
+    const data = new (window as any).FormData();
+    data.append("csrf_token", this.props.csrfToken);
```
aslagle: this shouldn't be here anymore

src/components/CustomListEditor.tsx
```diff
@@ -0,0 +1,169 @@
...
+    if (this.props.list) {
+      data.append("id", this.props.list.id);
+    }
+    let name = (this.refs["listName"] as TextWithEditMode).getText();
+    data.append("name", name);
+    let entries = (this.refs["listEntries"] as CustomListEntriesEditor).getEntries();
+    data.append("entries", JSON.stringify(entries));
+    this.props.editCustomList(data).then(() => {
```
aslagle: csrf token should also be here

src/components/CustomListEntriesEditor.tsx
```diff
@@ -0,0 +1,295 @@
...
+              )}
+              </Droppable>
+              { this.props.isFetchingMoreSearchResults &&
+                <button
+                  className="btn btn-default load-more-button"
+                  disabled={true}
+                  ><MoreDotsIcon />
+                </button>
+              }
+              { this.props.searchResults && !this.props.isFetchingMoreSearchResults &&
```
aslagle: this should only show if there's a next link

src/components/CustomLists.tsx
```diff
@@ -0,0 +1,264 @@
...
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.identifier && nextProps.identifier !== this.props.identifier) {
+      // We're opening the edit page for a list. Scroll back to the top.
+        const container = this.refs["container"] as HTMLDivElement;
+        if (container) {
+          container.scrollTop = 0;
+        }
```
aslagle: can probably get rid of this now

src/components/CustomLists.tsx
```diff
@@ -0,0 +1,264 @@
...
+    });
+  }
+
+  async editCustomList(data: FormData): Promise<void> {
+    await this.props.editCustomList(this.props.library, data, this.props.csrfToken);
+    this.props.fetchCustomLists(this.props.library);
+  }
+
+  async deleteCustomList(id: string): Promise<void> {
+    if (window.confirm("Are you sure you want to delete list " + id + "?")) {
```
aslagle: change to list name

src/components/CustomListEditor.tsx
aslagle: this one's okay because the parent component handles it.

aslagle:
@leonardr I think I've fixed all the things we talked about, except that part that compares the two lists.

PR #117, opened 2017-10-19
Moved all CSRF tokens to header instead of form data, and cleaned up types.

aslagle:
The goal of this branch was to make the CSRF token handling consistent across requests - all requests will use the "X-CSRF-Token" header instead of putting the token in form data.

The CSRF token now has to be an argument to all the actions, but it no longer needs to be passed down through as many of the components.

As part of this I also cleaned up some missing types that were previously `any`, mostly those used with Redux's `connect` function. This lets Typescript ensure I didn't mess up the arguments in all the `mapDispatchToProps` functions, which don't have their own tests and are now responsible for putting CSRF tokens in the action calls.

### DIFF
--- a/src/components/CirculationEventsDownloadForm.tsx
+++ b/src/components/CirculationEventsDownloadForm.tsx
@@ -7,7 +7,7 @@ export interface CirculationEventsDownloadFormProps extends React.Props<Circulat
   hide: () => void;
 }
 
-export default class CirculationEventsDownloadForm extends React.Component<CirculationEventsDownloadFormProps, any> {
+export default class CirculationEventsDownloadForm extends React.Component<CirculationEventsDownloadFormProps, void> {
   constructor(props) {
     super(props);
     this.download = this.download.bind(this);
--- a/src/components/ServiceEditForm.tsx
+++ b/src/components/ServiceEditForm.tsx
@@ -9,7 +9,6 @@ import { EditFormProps } from "./EditableConfigList";
 export interface ServiceEditFormProps<T> {
   data: T;
   item?: ServiceData;
-  csrfToken: string;
   disabled: boolean;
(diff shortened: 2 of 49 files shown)


aslagle:
I think I can improve this more by passing the CSRF token to the ActionCreator constructor instead of all the individual actions.

src/components/ServiceEditForm.tsx
```diff
@@ -418,7 +412,7 @@ export default class ServiceEditForm<T extends ServicesData> extends React.Compo
       (this.refs[setting.key] as any).clear();
     }
     const libraries = this.state.libraries.concat(newLibrary);
-    const newState = Object.assign({}, this.state, { libraries });
+    const newState = Object.assign({}, this.state, { libraries, selectedLibrary: null });
```
aslagle: I can't figure out how this was working before, or why my changes in this branch made it stop working.

src/components/CirculationEventsDownloadForm.tsx
```diff
@@ -7,7 +7,7 @@ export interface CirculationEventsDownloadFormProps extends React.Props<Circulat
   hide: () => void;
 }
 
-export default class CirculationEventsDownloadForm extends React.Component<CirculationEventsDownloadFormProps, any> {
+export default class CirculationEventsDownloadForm extends React.Component<CirculationEventsDownloadFormProps, void> {
```
leonardr: I don't understand the importance of this change.

aslagle: This is explicitly saying the component has no state instead of the state could be anything. Just type cleanup, doesn't change any functionality.

src/components/ServiceEditForm.tsx
leonardr: I am also mystified.

PR #118, opened 2017-10-20
Go to the edit page for a config item after creating it.

aslagle:
This branch brings some of the work I did for custom lists over to the config section. When you create a new config item, you end up on the edit page for the new item instead of back on the list of that type of item.

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -128,9 +128,11 @@ export default class ActionCreator extends BaseActionCreator {
             if (response.text) {
               response.text().then(text => {
                 dispatch(this.load<string>(type, text));
+                resolve(response);
               });
+            } else {
+              resolve(response);
             }
-            resolve(response);
           } else {
             response.json().then(data => {
               err = {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.0.46",
+  "version": "0.0.47",
   "description": "Web Front-end for Library Simplified Circulation Manager",
   "repository": {
     "type": "git",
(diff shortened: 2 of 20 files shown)


src/actions.ts
```diff
@@ -128,9 +128,11 @@ export default class ActionCreator extends BaseActionCreator {
             if (response.text) {
               response.text().then(text => {
                 dispatch(this.load<string>(type, text));
+                resolve(response);
               });
+            } else {
+              resolve(response);
             }
-            resolve(response);
```
aslagle: This was a bug in my previous implementation - it was possible for the response to resolve before the new identifier was loaded, in which case it would check whether to redirect too early and do nothing.

aslagle:
The code in postForm in actions.ts (https://github.com/NYPL-Simplified/circulation-web/pull/118/files#diff-162241b86351f640a8bb3127ed5fbd5bL129) extracts the entity-body, and dispatches a load action. The load action is processed by a reducer (https://github.com/NYPL-Simplified/circulation-web/blob/master/src/reducers/createFetchEditReducer.ts#L72) that sets it to `editedIdentifier` in the state. I create a separate reducer for each thing, and they're mapped to keys like `adminAuthServices` by combineReducers (https://github.com/NYPL-Simplified/circulation-web/blob/master/src/reducers/index.ts#L54).

I set all that up for custom lists so I didn't have to add it in this PR.

leonardr:
Got it, thanks for explaining in detail.

aslagle:
The test passing doesn't necessarily mean it works btw - it's all unit tests and a lot of stuff is mocked. We used to have some integration tests but I gave up when we lost the google account we were using for them.

PR #119, opened 2017-10-23
Delete config items

aslagle:
This pr makes it possible to delete things on the configuration pages, and changes the list layout a bit to accommodate that.

### DIFF
--- a/src/components/EditableConfigList.tsx
+++ b/src/components/EditableConfigList.tsx
@@ -163,7 +163,7 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
   }
 
   async delete(item: U): Promise<void> {
-    if (window.confirm("Are you sure you want to delete \"" + this.label(item) + "\"?")) {
+    if (window.confirm("Delete \"" + this.label(item) + "\"?")) {
       await this.props.deleteItem(item[this.identifierKey]);
       this.props.fetchData();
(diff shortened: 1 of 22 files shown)


src/components/EditableConfigList.tsx
```diff
@@ -141,6 +161,13 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
     }
     return null;
   }
+
+  async delete(item: U): Promise<void> {
+    if (window.confirm("Are you sure you want to delete \"" + this.label(item) + "\"?")) {
```
leonardr: This is unsolicited UI advice, but one thing I read a long time ago that has stuck with me is that people who try to delete something really want to delete _something_, just not necessarily the thing they selected when they triggered the delete action. So a more effective question to ask would be: "is X really the thing you want to delete?"

aslagle: That's interesting. It seems like the most important thing in that advice is including the item, rather than just "Are you sure?". I googled and I didn't find anything like what you were talking about, but some people recommend making it short, like "Delete X?", and a lot say you should implement undo instead.

leonardr: Including the name of the item is definitely the key takeaway. This was a _really_ long time ago, like, in a book on MS-DOS programming.

PR #135, opened 2018-01-15
Added outstanding metadata to book edit form.

aslagle:
This PR adds some outstanding metadata (authors, contributors, publisher, imprint, publication date, language, medium, and quality (sort of, staff can now put in a rating which affects quality)) to the book edit form. I think that covers everything librarians would need to edit.

For authors and contributors, name and role are editable, and role is a dropdown. Publisher and imprint are normal text fields. Publication date is a date picker. Language is a text field with autocomplete, which should work in most browsers except Safari. In Safari it's just a normal text field. Medium is a dropdown. Quality is a number from 1-5, although I'd like to come back to it and make it something nicer like stars sometime.

This depends on https://github.com/NYPL-Simplified/opds-feed-parser/pull/41 and https://github.com/NYPL-Simplified/opds-web-client/pull/214 and travis tests won't pass until those are merged and published.

### DIFF
--- a/src/components/BookEditForm.tsx
+++ b/src/components/BookEditForm.tsx
@@ -179,7 +179,7 @@ export default class BookEditForm extends React.Component<BookEditFormProps, Boo
           label="Rating (1-5, higher is better)"
           min={1}
           max={5}
-          value={this.props.rating && String(Math.round((this.props.rating * 4) + 1))}
+          value={this.props.rating && String(Math.round(this.props.rating))}
           />
         <EditableInput
           elementType="textarea"
--- a/src/components/BookEditForm.tsx
+++ b/src/components/BookEditForm.tsx
@@ -1,15 +1,33 @@
 import * as React from "react";
 import EditableInput from "./EditableInput";
-import { BookData } from "../interfaces";
+import { BookData, ContributorData, RolesData, MediaData, LanguagesData } from "../interfaces";
+import WithRemoveButton from "./WithRemoveButton";
+import Autocomplete from "./Autocomplete";
 
 export interface BookEditFormProps extends BookData {
+  roles: RolesData;
+  languages: LanguagesData;
+  media: MediaData;
   disabled: boolean;
   refresh: () => any;
   editBook: (url: string, data: FormData) => Promise<any>;
 }
 
+export interface BookEditFormState {
+  contributors?: ContributorData[];
+}
+
 /** Edit a book's metadata in the edit tab on the book details page. */
-export default class BookEditForm extends React.Component<BookEditFormProps, any> {
+export default class BookEditForm extends React.Component<BookEditFormProps, BookEditFormState> {
+  constructor(props) {
+    super(props);
+    this.state = {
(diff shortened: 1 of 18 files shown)


src/components/BookEditForm.tsx
```diff
@@ -47,10 +124,63 @@ export default class BookEditForm extends React.Component<BookEditFormProps, any
               disabled={this.props.disabled}
               name="series_position"
               placeholder="#"
-              value={this.props.seriesPosition}
+              value={this.props.seriesPosition && String(this.props.seriesPosition)}
```
leonardr: Would this work if the series position was 0?

src/components/__tests__/BookDetailsEditor-test.tsx
```diff
@@ -10,31 +10,49 @@ import BookEditForm from "../BookEditForm";
...
   it("loads admin book url when given a new book url", () => {
     let permalink = "works/1234";
     let newPermalink = "works/5555";
-    let fetchBook = stub();
     let element = document.createElement("div");
     let wrapper = shallow(
       <BookDetailsEditor
         bookUrl={permalink}
-        fetchBook={fetchBook}
+        {...dispatchProps}
```
leonardr: I'm assuming this is a construct similar to Python `*args`.

src/components/__tests__/BookDetailsEditor-test.tsx
```diff
@@ -167,4 +179,40 @@ describe("BookDetailsEditor", () => {
...
+    let roles = {
+      "aut": "Author",
+      "nar": "Narrator"
+    };
+    let languages = {
+      "eng": ["English"],
+      "spa": ["Spanish"]
+    };
+    let media = {
+      "http://schema.org/AudioObject": "Audio",
```
leonardr: This is fine for this branch but I think we want to switch to using `http://bib.schema.org/Audiobook` and `http://schema.org/EBook`; these identifiers didn't exist when I originally made these choices.

src/components/__tests__/BookDetailsEditor-test.tsx
aslagle: More like `**kwargs` but yeah.

src/components/__tests__/BookDetailsEditor-test.tsx
aslagle: This is just a test, for the actual app I made a route at "/admin/media" for the circ manager to tell it these values. I don't think I hard-coded them anywhere else.

leonardr:
As always I'm afraid I missed something when reviewing this, but nothing jumps out as a possible problem except the "series position 0" thing.

aslagle:
Fixed the series position thing here. I think the metadata layer does not currently allow the series position to be set to 0 in the presentation edition though, should I change that too?

leonardr:
Yeah, I think that's smart. I don't know if series position 0 shows up in real life but it seems like an obvious thing you'd have to do if someone went back and wrote a prequel.

PR #136, opened 2018-01-25
Added tab on book details page for managing a book's list membership.

aslagle:
The idea is to let librarians/curators coordinate their work by making a list like "needs review". Once a librarian has reviewed a book they will remove it from the "needs review" list and possibly add it to a list that controls a lane. It's hard to do this from the list page since the list may have a large number of books. So this adds a way to do it right from the book page.

I'm not sure it belongs on its own tab, but it would be harder to put it on the details tab and I'm not sure if that's actually better.

### DIFF
--- a/src/components/__tests__/BookDetailsContainer-test.tsx
+++ b/src/components/__tests__/BookDetailsContainer-test.tsx
@@ -24,7 +24,8 @@ describe("BookDetailsContainer", () => {
     context = {
       editorStore: store,
       tab: "tab",
-      csrfToken: "token"
+      csrfToken: "token",
+      library: stub()
     };
     refreshCatalog = stub();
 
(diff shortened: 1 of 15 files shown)


src/components/__tests__/BookDetailsContainer-test.tsx
```diff
@@ -24,7 +24,8 @@ describe("BookDetailsContainer", () => {
     context = {
       editorStore: store,
       tab: "tab",
-      csrfToken: "token"
+      csrfToken: "token",
+      library: stub()
```
leonardr: I see that this is used later but I don't get why it was added in this branch in particular. Why wasn't it necessary before?

aslagle: Because the catalog part of the admin interface is built on opds-web-client, its URLs don't include the library separately - the book URL and collection URL are part of the URLs and they include it. Until now, nothing on the book details page needed to construct a URL that wasn't for part of the catalog. Now the lists tab links to the list edit page and it also makes a request to get all lists, and the library is necessary to make those URLs.

leonardr:
I'm missing a conceptual step about why it's hard to do this from the list page. Is it because you would have to find the book you reviewed in the list and then move it to another list? In that case I agree that it's better to do that from a book detail page.

aslagle:
Yeah. You would have to go to the list you wanted to remove the book from, find the book in a potentially long list that's not sortable, then go to the list you wanted to add it to and search for it, and find it in the search results. And if there are multiple books with the same title it can be hard to make sure you got the right one.

PR #138, opened 2018-01-29
Made it possible to associate collections with a custom list.

aslagle:
This adds a section to the custom list edit page where you can check off collections that should be associated with the list. I debated whether to put this on the collection edit page or the list page, and I think the list page makes sense plus it was easier to implement that way.  But I'm still concerned that it clutters up the page a lot, and particularly when we have advanced search it won't be a good use of space.

### DIFF
--- a/src/components/CustomListEditor.tsx
+++ b/src/components/CustomListEditor.tsx
@@ -179,16 +179,12 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
   }
 
   changeCollection(collection: AdminCollectionData) {
-    let hadCollection = false;
-    const newCollections = [];
-    for (const stateCollection of this.state.collections) {
-      if (stateCollection.id === collection.id) {
-        hadCollection = true;
-      } else {
-        newCollections.push(stateCollection);
-      }
-    }
-    if (!hadCollection) {
+    const hasCollection = this.hasCollection(collection);
+    let newCollections;
+    if (hasCollection) {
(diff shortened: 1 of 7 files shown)


src/components/CustomListEditor.tsx
```diff
@@ -1,14 +1,16 @@
 import * as React from "react";
-import { CustomListData, CustomListEntryData } from "../interfaces";
+import { CustomListData, CustomListEntryData, CollectionData as AdminCollectionData } from "../interfaces";
 import { CollectionData } from "opds-web-client/lib/interfaces";
```
aslagle: It's very confusing that there are two things called CollectionData. This is the first time I needed both in the same component. One represents a `Collection` in the sense the circ manager uses it, and the other is a collection of books from an OPDS catalog.

src/components/CustomListEditor.tsx
```diff
@@ -121,15 +145,53 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
+    for (const stateCollection of this.state.collections) {
+      if (stateCollection.id === collection.id) {
+        hadCollection = true;
+      } else {
+        newCollections.push(stateCollection);
+      }
+    }
+    if (!hadCollection) {
+      newCollections.push(collection);
+    }
```
leonardr: Can you use the `hasCollection` function defined above to check membership, and add the collection to `this.state.Collections` (or a clone of it) if it's not there? This seems like a lot of work to make a copy of a list and maybe add one item to it, but I don't know enough Javascript to know if the simpler way actually works.

PR #139, opened 2018-02-01
Separate route for custom list with entries

aslagle:
This uses the changes from https://github.com/NYPL-Simplified/circulation/pull/845.

Custom list entries are no longer fetched unless you're on the edit page for a list. I added a new action to hit the custom list endpoint to get entries when you are on that page.

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -7,8 +7,8 @@ import {
   MetadataServicesData, AnalyticsServicesData,
   CDNServicesData, SearchServicesData,
   DiscoveryServicesData, LibraryRegistrationsData,
-  CustomListsData, LanesData, RolesData, MediaData,
-  LanguagesData
+  CustomListsData, CustomListDetailsData, LanesData,
+  RolesData, MediaData, LanguagesData
 } from "./interfaces";
 import DataFetcher from "opds-web-client/lib/DataFetcher";
 import { RequestError, RequestRejector } from "opds-web-client/lib/DataFetcher";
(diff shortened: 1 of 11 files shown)


src/actions.ts
```diff
@@ -487,8 +488,16 @@ export default class ActionCreator extends BaseActionCreator {
...
-    const url = "/" + library + "/admin/custom_lists";
+  fetchCustomListDetails(library: string, id: string) {
+    const url = "/" + library + "/admin/custom_list/" + id;
+    return this.fetchJSON<CustomListDetailsData>(ActionCreator.CUSTOM_LIST_DETAILS, url).bind(this);
+  }
+
+  editCustomList(library: string, data: FormData, id?: string) {
+    let url = "/" + library + "/admin/custom_lists";
+    if (id) {
+      url = "/" + library + "/admin/custom_list/" + id;
```
aslagle: I changed this even though the old URL still works too.

src/components/CustomListEditor.tsx
```diff
@@ -56,7 +56,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
             { this.props.list &&
               <h4>ID-{this.props.list.id}</h4>
             }
-            { this.props.collections && this.props.collections.length &&
+            { this.props.collections && this.props.collections.length > 0 &&
```
aslagle: This was an unrelated bug - when the value here is `0` rather than `false` or `undefined` it actually gets rendered in the UI as `0`.

PR #141, opened 2018-02-12
Storage configuration

EdwinGuzman:
@aslagle I copied most of the boilerplate code for `SearchServices` to create `StorageServices` and it's accompanying reducers and actions.

For an updated version, do you do anything else besides bumping up the version in `package.json`?


src/components/ConfigTabContainer.tsx
```diff
@@ -109,6 +110,14 @@ export default class ConfigTabContainer extends TabContainer<ConfigTabContainerP
...
           identifier={this.props.identifier}
           />
+      ),
+      storage: (
+        <StorageServices
+          store={this.props.store}
+          csrfToken={this.props.csrfToken}
+          editOrCreate={this.props.editOrCreate}
+          identifier={this.props.identifier}
+          />
```
aslagle: I would move this tab up in the list - I was keeping the discovery one at the bottom since it's kind of different than everything else, and it would typically be the last thing you'd do.

aslagle:
You didn't write any tests and this code doesn't really need them, but you could at least add to https://github.com/NYPL-Simplified/circulation-web/blob/master/src/components/__tests__/ConfigTabContainer-test.tsx#L63 to make sure your new tab gets rendered.

src/components/ConfigTabContainer.tsx
EdwinGuzman: Updated.

EdwinGuzman:
I added it to the `ConfigTabContainer-test.tsx` file, but when I run the tests locally, some tests fail. But, it seems tests pass in Travis. Is that correct?

aslagle:
The tests should pass locally, so you should try to figure out what the problem is. But if they pass on travis I'm fine with merging this.

PR #142, opened 2018-02-20
Logging configurations

EdwinGuzman:
This PR adds a new component for rendering the sitewide logging settings. It's mostly similar to the `SitewideSettings` component and I modified `LoggingSettingsEditForm` so that it will dynamically render a select input based on what value was first selected in the parent select input. Everything seems to be working fine but I'm having trouble correctly running tests locally (I also am not too sure how/what I'm testing so I will revisit that again).

I made minor changes to `EditableInput.tsx` because I wanted to get back the value of the select input for the `onChange` handler. Types is still getting me confused.

Also, the tab on the left sidebar displays "LoggingSettings" and I can't seem to separate that into two words... I noticed it's the `key` when creating the tab, but when I change it to `logging` the routes seem to get messed up. Does that need to be updated in the `Circulation` repo?


aslagle:
For the tab name look at the `tabDisplayName` function lower down in the file.

aslagle:
I made a comment in https://github.com/NYPL-Simplified/circulation/pull/868 that would affect this so I think I'll wait to review it for now. Let me know if you need help getting the tests to run though.

src/components/SitewideSettingEditForm.tsx
```diff
@@ -12,13 +12,19 @@ export interface SitewideSettingEditFormProps {
 }
 
 /** Form for editing a single sitewide setting. */
-export default class SitewideSettingEditForm extends React.Component<SitewideSettingEditFormProps, void> {
+export default class SitewideSettingEditForm extends React.Component<SitewideSettingEditFormProps, any> {
```
aslagle: You should define a `SitewideSettingEditFormState` type and use that instead of `any`.

src/components/SitewideSettingEditForm.tsx
```diff
@@ -12,13 +12,19 @@ export interface SitewideSettingEditFormProps {
...
     this.save = this.save.bind(this);
+    this.onChange = this.onChange.bind(this);
+    this.state = {
+      inputKey: this.availableSettings().length ? this.availableSettings()[0].key : "",
+    };
   }
 
   render(): JSX.Element {
+    const inputKey = this.state.inputKey;
+    const logType = !!(inputKey === "log_level" || inputKey === "database_log_level");
```
aslagle: It might be better to have a way to check the setting type is "select" rather than checking for the hard-coded keys.

src/components/EditableInput.tsx
```diff
@@ -4,7 +4,7 @@ export interface EditableInputProps extends React.HTMLProps<EditableInput> {
   elementType?: string;
   label?: string;
   description?: string;
-  onChange?: () => any;
+  onChange?: (e: any) => any;
```
aslagle: I feel like this argument should have a type but I'm not sure what it is, did you try string?

src/components/__tests__/LoggingServices-test.tsx
```diff
@@ -0,0 +1,99 @@
...
+        fetchData={fetchData}
+        editItem={editItem}
+        csrfToken="token"
+        isFetching={false}
+        />
+    );
+  });
+
+  it("shows logging service list", () => {
+    console.log(wrapper);
```
aslagle: I wish there was a better way to debug these tests...

src/components/EditableInput.tsx
EdwinGuzman: I did, and I got that it was not assignable to type `'EventHandler<FormEvent>'`.

aslagle: Interesting, that must be because React.HTMLProps already has an onChange prop. Maybe we should rename this prop so it doesn't conflict.

src/components/__tests__/LoggingServices-test.tsx
EdwinGuzman: It was also me not fully understanding how/what it was rendering.

EdwinGuzman:
Updated.

aslagle:
Cool. I'd also like to have a test for the select in SitewideSettingsEditForm, and maybe for switching between a select and a text input.

EdwinGuzman:
Updated. I'm going to try to figure out why my `.simulate("change")` isn't working.

PR #144, opened 2018-03-09
Added library registrations to collection settings page

aslagle:
This branch takes some of the code for registering libraries with a discovery service like the library registry and reuses it for the collections page. It's for a branch in circulation that will enable one circulation manager to share its ODL collections with libraries on other circ managers. The other libraries will need to use this to register with the main circ manager.


aslagle:
One thing I don't like about this is that registering libraries is separate from associating them with the collection. So you could register a library but not give it access to the collection, and vice versa, which is not great.

src/components/__tests__/Collections-test.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    const library = { short_name: "nypl" };
+    context.registerLibrary(library);
+
+    expect(registerLibrary.callCount).to.equal(1);
+    const formData = registerLibrary.args[0][0];
+    expect(formData.get("library_short_name")).to.equal("nypl");
+    expect(formData.get("collection_id")).to.equal("2");
+
+    let pause = new Promise<void>(resolve => setTimeout(resolve, 0));
```
EdwinGuzman: Is this because the test is async?

aslagle: Yeah, I wanted to make sure fetchLibraryRegistrations got called. That happens when a promise is resolved (from the mock function on line 14). That waits until the call stack clears, so I need the setTimeout to let the waiting code run before finishing the test.

aslagle:
Yeah, quite a bit is boilerplate and it's similar to the other code. It could probably be improved for the whole app later.

PR #147, opened 2018-03-28
Audiobook admin search

EdwinGuzman:
Added a radio select option in the admin search to select between audiobook, ebook, or all media types.

Ties in with [circulation #907](https://github.com/NYPL-Simplified/circulation/pull/907) and [server-core #833](https://github.com/NYPL-Simplified/server_core/pull/833).

### DIFF
--- a/src/components/CustomListEditor.tsx
+++ b/src/components/CustomListEditor.tsx
@@ -19,10 +19,17 @@ export interface CustomListEditorProps extends React.Props<CustomListEditor> {
   isFetchingMoreSearchResults: boolean;
 }
 
+export interface Media {
+  type: string;
+  label: string;
+  active: boolean;
+}
+
 export interface CustomListEditorState {
   name: string;
   entries: CustomListEntryData[];
   collections: AdminCollectionData[];
+  media?: Media[];
 }
 
 /** Right panel of the lists page for editing a single list. */
@@ -32,7 +39,12 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     this.state = {
       name: this.props.list && this.props.list.name,
       entries: (this.props.list && this.props.list.entries) || [],
-      collections: (this.props.list && this.props.list.collections) || []
+      collections: (this.props.list && this.props.list.collections) || [],
+      media: [
+        { type: "all", label: "All", active: true },
+        { type: "audiobooks", label: "Audiobooks", active: false },
+        { type: "ebooks", label: "E-books", active: false }
+      ],
     };
 
     this.changeName = this.changeName.bind(this);
@@ -40,6 +52,8 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     this.save = this.save.bind(this);
     this.reset = this.reset.bind(this);
     this.search = this.search.bind(this);
+    this.changeMedia = this.changeMedia.bind(this);
+    this.getMediaTerm = this.getMediaTerm.bind(this);
   }
 
   render(): JSX.Element {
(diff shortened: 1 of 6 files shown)


src/interfaces.ts
```diff
@@ -325,3 +325,9 @@ export interface LaneData {
 export interface LanesData {
   lanes: LaneData[];
 }
+
+export interface Media {
```
aslagle: There's already a `MediaData` interface in here, used when getting all the media types from "/admin/media". Can you reuse that?

src/components/CustomListEditor.tsx
```diff
@@ -23,6 +28,7 @@ export interface CustomListEditorState {
   name: string;
   entries: CustomListEntryData[];
   collections: AdminCollectionData[];
+  media?: Media[];
```
aslagle: If you were going to use `MediaData` here, I would suggest adding a prop that contains all the media types (make a request to get them from the server on mount). And then the state can contain a list of the currently active media types rather than all of them.

EdwinGuzman:
Updated. I talked to Leonard about the options and agreed to filter the options so that only `Audio` and `Book` will come up in the UI.

src/components/CustomListEditor.tsx
```diff
@@ -32,14 +40,19 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     this.state = {
       name: this.props.list && this.props.list.name,
       entries: (this.props.list && this.props.list.entries) || [],
-      collections: (this.props.list && this.props.list.collections) || []
+      collections: (this.props.list && this.props.list.collections) || [],
+      media: this.props.media || {},
```
aslagle: I don't think you need `media` in the state, doesn't look like you use it.

src/components/CustomListEditor.tsx
```diff
@@ -32,14 +40,19 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
+      mediaSelected: "all",
     };
 
     this.changeName = this.changeName.bind(this);
     this.changeEntries = this.changeEntries.bind(this);
     this.save = this.save.bind(this);
     this.reset = this.reset.bind(this);
     this.search = this.search.bind(this);
+    this.changeMedia = this.changeMedia.bind(this);
+    this.getMediaQuery = this.getMediaQuery.bind(this);
```
aslagle: I don't think you need to bind `getMediaQuery` here, since it's called from other functions in the component and not used as a React prop. It doesn't hurt anything though.

aslagle:
It looks like this will save your media selection when switching lists, which seems fine to me. If that is what we want I'd like to have a test for it.

src/components/CustomListEditor.tsx
EdwinGuzman: I typically bind functions even if they're not used as a prop. But, makes sense to not include.

EdwinGuzman:
Hmm I actually don't know if that is something we want. Resetting it makes more sense to me. What do you think?

aslagle:
I could go either way. I can't remember if it resets your search term when you change lists - maybe do the same thing it does with the search term.

src/components/CustomListEditor.tsx
aslagle: It's fine either way, just wanted to mention it.

EdwinGuzman:
Updated. Added a test in `CustomListEditor-test` that checks the state and value of the text input as the list prop gets updated. At first I tried to do it in `CustomList-test` since it seemed more appropriate to test there when you select a new list (or go to a new route). But, that proved difficult because I couldn't access the `CustomListEditor` component as a rendered element. And when I could access it, it didn't seem like I was testing it correctly.

EdwinGuzman:
Updated circulation and server_core with that update.

PR #148, opened 2018-04-04
Audio book icons

EdwinGuzman:
* Imported `@nypl/dgx-svg-icons` for the `AudioHeadphoneIcon` and `BookIcon` SVG components.
* Displaying the respective icon for each item in search results and entries in the the custom lists editor.
* Updated the custom list editor to use `ul` and `li` instead of divs.
* Updated SVG components to have better titles and hid them using `aria-hidden`.


package.json
```diff
@@ -13,13 +13,15 @@
     "prepublish": "typings install && npm run prod",
     "lint": "tslint -c tslint.json src/*.ts src/*.tsx src/**/*.ts src/**/*.tsx src/**/**/*.ts src/**/**/*.tsx && sass-lint -c .sass-lint.yml -v -q",
     "test": "npm run lint && tsc && cp -r src/stylesheets lib && mocha --require lib/testHelper.js lib/__tests__/*.js lib/**/__tests__/*.js lib/**/**/__tests__/*.js",
+    "test-file": "npm run lint && tsc && cp -r src/stylesheets lib && mocha --require lib/testHelper.js",
```
EdwinGuzman: Added this here just because testing one file at a time is faster.

src/components/CustomListEntriesEditor.tsx
```diff
@@ -188,6 +194,19 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
     return book.raw["simplified:pwid"][0]["_"];
   }
 
+  getMedium(book) {
+    return book.raw["$"]["schema:additionalType"].value;
```
EdwinGuzman: This, as the above which is similar, worries me a little bit because there are no checks to see if `$` or `schema:additionalType` are in the `raw` attribute. Is this something that we always expect to have according to the OPDS spec? (Must admit that still haven't delved too deep into the docs for OPDS).

aslagle: I don't think this is part of the OPDS spec at all, it's something we added. However, this code is only used with the circ manager so if the circ manager always includes that attribute it's ok. For the pwid, I felt it was safe to assume it will include the pwid in OPDS entries- if there was a bug that broke them, this page wouldn't work anyway.

src/components/CustomListEntriesEditor.tsx
```diff
@@ -188,6 +194,19 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
+  }
+
+  getMediumSVG(medium) {
+    const svgMediumTypes = {
+      "http://bib.schema.org/Audiobook": <AudioHeadphoneIcon ariaHidden className="draggable-item-icon" />,
+      "http://schema.org/EBook": <BookIcon ariaHidden className="draggable-item-icon" />,
+    };
+
+    return svgMediumTypes[medium];
```
aslagle: Here you should probably handle the case where you get an unexpected medium.

EdwinGuzman: Updated.

aslagle:
@EdwinGuzman This looks good but I'm having CSS problems from your previous PR now that I'm actually testing it (probably should've tried running that one before I approved). Do you mind fixing this as well?
screen shot 2018-04-04 at 10 35 00 pm

aslagle:
I would also suggest putting those radio buttons in a horizontal list instead of vertical - the drag and drop UI at the bottom should get as much vertical space as possible.

EdwinGuzman:
Thanks. Updated the CSS. Let me know if it's better because I didn't see that issue on my end.

Issue #146, opened 2018-03-21
Provide detail view for books selected from a list


leonardr:
Clicking on a book in a list of books (e.g. when building a custom list) should bring up the detail page used to edit a book's metadata (or possibly a separate read-only detail page like that used in opds-web-client). This will allow librarians to see more information about a book before adding it to a list.

PR #149, opened 2018-04-09
Search results link

### LINKED ISSUE #120: Link to book detail pages from list view
If there are multiple books with the same title or an incorrect title, the admin needs to be able to get to the book detail page for a specific book in the list.

Also, add some padding below the "Load more" button.

EdwinGuzman:
Fixes #120  and #146  (which are the same).

Added a link to see the details for each item that comes up in the search results and also a link to go back to the previous page from the book details page. I only added it to the search results and not the entries that are already in the list.

Updated some of the flex-box css for each search result item:
screen shot 2018-04-09 at 12 36 58 pm


aslagle:
I'd say it should be in the entries in the list as well....

src/components/BookDetailsContainer.tsx
```diff
@@ -28,6 +30,13 @@ export default class BookDetailsContainer extends React.Component<BookDetailsCon
   render(): JSX.Element {
     return (
       <div className="book-details-container">
+        { !this.props.collectionUrl &&
+          <a
+            href="#"
+            onClick={() => history.back()}
+            className="list-back-link"
+          ><GenericWedgeIcon /> Go Back to list</a>
```
aslagle: Seems risky to assume you came from a list here. Someone might have shared the link, or we might add more places that link to individual books in the future. I'd probably prefer for the links open a new window/tab anyway.

EdwinGuzman: Yes, I thought the same but didn't know how else to pass the list id to this component. I tried a few ways and none worked.

I think everything should be kept on the same window, if possible. It's less of a UX issue if the app is kept on the same page, but for now it seems like the better approach. I'll update it.

EdwinGuzman:
I figured that if an item is in the list already, the user would already about it and wouldn't need that link in the entries, but I can add it.

aslagle:
There might be multiple users working on the same list, so they wouldn't necessarily know about it.

src/components/BookDetailsContainer.tsx
aslagle: One thing you could do is use the OPDS feed URL for the list as the collection URL. I still wouldn't use a "back to list" button since you don't know how they got there, but it's a way to show the book in the context of a list, so it would be reasonable to have an "edit list" button.

EdwinGuzman:
Added the link for each item in the entry list. Just have some comments about it.

src/components/CustomListEntriesEditor.tsx
```diff
@@ -194,6 +198,19 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
   }
 
+  getCatalogLink(book, type = "") {
+    if (!book) {
+      return null;
+    }
+    let bookUrl = book.url;
+
+    if (type === "entry") {
+      bookUrl = (this.props.library ? this.props.library : "") + encodeURIComponent(`/URI/${book.url}`);
```
EdwinGuzman: The url I'm exposing for each entry isn't the full url to each item, so I'm creating it here. The only problem is that if `this.props.library` is not defined, this won't link properly.

aslagle: This won't work in general - "URI" is one identifier type but not all books use a URI as the identifier. You would need to also expose the identifier type. I think it would be better to use `url_for` on the server to get the url for the `permalink` route and send that down.

aslagle:
The OPDS feed url for a list is `/<library>/lists/<list name>/crawlable`.

EdwinGuzman:
Thanks. Added that now.

src/components/CustomListEntriesEditor.tsx
```diff
@@ -194,6 +198,24 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
     return book.raw["simplified:pwid"][0]["_"];
   }
 
+  getCatalogLink(book) {
+    if (!book) {
+      return null;
+    }
+    let bookUrl = book.url ? book.url.replace(/\/admin/, "") : "";
```
EdwinGuzman: I used `url_for` on the server but I couldn't figure out how to get an URL that didn't contain `/admin/` in it. So I'm removing it here if it's in the url that gets passed.

aslagle:
This all looks good except when you add an item from search results to the list it loses its 'view details'  button.

EdwinGuzman:
Nice catch. Fixed that.

PR #151, opened 2018-04-12
Updating the media filter when searching to use entry points instead.

EdwinGuzman:
Help fix [circulation #931](https://github.com/NYPL-Simplified/circulation/issues/931).

Uses a library's entry points settings instead of media to generate the available radio button options in the UI for searching. The "All" doesn't work yet and will be changed later.

### DIFF
--- a/src/components/CustomListEditor.tsx
+++ b/src/components/CustomListEditor.tsx
@@ -3,7 +3,6 @@ import {
   CustomListDetailsData,
   CustomListEntryData,
   CollectionData as AdminCollectionData,
-  MediaData,
 } from "../interfaces";
 import { CollectionData } from "opds-web-client/lib/interfaces";
 import TextWithEditMode from "./TextWithEditMode";
@@ -22,14 +21,14 @@ export interface CustomListEditorProps extends React.Props<CustomListEditor> {
   search: (url: string) => Promise<CollectionData>;
   loadMoreSearchResults: (url: string) => Promise<CollectionData>;
   isFetchingMoreSearchResults: boolean;
-  media?: MediaData;
+  entryPoints?: string[];
 }
 
 export interface CustomListEditorState {
   name: string;
   entries: CustomListEntryData[];
   collections: AdminCollectionData[];
-  mediaSelected?: string;
+  entryPointSelected?: string;
 }
 
 /** Right panel of the lists page for editing a single list. */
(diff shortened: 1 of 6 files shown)


aslagle:
If there aren't any entry points, I'm getting a single "All" button. I think it just shouldn't show anything in that case.

I'm also not sure what you mean about "All" not working, but I would probably take it out in that case and put it back in when it works.

There is also some CSS that needs updated.

src/components/CustomListEditor.tsx
```diff
@@ -254,65 +253,57 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
-    mediaElems.push(
+  getEntryPointsElms(entryPoints) {
+    const entryPointsElms = [];
+    entryPointsElms.push(
       <EditableInput
         key="all"
         type="radio"
-        name="media-selection"
-        checked={"all" === this.state.mediaSelected}
+        name="enty-points-selection"
```
aslagle: you have a typo here

src/components/CustomListEditor.tsx
```diff
@@ -254,65 +253,57 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
         <EditableInput
-          key={key}
+          key={entryPoint}
           type="radio"
-          name="media-selection"
-          checked={key === this.state.mediaSelected}
-          label={label}
-          value={key}
-          onChange={() => this.changeMedia(key)}
+          name="enty-points-selection"
```
aslagle: and here

src/components/CustomLists.tsx
```diff
@@ -279,7 +298,7 @@ function mapStateToProps(state, ownProps) {
     searchResults: state.editor.collection && state.editor.collection.data,
     isFetchingMoreSearchResults: state.editor.collection && state.editor.collection.isFetchingPage,
     collections: state.editor.collections && state.editor.collections.data && state.editor.collections.data.collections,
-    media: state.editor.media.data,
+    libraries: state.editor.libraries.data,
```
aslagle: it doesn't really matter but i'd probably get `libraries.data.libraries` here since nothing else in this object is needed later.

EdwinGuzman:
Updated. Thanks for catching those issues.

aslagle:
I'm getting this again:
screen shot 2018-04-18 at 12 02 04 am

But otherwise looks good :)

EdwinGuzman:
I added flexbox so that shouldn't happen anymore.

aslagle:
I'm not testing this last thing, but if you were able to reproduce the problem I trust that this will fix it.

PR #153, opened 2018-04-23
Admin roles

aslagle:
This goes with https://github.com/NYPL-Simplified/circulation/pull/935. It adds an Admin object to the context that's used by various components to check the admin's roles and decide what to display. It also adds a bunch of checkboxes to the admin edit page for setting an admin's roles.

### DIFF
--- a/src/components/EditableConfigList.tsx
+++ b/src/components/EditableConfigList.tsx
@@ -74,7 +74,7 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
 
         { !this.props.isFetching && !this.props.editOrCreate && this.props.data && this.props.data[this.listDataKey] &&
           <div>
-            { (!this.limitOne || this.props.data[this.listDataKey].length === 0) &&
+            { (!this.limitOne || this.props.data[this.listDataKey].length === 0) && this.canCreate() &&
               <a
                 className="btn btn-default create-item"
                 href={this.urlBase + "create"}
@@ -95,15 +95,17 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
                     <h4>
                       {this.label(item)}
                     </h4>
-                    <button
-                      className="btn btn-danger delete-item"
-                      onClick={() => this.delete(item) }
-                      >
-                        <span>
-                          Delete
-                          <TrashIcon />
-                        </span>
-                    </button>
+                    { this.canDelete(item) &&
+                      <button
+                        className="btn btn-danger delete-item"
+                        onClick={() => this.delete(item) }
+                        >
+                          <span>
+                            Delete
+                            <TrashIcon />
+                          </span>
+                      </button>
+                    }
                   </li>
                 )
               }
@@ -144,6 +146,14 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
 
   label(item): string {
     return item[this.labelKey];
+  }
+
+  canCreate() {
+    return true;
+  }
+
+  canDelete(item) {
+    return true;
   }
 
   componentWillMount() {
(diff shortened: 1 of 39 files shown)


aslagle:
The thing I'm least confident about in this PR is having an Admin model object and using it in the context as well as in a component state. I definitely could be making better use of Redux in this whole app.

src/components/IndividualAdminEditForm.tsx
```diff
@@ -57,9 +143,138 @@ export default class IndividualAdminEditForm extends React.Component<IndividualA
...
+    if (role === "librarian") {
+      if (this.isSelected("librarian-all")) {
+        return !this.context.admin.isSitewideLibraryManager();
+      }
+      return !this.context.admin.isLibraryManager(library);
+    }
+  }
+
+  handleRoleChange(role: string, library?: string) {
+    if (role === "system") {
```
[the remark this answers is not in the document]
leonardr: This makes sense; the only thing I'd say is that it's a little confusing when a method indicates a negative state rather than a positive ('disabled' rather than 'enabled'), and leads to double negatives in the code. But I don't think it's worth changing the amount of code you'd need to change to 'fix' that.

src/models/Admin.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (this.isSystemAdmin()) {
+      return true;
+    }
+    if (role !== "system" && this.isSitewideLibraryManager()) {
+      return true;
+    }
+    if (role === "librarian-all" && this.isSitewideLibrarian()) {
+      return true;
+    }
+    if (role === "manager" && this.isLibraryManager(library)) {
```
leonardr: There's a role called "manager-all" that isn't handled here.

aslagle: That would get handled by the `role !== "system" && this.isSitewideLibraryManager()` clause.

EdwinGuzman:
Is it possible to be System Admin and a Library Manager of one library but not a Library Manager of a second Library? When I try to disable a checkbox under "Library Manager" for a library, I also lose the System Admin role.

EdwinGuzman:
I also created another account that is a `Library Manager` of just one Library in the system, but when I log in using that account, I can see both Libraries in the System Configuration and I can edit the configuration of the second Library in my system that I should not be a `Library Manager` of.

aslagle:
That second thing sounds like a bug though. If you're only a `Library Manager` of one library, you shouldn't see other libraries.

src/components/Header.tsx
```diff
@@ -93,18 +101,40 @@ export class Header extends React.Component<HeaderProps, void> {
...
+              <li className="dropdown">
+                <a
+                  className="dropdown-toggle"
+                  href="#"
+                  role="button"
+                  aria-haspopup="true"
+                  aria-expanded={this.state.showAccountDropdown}
+                  onClick={this.toggleAccountDropdown}
+                  >{ this.context.admin.email } &#9660;</a>
+                { this.state.showAccountDropdown &&
```
EdwinGuzman: Not a big deal for this PR but for accessibility we want to close dropdowns when the user clicks outside of the element (which we've had plenty of issues trying to get that done).

src/components/EditableConfigList.tsx
```diff
@@ -146,6 +148,14 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
     return item[this.labelKey];
   }
 
+  canCreate() {
```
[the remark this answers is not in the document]
EdwinGuzman: Got it.

EdwinGuzman:
For admins that do not have the `System Admin` role, I noticed I can't delete any admins or update their roles, but I can still change any other admin's password.

aslagle:
Did it actually work to change the password? If you're a library manager you should not be able to change a system admin's password, but you should be able to change the passwords of other managers and librarians of your library. Or maybe I should remove that functionality? Either way, I should remove the password box if it doesn't work.

aslagle:
Thanks, I'm fixing that so it does a better job of checking the roles and hiding the password box.

aslagle:
@EdwinGuzman I think I've fixed everything except closing the dropdown when you click somewhere else.

PR #154, opened 2018-05-03
Different sources book list

EdwinGuzman:
Helps fix [circulation 938](https://github.com/NYPL-Simplified/circulation/issues/938).

Extracting the data source value from each search result item's opds feed object and passing it as `data_source_id` in the respective entry object when it gets added to a list. The `data_source_id` value of an entry object is an id number, but the value from the opds feed is a string. I'm looking that value up in the circulation manager and converting it to correct id before making a query. That's the only thing that bothers me about this since the value from a search result item to an entry will change from something like `"Standard eBooks"` to `"34"`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.0.66",
+  "version": "0.0.67",
   "description": "Web Front-end for Library Simplified Circulation Manager Admin Interface",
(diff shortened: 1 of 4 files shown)


aslagle:
I would use the data source name string for both the OPDS feed and the list entries to be consistent.

EdwinGuzman:
Updated.

PR #155, opened 2018-05-29
Added tab for changing a book's cover.

aslagle:
This branch goes with https://github.com/NYPL-Simplified/circulation/pull/968. It adds a tab to the book details page for uploading a new open-access image to use as a book cover.

### DIFF
--- a/src/reducers/bookCover.ts
+++ b/src/reducers/bookCover.ts
@@ -1,0 +1,4 @@
+import ActionCreator from "../actions";
+import createFetchEditReducer from "./createFetchEditReducer";
+
+export default createFetchEditReducer<string>(ActionCreator.BOOK_COVER, ActionCreator.EDIT_BOOK_COVER);
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.0.67",
+  "version": "0.0.68",
   "description": "Web Front-end for Library Simplified Circulation Manager Admin Interface",
   "repository": {
     "type": "git",
--- a/src/__tests__/actions-test.ts
+++ b/src/__tests__/actions-test.ts
@@ -115,6 +115,31 @@ describe("actions", () => {
       }
     });
 
+    it("dispatches failure on non-JSON response", async () => {
+      const dispatch = stub();
+      const nonJsonResponse = new Promise(() => { throw "Not JSON"; });
+      const fetchMock = stub().returns(new Promise<any>(resolve => {
+        resolve({ status: 500, json: () => nonJsonResponse });
+      }));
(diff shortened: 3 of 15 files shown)


src/reducers/bookCover.ts
```diff
@@ -0,0 +1,4 @@
+import ActionCreator from "../actions";
+import createFetchEditReducer from "./createFetchEditReducer";
+
+export default createFetchEditReducer<string>(ActionCreator.BOOK_COVER, ActionCreator.EDIT_BOOK_COVER);
```
aslagle: There are actually no "fetch" actions for this reducer, only edit, but there may need to be a fetch action later if we ever add a feature for an admin to see the rights for an existing uploaded cover. For now, the ActionCreator.BOOK_COVER constant is not used except to define this reducer.

EdwinGuzman:
This looks great. I tested it out but could not upload any images. I get this as 

```
File \"/Users/edwinguzman/projects/NYPL-Simplified/circulation/api/admin/controller.py\", line 1275, in change_book_cover\n    work.calculate_presentation(policy=presentation_policy)\n  File \"/Users/edwinguzman/projects/NYPL-Simplified/circulation/core/model.py\", line 4654, in calculate_presentation\n    float(quality) != float(self.quality)\nTypeError: float() argument must be a string or a number", "message": "Exception in web app: float() argument must be a string or a number", "filename": "app_server.py"
```

I tested it by just downloading a the book cover of the book I wanted to update and re-upload it. Using a book cover's url didn't work for me either:

```
 image = Image.open(image_file)\n  File \"/Users/edwinguzman/projects/NYPL-Simplified/circulation/env/lib/python2.7/site-packages/PIL/Image.py\", line 2572, in open\n    % (filename if filename else fp))\nIOError: cannot identify image file <StringIO.StringIO instance at 0x110e69d40>", "message": "Exception in web app: cannot identify image file <StringIO.StringIO instance at 0x110e69d40>", "filename": "app_server.py"}
```

Also, is there a list of requirements for this feature? Not necessary but just curious.

aslagle:
The first error is weird, I didn't change anything related to quality. Can you change other metadata for that book?

For the second one, could you share the cover URL you tried?

The only enforced requirements for the image are that it be at least 600px wide and 900px high. It should be tall, but the server doesn't check. It also doesn't verify the file type.

EdwinGuzman:
I tried different combinations of title/author position and license but I kept getting the same first error. For this example, I downloaded and tried to upload this cover image: https://standardebooks.org/ebooks/abraham-merritt/the-moon-pool/dist/cover.jpg

For the second issue, I used the same image above which also couldn't give me a preview.

EdwinGuzman:
Working for me.

PR #156, opened 2018-06-03
Changed list edit page to use identifiers instead of pwids.

aslagle:
This is the front-end part of https://github.com/NYPL-Simplified/circulation/pull/974


aslagle:
Right.

PR #157, opened 2018-06-06
Added distributor, publisher, and audience to book details tab.

aslagle:
This stuff was already in circulation-patron-web but is also useful for admins so I copied it over. It adds a subclass of BookDetails from opds-web-client so it can include some additional info on that tab.

Might be worth figuring out how to reuse but I'm not sure what the future of circulation-patron-web is.

### DIFF
--- a/src/components/BookDetails.tsx
+++ b/src/components/BookDetails.tsx
@@ -1,0 +1,107 @@
+import * as React from "react";
+import DefaultBookDetails, { BookDetailsProps as DefaultBookDetailsProps } from "opds-web-client/lib/components/BookDetails";
+
+export default class BookDetails extends DefaultBookDetails<DefaultBookDetailsProps> {
+
+  fieldNames() {
+    return ["Published", "Publisher", "Audience", "Categories", "Distributed By"];
+  }
+
+  fields() {
+    let fields = super.fields();
+    let categoriesIndex = fields.findIndex(field => field.name === "Categories");
+    fields[categoriesIndex].value = this.categories();
+    fields.push({
+      name: "Audience",
+      value: this.audience()
+    });
+    fields.push({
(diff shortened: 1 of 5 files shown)


EdwinGuzman:
What do you mean regarding the the future holds circulation-patron-web?

aslagle:
No one has concrete plans to work on it yet and I don't think the way it's set up now will scale well for supporting multiple libraries. So I'm not sure how much of that repo will end up being used in the future.

src/components/BookDetails.tsx
```diff
@@ -0,0 +1,107 @@
...
+    });
+    return fields;
+  }
+
+  audience() {
+    if (!this.props.book) {
+      return null;
+    }
+
+    let categories = this.props.book.raw.category;
```
EdwinGuzman: If there's no `book` or `raw` property, then this will throw errors. There are no tests to cover those cases but is it possible to have those properties missing in the opds feed?

aslagle: `book` is a required prop in the opds-web-client class this inherits from, but `raw` is indeed optional in the `BookData` interface. I think the XML parsing library we use always includes it though. I'm less worried about things like this in the admin interface since we know the feeds are always coming from the circ manager. If categories weren't there, that would be a bug in the circ manager that should be fixed.

EdwinGuzman: Cool, I'm the opposite and things like that worry me, even if I know the data source should always return it.

aslagle: Yeah, it's probably good to be worried about that :)

aslagle: @EdwinGuzman You were right about this :)
https://github.com/NYPL-Simplified/circulation-web/pull/158

PR #158, opened 2018-06-19
Fixed exception when a book has no categories.

aslagle:
Due to https://github.com/NYPL-Simplified/server_core/pull/889 and https://github.com/NYPL-Simplified/server_core/pull/888, the circ manager no longer assumes books with no classification info are adult nonfiction. This means it is now possible for an opds entry to have no categories and we need to handle that case.


EdwinGuzman:
I can create a ticket for handling opds feeds with no categories.

aslagle:
This might be enough to handle it. I'm not sure if there are any other places it would cause issues.

PR #159, opened 2018-06-21
Entrypoints tabs

EdwinGuzman:
This is related to https://github.com/NYPL-Simplified/opds-web-client/pull/216

This PR creates two entry points related components that will get rendered when there are entry points configured for a library. Right now it's only ebooks and audiobooks.


src/components/__tests__/EntryPointsTabContainer-test.tsx
```diff
@@ -0,0 +1,140 @@
...
+    let catalogLinks = wrapper.find(CatalogLink);
+    expect(catalogLinks.length).to.equal(1);
+
+    let bookLink = catalogLinks.at(0);
+    expect(bookLink.prop("collectionUrl")).to.equal("home url?entrypoint=Audio");
+    expect(bookLink.prop("bookUrl")).to.equal(null);
+    expect(bookLink.children().at(1).text()).to.equal("Audio Books");
+  });
+
+  it("uses router to navigate when a tab is clicked", () => {
```
EdwinGuzman: I'm having trouble with this test. I tried to configure it to another test file that was doing something similar but I think it was using an `<a>` element and not a `<CatalogLink>` element. Is this the right approach?

package.json
```diff
@@ -25,8 +25,8 @@
     "bootstrap": "^3.3.6",
     "font-awesome": "^4.6.3",
     "isomorphic-fetch": "^2.2.1",
-    "opds-feed-parser": "^0.0.14",
-    "opds-web-client": "^0.1.18",
+    "opds-feed-parser": "0.0.14",
+    "opds-web-client": "0.1.19",
```
EdwinGuzman: This needs to be published first.

src/components/EntryPointsContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  render(): JSX.Element {
+    const mapEntryPointsToLabels = {
+      "http://schema.org/EBook": "eBooks",
+      "http://bib.schema.org/Audiobook": "Audio Books",
+    };
+
+    const { collectionUrl } = this.props;
+    const entryPointQuery = collectionUrl ? collectionUrl.split("?") : [];
```
aslagle: I'm wondering the entry point could come from the collection data instead of getting it from the URL. I'm not sure if it's in there already, but it is represented in the OPDS feed so it should be possible. It might be cleaner and might make it easy for the server to specify the entry point labels instead of hard-coding them here.

src/components/EntryPointsContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+    // Making Book the default value
+    const entrypointValue =
+      entryPointQuery[1] ? entryPointQuery[1].split("=")[1] : "Book";
+    const homeLink = this.props.links[0] ? this.props.links[0].url : "";
+
+    // The tabs should only display on the homepage
+    if (this.props.links.length > 2) {
```
aslagle: Now that I see how this is used I would call the prop breadcrumbs, links could be anything.

src/components/EntryPointsTabContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+  activeValue?: string;
+  homeLink?: string;
+  collectionUrl: string;
+  library: (collectionUrl, bookUrl) => string;
+}
+
+export interface EntryPointsTabContainerProps extends EntryPointsTabContainerDispatchProps, EntryPointsTabContainerOwnProps {}
+
+/** Wraps the book details component from OPDSWebClient with additional tabs
+    for editing metadata, classifications, and complaints. */
```
aslagle: This comment needs an update.

src/components/EntryPointsTabContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+export class EntryPointsTabContainer extends React.Component<EntryPointsTabContainerProps, any> {
```
aslagle: The state type can be void here.

src/components/EntryPointsTabContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+  render(): JSX.Element {
+    const entryPoints = this.getEnabledEntryPoints(this.props.libraries);
+    const mapEntryPointsToSchema = {
+      "Book": {
+        value: "http://schema.org/EBook",
+        label: "eBooks",
+      },
+      "Audio": {
+        value: "http://bib.schema.org/Audiobook",
+        label: "Audio Books",
```
aslagle: I'd like to figure out a way to avoid hard-coding these but if we have to we could at least put it in one file and use it from both components.

src/components/EntryPointsTabContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+      },
+    };
+    const svgMediumTypes = {
+      "http://bib.schema.org/Audiobook":
+        <AudioHeadphoneIcon ariaHidden className="draggable-item-icon" title="Audio Headphones Icon" />,
+      "http://schema.org/EBook":
+        <BookIcon ariaHidden className="draggable-item-icon" title="Book Icon" />,
```
aslagle: are these draggable?

src/components/EntryPointsTabContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+      return null;
+    }
+
+    return (
+      <ul className="nav nav-tabs">
+        { entryPoints.map(entryPoint => {
+            const { value, label } = mapEntryPointsToSchema[entryPoint];
+            const activeClass = !!(this.props.activeValue === entryPoint) ?
+              "active" : "";
+            const url = `${this.props.homeLink}?entrypoint=${entryPoint}`;
```
aslagle: I thought this URL would come from the OPDS feed.

src/components/EntryPointsTabContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+    let library;
+    const libraryStr = this.props.library(this.props.collectionUrl, null);
+
+    libraries.forEach(lib => {
+      if (lib.short_name === libraryStr) {
+        library = lib;
+      }
+    });
+
+    return library.settings.enabled_entry_points || [];
```
aslagle: I also thought this would come from the OPDS feed for some reason.

aslagle: Yeah, I just looked at the feed and I think you should be able to get all this stuff from the "facetGroups" in CollectionData.

src/components/EntryPointsTabContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+function mapDispatchToProps(dispatch) {
+  let fetcher = new DataFetcher({ adapter });
+  let actions = new ActionCreator(fetcher);
+  return {
+    fetchLibraries: () => dispatch(actions.fetchLibraries()),
+  };
+}
+
+const ConnectedEntryPointsTabContainer = connect<any, any, any>(
```
aslagle: This should be `connect<EntryPointsTabContainerStateProps, EntryPointsTabContainerDispatchProps, EntryPointsTabContainerOwnProps>`.

You didn't define the StateProps, that's the type returned by `mapStateToProps`.

src/components/__tests__/EntryPointsTabContainer-test.tsx
aslagle: I think it would be good enough to test that the `CatalogLink` has the right props, and this component renders correctly with any props. The unit tests for `CatalogLink` should verify that it calls the router. That said, since you're using `mount` instead of `shallow` it shouldn't matter that the link is a `CatalogLink` instead of an `a`.

src/stylesheets/entry_point_tabs.scss
```diff
@@ -0,0 +1,36 @@
+.entry-points-tab-container {
+  margin: 10px 0 15px;
+  padding: 0 15px;
+
+  .nav-tabs {
+    border-bottom: 1px solid #000;
```
aslagle: I'd use one of the variables from opds-web-client for the colors in here.

src/components/EntryPointsTabContainer.tsx
EdwinGuzman: Are `facetGroups` the same as `entrypoints`? Even before entry points were set up I think I remember seeing both "Books" and "Audio" in the facets group. The entry points are also set up per library. I'm not sure if that makes a difference or not?

src/components/EntryPointsTabContainer.tsx
EdwinGuzman: So `EntryPointsTabContainerStateProps` should contain `libraries` since that value is the one being used in `mapStateToProps`?

src/components/EntryPointsTabContainer.tsx
aslagle: There are some facet groups that aren't entry points, but you can identify the ones that are by the `simplified:facetGroupType="http://librarysimplified.org/terms/rel/entrypoint"`. I'm pretty sure "Books" and "Audio" were added for entry points, even if they were put in before entry points were implemented. 

The feed should contain the correct ones for the current library. If it doesn't, that should be fixed on the server.

src/components/EntryPointsTabContainer.tsx
aslagle: Yeah, exactly.

src/components/EntryPointsTabContainer.tsx
aslagle: You might have to modify the data adapter to handle that facet group type - it's a new thing we added for entry points so it might not be getting parsed yet.

src/components/EntryPointsTabContainer.tsx
EdwinGuzman: Just realized that it's not needed in the other `EntryPointsContainer` component so it'll only be used in this component.

src/components/EntryPointsTabContainer.tsx
EdwinGuzman: I'm not completely sure I get this, but I should update the adapter from `opds-web-client` to be able to parse that correctly and fetch that data in this component?

src/components/EntryPointsContainer.tsx
EdwinGuzman: This is the collection data that is being passed from `root` in the `opds-web-client`?

src/components/EntryPointsTabContainer.tsx
EdwinGuzman: Okay, I see what you mean. It seems that `facetGroups` and `links` (siblings in `collectionData`) both have the data I want. The links says that the link I want is of type `http://opds-spec.org/facet`. The `facetsGroups` array doesn't have any attributes stating the `facetGroupType`. It's possible that I could encounter a facet that is _not_ an entrypoint, and so that's why I should update the adapter?

aslagle: Yes, the grouped feeds don't have any other facets but the paged feeds have facets for sorting and filtering. However, I don't think you should update the adapter in `opds-web-client` since this custom attribute isn't part of OPDS. I was thinking you could update the editorAdapter in here but now that I look at it it's only for a book, not a collection. Not sure if the facets have raw data you can grab the attribute from - if not, that might be the thing to add in `opds-web-client`.

src/components/EntryPointsContainer.tsx
aslagle: Yeah, I think you could also get this from the facets.

src/components/EntryPointsTabContainer.tsx
EdwinGuzman: I'm still a bit confused about this, especially now that I decided to remove "Audio" from the entrypoints list in my library's configuration. So now I only have "Books" enabled but the `facetGroups` still has "Books" and "Audio" as facets. So if I use the values from the `facetGroups`, it won't match the entry points that were enabled in the admin settings.

aslagle: That sounds like the facet groups from the server are wrong. It's possible you had a cached feed that didn't update after you changed the settings.

EdwinGuzman: We might to link the icon for each item in a lane. Let me try using `facetGroups` and see how that goes.

EdwinGuzman:
@aslagle to go along with the last update on `opds-web-client` [#216](https://github.com/NYPL-Simplified/opds-web-client/pull/216), I am now using just the `facetGroups` from `collectionData` to populate the entrypoint tabs.

src/components/EntryPointsTabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+  AudioHeadphoneIcon,
+  BookIcon,
+} from "@nypl/dgx-svg-icons";
+
+export interface EntryPointsTabsContext {
+  pathFor: PathFor;
+  router: any;
+}
+export interface EntryPointsTabsProps {
+  formats?: FacetGroupData;
```
aslagle: This could be `FacetData[]` instead so you don't have to pass in the empty label.

src/components/EntryPointsTabs.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+    let entryPoints = this.props.formats ? this.props.formats.facets : [];
+    if (!entryPoints.length) {
+      return null;
+    }
+
+    return (
+      <ul className="nav nav-tabs">
+        { entryPoints.map(entryPoint => {
+            const { value, label } = mapEntryPointsToSchema[entryPoint.label];
```
aslagle: I think you should use `entryPoint.label` instead of mapping it to another label. If those labels aren't what you want, you should change them on the server.

src/components/__tests__/EntryPointsContainer-test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  let wrapper;
+
+  it("should not render if no facets were passed", () => {
+    facetGroups = [];
+
+    wrapper = shallow(<EntryPointsContainer facetGroups={facetGroups} />);
+    expect(wrapper.html()).to.equal(null);
+  });
+
+  it("shows a tab container with Audio selected value", () => {
```
aslagle: The title of this test doesn't seem right.

aslagle:
Found a bug on the server while testing this. I enabled the book and audio entrypoints but not "All", and I don't actually have any audiobooks. When I go to the feed for audio both format facets have the audio URL. So in the admin interface the ebook tab doesn't work.

src/stylesheets/entry_point_tabs.scss
```diff
@@ -0,0 +1,36 @@
...
+        border-bottom-color: transparent;
+      }
+    }
+
+    a {
+      min-width: 100px;
+      text-align: center;
+    }
+
+    svg {
```
aslagle: The SVG should probably be the same color as the text (right now it is for active, but the nonactive text is the link color).

aslagle:
I don't know what to do about this, but if the library doesn't have any lanes, it will show the format links in the left sidebar as well as the tabs.

aslagle:
Another server bug: when I enabled the "All" entry point, the feeds for the books and audio entry points don't have the "All" facet.

aslagle:
When I only enable the "All" entry point, the feeds have all three format facets, but the book and audio feeds ignore the entry point in the URL.

I get why you weren't using the facets before, but I think they should be fixed rather than abandoned :)

EdwinGuzman:
>When I go to the feed for audio both format facets have the audio URL. So in the admin interface the ebook tab doesn't work. 
- This seems like a bug in `circulation`, possibly?

>I don't know what to do about this, but if the library doesn't have any lanes, it will show the format links in the left sidebar as well as the tabs.
I don't know what to do about that either and @ricardoom was confused about it, too. I actually see it even if I do have lanes, but not all the time. In general, I'm a bit confused about _when_ it should be displaying. Maybe it's not needed if I'm making tabs instead?

>Another server bug: when I enabled the "All" entry point, the feeds for the books and audio entry points don't have the "All" facet.
I think not having an entrypoint (or not having it explicitly say `?entrypoint=All`) is the same as `?entrypoint=All` but I'll double check.

>When I only enable the "All" entry point, the feeds have all three format facets, but the book and audio feeds ignore the entry point in the URL.
Can you further explain this?

aslagle:
> When I go to the feed for audio both format facets have the audio URL. So in the admin interface the ebook tab doesn't work.
> 
> This seems like a bug in circulation, possibly?

Yeah, I think it's a bug in circulation.

> I don't know what to do about this, but if the library doesn't have any lanes, it will show the format links in the left sidebar as well as the tabs.
> I don't know what to do about that either and @ricardoom was confused about it, too. I actually see it even if I do have lanes, but not all the time. In general, I'm a bit confused about when it should be displaying. Maybe it's not needed if I'm making tabs instead?

It's because the sidebar is how we normally display facets, but these facets are special since we're already displaying them in the tabs. I think it would be good to remove the formats from the left sidebar here, but I don't know how you'd do it.

> When I only enable the "All" entry point, the feeds have all three format facets, but the book and audio feeds ignore the entry point in the URL.
> Can you further explain this?

I think this is another circulation bug. I enabled "all", "ebooks", and "audio" as entry points, and there were tabs for all three, but going to "ebooks" or "audio" didn't change the feed, i got the same books on all the tabs.

EdwinGuzman:
After briefly talking to @leonardr, I think we do want to use the entrypoints from `library.settings.enabled_entry_points` instead of the data from `facetGroups`. That would mean not displaying the facets on the left sidebar for these entrypoints and just using the tabs.

>It's because the sidebar is how we normally display facets, but these facets are special since we're already displaying them in the tabs. I think it would be good to remove the formats from the left sidebar here, but I don't know how you'd do it.

It doesn't seem too difficult to remove the whole left sidebar but that might mean removing other facets which could be important (I'm not sure what other facets are available).

Since I was generating the urls for the tabs when using the `enabled_entry_points` data, I did not previously encounter the two bugs you mentioned. I do see that the facet URL is wrong in the facets sidebar (besides in the tabs) so that may be a bug to fix regardless of what data source is decided to use.

aslagle:
Using `library.settings.enabled_entry_points` instead of the data from facetGroups wouldn't affect the left sidebar - it would change how you're generating the tabs but unless you remove the facet group from the feed entirely the left sidebar would still be there.

I think the facet group will be used by the mobile apps, and I would expect the admin interface to match the mobile apps. I do agree that there are server bugs to fix regardless of what you use.

src/components/EntryPointsTabs.tsx
EdwinGuzman: Updated.

src/components/__tests__/EntryPointsContainer-test.tsx
EdwinGuzman: Updated.

src/stylesheets/entry_point_tabs.scss
EdwinGuzman: Fixed.

EdwinGuzman:
@aslagle I added two PRs to circulation [#993](https://github.com/NYPL-Simplified/circulation/pull/993) and core [#901](https://github.com/NYPL-Simplified/server_core/pull/901) to help fix the entry point facet link issues. 

>I think this is another circulation bug. I enabled "all", "ebooks", and "audio" as entry points, and there were tabs for all three, but going to "ebooks" or "audio" didn't change the feed, i got the same books on all the tabs.

The above should be fixed now, as well as when you go further down into a lane in an entrypoint.
Also, the issue with the "All" entrypoint tab not having the correct url should be fixed, too.

I believe it's just down to:
>I don't know what to do about this, but if the library doesn't have any lanes, it will show the format links in the left sidebar as well as the tabs.

I can remove the sidebar if the facets are entrypoints.

src/components/EntryPointsTabs.tsx
aslagle: This comment still applies.

EdwinGuzman: Address in https://github.com/NYPL-Simplified/server_core/pull/901/commits/3d620c94d3d97544428a839d9970415ddf2622a5 and will be updating it here soon.

EdwinGuzman:
@aslagle updated the PR to update the map of entrypoint display names so it comes from the server, and updated `EntryPointsContainer` so that it wraps around the `Collection` component from `opds-web-client`. There's a separate PR for that.

EdwinGuzman:
Tests fail because the `opds-web-client` needs to be updated first.

package.json
aslagle: The version number needs another update now.

aslagle:
Made one last minor comment but other than that it looks good. Thanks for all your work on this!

EdwinGuzman:
Updated the package version and used a different way to deep copy. Thanks!

src/components/EntryPointsContainer.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+/** Wrapper for `EntryPointsTabs`. This component is passed into the
+    OPDSCatalog from opds-web-client. */
+export default class EntryPointsContainer extends React.Component<CollectionContainerProps, void> {
+  render(): JSX.Element {
+    const child = React.Children.only(this.props.children);
+    const collectionCopy = JSON.parse(JSON.stringify(child.props.collection));
+    const facetGroups = collectionCopy && collectionCopy.facetGroups ?
+      collectionCopy.facetGroups : [];
+    const newProps = JSON.parse(JSON.stringify(child.props));
```
aslagle: This might not work if some of the props are functions. It should be fine for the collection though.

EdwinGuzman: You're right. I changed it back and updated how `newProps` for the `Collection` component added. The `facetGroups` for child.props and newProps are each their own now.

PR #160, opened 2018-06-25
Updated dependencies because of vulnerability in node-growl 1.9.2.

aslagle:
Github told me about a security vulnerability in one of our dependencies so I updated a couple versions to fix it. One of them was nightwatch, and the selenium tests already don't work, so I can't tell if the update made anything worse. I made an [epic](https://jira.nypl.org/browse/SIMPLY-987) to fix those tests since they need major work.


src/components/__tests__/CirculationEvents-test.tsx
```diff
@@ -56,6 +56,10 @@ describe("CirculationEvents", () => {
       );
     });
 
+    afterEach(() => {
+      wrapper.instance().componentWillUnmount();
+    });
```
aslagle: In the newer version of mocha, the test command doesn't automatically exit if there's still something running. Unmounting this component stops a timer that would otherwise keep the tests from exiting.

src/components/__tests__/CirculationEvents-test.tsx
```diff
@@ -142,13 +150,11 @@ describe("CirculationEvents", () => {
...
-          expect(fetchAndQueueSpy.callCount).to.equal(1);
-          fakeTimer.tick(1000);
-          expect(fetchAndQueueSpy.callCount).to.equal(2);
-          done();
-        }).catch(err => { console.log(err); throw(err); });
+      it("sets timeout for fetches", async () => {
+        await wrapper.instance().fetchAndQueue();
+        expect(fetchAndQueueSpy.callCount).to.equal(1);
+        fakeTimer.tick(1000);
+        expect(fetchAndQueueSpy.callCount).to.equal(2);
```
aslagle: This change doesn't affect anything, but I like the async/await syntax better :)

EdwinGuzman: I agree. I can't seem to get this syntax to work on the other repos I've worked on but do prefer it.

aslagle: It's ES7 I think. It works here because Typescript supports it.

PR #161, opened 2018-06-28
Improved ambiguous and overly long message.

aslagle:
It was unclear from the previous message whether it would resolve all complaints of this type for the book or across the library, and it didn't include the complaint type so you wouldn't know if you accidentally clicked the wrong button. I also took out the "Are you sure" since I've heard it's better to have shorter messages.

### DIFF
--- a/src/components/Complaints.tsx
+++ b/src/components/Complaints.tsx
@@ -122,7 +122,7 @@ export class Complaints extends React.Component<ComplaintsProps, void> {
     return this.props.bookUrl.replace("works", "admin/works") + "/resolve_complaints";
   }
 
-  readableComplaintType(type) {
+  readableComplaintType(type: string) {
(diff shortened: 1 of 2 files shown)


EdwinGuzman:
Where do these complaints show up?

src/components/Complaints.tsx
```diff
@@ -122,7 +122,7 @@ export class Complaints extends React.Component<ComplaintsProps, void> {
     return this.props.bookUrl.replace("works", "admin/works") + "/resolve_complaints";
   }
 
-  readableComplaintType(type) {
+  readableComplaintType(type: string) {
```
EdwinGuzman: Just curious, why didn't Typescript complain about this?

aslagle: Typescript doesn't require you to give everything a type. It will attempt to infer it if there's no explicit type.

PR #162, opened 2018-07-19
Changed dashboard to display library-specific stats as well as sitewide, and show all collections in a chart.

aslagle:
I also changed the server code to generate the dashboard data in https://github.com/NYPL-Simplified/circulation/pull/997.


src/components/DashboardPage.tsx
```diff
@@ -5,28 +5,44 @@ import Stats from "./Stats";
...
+
   render(): JSX.Element {
     return (
       <div className="dashboard">
         <Header />
         <div className="body">
-          <Stats store={this.context.editorStore} />
-          <CirculationEvents store={this.context.editorStore} />
+          <Stats store={this.context.editorStore} library={this.props.params.library}/>
+          { !this.props.params.library &&
```
EdwinGuzman: "Circulation Events" displays the current list of items being read, loaned, downloaded, etc. Why don't you want to display this when there are libraries? Is it a general list of events for all libraries?

aslagle: The circulationevents table currently doesn't track the library, so there's no way to display the list for a specific library unless we change that. It might be worth doing, but it's a separate thing than the rest of this PR. I don't think the list of events is very useful and I'll probably want to get rid of it once we have better things to put in the dashboard, but it still might be useful to track the library for events.

src/components/LibraryStats.tsx
```diff
@@ -0,0 +1,105 @@
...
+/** Displays statistics about patrons, licenses, and collections from the server,
+    for a single library or all libraries the admin has access to. */
+export default class LibraryStats extends React.Component<LibraryStatsProps, void> {
+  render(): JSX.Element {
+    const collectionCounts = this.props.stats && this.props.stats.collections && Object.keys(this.props.stats.collections).map(collection => {
+      const data = this.props.stats.collections[collection];
+      return {
+        label: collection,
+        // The "Titles" key is displayed in the default recharts tooltip.
+        Titles: data.licensed_titles + data.open_access_titles
```
EdwinGuzman: Would it be useful to also see this displayed in the chart? So that the admin knows that the number of "Titles" is a combination of licensed titles and open access titles.

src/components/LibraryStats.tsx
```diff
@@ -0,0 +1,105 @@
...
+                  <span className="stat-value">{this.formatNumber(this.props.stats.patrons.with_active_loans)}</span>
+                  <span className="stat-label">Patrons With Active Loans</span>
+                </div>
+                <div className="single-stat">
+                  <span className="stat-value">{this.formatNumber(this.props.stats.patrons.with_active_loans_or_holds)}</span>
+                  <span className="stat-label">Patrons With Active Loans or Holds</span>
+                </div>
+              </div>
+            }
+            { this.props.stats.patrons.total > 0 &&
```
EdwinGuzman: Are you not combining these two checks so that they can each be in its own div?

src/components/LibraryStats.tsx
```diff
@@ -0,0 +1,105 @@
...
+                  <div className="stat-value">{this.formatNumber(this.props.stats.patrons.loans)}</div>
+                  <div className="stat-label">Active Loans</div>
+                </div>
+                <div className="single-stat">
+                  <div className="stat-value">{this.formatNumber(this.props.stats.patrons.holds)}</div>
+                  <div className="stat-label">Active Holds</div>
+                </div>
+              </div>
+            }
+            <div className="stat-group">
```
EdwinGuzman: I think these might be better as a list so you can tab through the list of values.

src/stylesheets/stats.scss
```diff
@@ -1,25 +1,49 @@
 .stats {
-  ul {
-    border-bottom: 1px solid black;
-  }
-
-  li {
-    text-align: center;
-    margin: 10px;
+  display: grid;
```
EdwinGuzman: Great use of grid.

src/components/LibraryStats.tsx
aslagle: Exactly, I didn't want an extra wrapper div. I think we'd have to update react to avoid that.

EdwinGuzman:
This works really great for me and I'm liking the charts. I'd just like that if I'm on the dashboard page and I select a different library, that it would stay on the dashboard page with the newly selected library. But, I see why that was done with the two `Dashboard` links in the header.

src/stylesheets/stats.scss
aslagle: this was from mauricio :)

src/components/LibraryStats.tsx
EdwinGuzman: Yes, that's very annoying.

aslagle:
I've gone back and forth on what should happen when you change libraries in the drop down. 

If you've navigated at all from a catalog page, we have to go back to the top since the thing you're on may not exist in the other library. And since we have to do that, it seems like the other pages should do the same to be consistent.

EdwinGuzman:
And version bump for this.

PR #164, opened 2018-07-27
Self test admin

EdwinGuzman:
This is functional but I have not written any tests (this is still a WIP). Do you mind checking the redux action/reducer flow before I go any further? This is part of `circulation` branch `self-test-controller`.

The results for each collection render and running a new self test for individual collections and re-rendering those results work. But I do see weirdness when I run self tests again for Axis 360 and then decide to run self tests for RBDigital and the rendered results for Axis 360 revert to the previous self test results. I think it may be because of how I'm making the state and props.


src/components/Libraries.tsx
```diff
@@ -11,6 +11,7 @@ import LibraryEditForm from "./LibraryEditForm";
     editing or deleting an existing library. */
 export class Libraries extends EditableConfigList<LibrariesData, LibraryData> {
   EditForm = LibraryEditForm;
+  AdditionalContent = null;
```
aslagle: You could probably set this to null in EditableConfigList so you don't have to define it in every subclass. Or even better, have the server tell the client which things have self-tests as part of the GET requests and dynamically decide whether to show the self-tests.

src/components/SelfTests.tsx
```diff
@@ -0,0 +1,137 @@
...
+  XIcon,
+} from "@nypl/dgx-svg-icons";
+
+export interface SelfTestsStateProps {
+  item: ServiceData;
+}
+
+export interface SelfTestsDispatchProps {
+  runSelfTests?: (identifier: string | number) => Promise<void>;
+  getSelfTests?: (identifier: string | number) => Promise<void>;
```
aslagle: Seems weird that this one is void, I'd have thought it would return the results.

aslagle:
I'll try to explain what I think your problem is. The collections action/reducer returns the self test results for all collections. The self tests action reducer returns the self test results for the current collection. When you run the self tests action reducer again, it replaces the previous self test result for Axis 360, so Axis 360 goes back to using whatever came from the collections action.

You could fix this by making another GET request for all collections after running self tests, but it's not very efficient. That's sort of the pattern that's currently used in the repo.

You could also make the collections reducer handle the self tests action and modify its part of the store, so the result gets updated in the set of all collections. Since it uses the fetchEditReducer, this would likely require modifying fetchEditReducer so you could pass in additional actions and functions for handling them.

EdwinGuzman:
@aslagle thanks for that feedback. Updated that reducer and I'm liking how much cleaner and generic it is now.

I think I am done now with this part. Will create a PR for the circulation manager soon.

.travis.yml
```diff
@@ -2,3 +2,9 @@ language: node_js
 
 node_js:
   - "7"
+
+before_script:
```
EdwinGuzman: Added this so that the rest result time display will be consistent in Travis.

aslagle: I get the time zone part but why do you need xvfb?

src/components/SelfTests.tsx
```diff
@@ -0,0 +1,152 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = { expand: false };
+    this.toggleView = this.toggleView.bind(this);
+    this.runSelfTests = this.runSelfTests.bind(this);
+  }
+
+  render() {
+    const collection = this.props.item;
```
aslagle: Looks like everything in here would already work fine if the item was something other than a collection - maybe just rename this variable.

src/components/SelfTests.tsx
```diff
@@ -0,0 +1,152 @@
...
+    const expand = this.state.expand;
+
+    if (!collection || !collection.self_test_results) {
+      return null;
+    }
+    const date = new Date(collection.self_test_results.start);
+    const startDate = date.toDateString();
+    const hours = ("0" + date.getHours()).slice(-2);
```
aslagle: Does this do something different than `String(date.getHours())`?

src/components/SelfTests.tsx
```diff
@@ -0,0 +1,152 @@
...
+  async runSelfTests(e, identifier: string | number) {
+    e.preventDefault();
+    await this.props.runSelfTests(identifier);
+    this.props.getSelfTests(identifier);
+  }
+}
+
+function mapStateToProps(state, ownProps) {
+  const stateResults = state.editor.selfTests &&
+    state.editor.selfTests.data && state.editor.selfTests.data.collection;
```
aslagle: Do you still need this given your changes to the reducers? You might want to get the `isFetching` from `state.editor.selfTests` so you can show something when the test is running, but you should always have the self test results in the `item` now.

src/components/SelfTests.tsx
```diff
@@ -0,0 +1,152 @@
...
+}
+
+function mapDispatchToProps(dispatch, ownProps) {
+  let fetcher = new DataFetcher();
+  let actions = new ActionCreator(fetcher, ownProps.csrfToken);
+  const collectionId = ownProps.item.id;
+
+  return {
+    getSelfTests: (identifier: string | number) => dispatch(actions.getSelfTests(collectionId)),
+    runSelfTests: (identifier: string | number) => dispatch(actions.runSelfTests(identifier))
```
aslagle: If you're passing `collectionId` to the actions here, these props do not need `identifier` as an argument.

src/reducers/__tests__/createFetchEditReducer-test.ts
```diff
@@ -152,4 +164,23 @@ describe("fetch-edit reducer", () => {
...
+    expect(extraActionReducer(initState, action)).to.deep.equal(newState);
+  });
+
+  it("handles extra action data load", () => {
+    let action = { type: "TEST_EXTRA_LOAD", url: "test url" };
+    let newState = Object.assign({}, initState, {
+      isLoaded: false,
+    });
+
+    expect(extraActionReducer(initState, action)).to.deep.equal(newState);
```
aslagle: I feel like this test should be checking that `manipulateData` was called. Shouldn't the manipulated data be in the new state?

src/reducers/collections.ts
```diff
@@ -2,4 +2,49 @@ import { CollectionsData } from "../interfaces";
...
+      collection.self_test_results = action.data.collection.self_test_results;
+    }
+  });
+
+  return Object.assign({}, state, {
+    data: data,
+    isFetching: false,
+    isLoaded: true,
+  });
+};
```
aslagle: I don't see any tests for this yet.

src/reducers/collections.ts
```diff
@@ -2,4 +2,49 @@ import { CollectionsData } from "../interfaces";
...
+  const data = state.data;
+  if (!data.collections.length) {
+    return {};
+  }
+
+  data.collections.forEach(collection => {
+    if (collection.id === action.data.collection.id) {
+      collection.self_test_results = action.data.collection.self_test_results;
+    }
+  });
```
aslagle: I think you're mutating the data in the old state here and redux [recommends against that](https://redux.js.org/faq/immutabledata). You are still returning a new object below, so I'm not sure if this would actually cause any problems.

src/reducers/collections.ts
```diff
@@ -2,4 +2,49 @@ import { CollectionsData } from "../interfaces";
...
+
+const successCB = (state, action) => {
+  return Object.assign({}, state, {
+    fetchError: null
+  });
+};
+
+const extraActions = {
+  [`${ActionCreator.GET_SELF_TESTS}_${ActionCreator.REQUEST}`]: requestCB,
+  [`${ActionCreator.GET_SELF_TESTS}_${ActionCreator.SUCCESS}`]: successCB,
```
aslagle: I don't understand why you have request and success in here. Is that just to make it easier to show the message when there's an error?

src/reducers/collections.ts
```diff
@@ -2,4 +2,49 @@ import { CollectionsData } from "../interfaces";
...
+    data: state.data,
+    isLoaded: false,
+    isFetching: true,
+    fetchError: null,
+  });
+};
+
+const successCB = (state, action) => {
+  return Object.assign({}, state, {
+    fetchError: null
```
aslagle: It seems wrong that if there was a previous error getting collections, a successful request for self tests would clear it out.

src/stylesheets/self_tests.scss
```diff
@@ -0,0 +1,83 @@
...
+
+
+        svg {
+          height: 30px;
+          width: 30px;
+          align-self: center;
+
+          &.success {
+            // $nypl-green-bright
+            fill: #008918;
```
aslagle: why didn't you use the variable?

src/reducers/collections.ts
EdwinGuzman: If I didn't include `REQUEST` or `SUCCESS`, then the data would not automatically update on the page (unless I clicked the button to show the results which I imagine meant that the component was not rerendering). If I included those actions, I would see the loading component and the individual data for a collection would update.

src/components/SelfTests.tsx
EdwinGuzman: Sometimes `date.getMinutes()`/`date.getSeconds()` would return `4` instead of `04`, so it added an extra 0 for padding.

.travis.yml
EdwinGuzman: I don't. It was part of a different travis configuration that also had timezone configured and I forgot to remove it.

src/actions.ts
```diff
@@ -612,4 +615,14 @@ export default class ActionCreator extends BaseActionCreator {
     const url = "/admin/change_password";
     return this.postForm(ActionCreator.CHANGE_PASSWORD, url, data).bind(this);
   }
+
+  getSelfTests(identifier: string | number) {
+    const url = `/admin/collection_self_tests/${identifier}`;
```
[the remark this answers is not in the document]
EdwinGuzman: Why would the reducer part be tricky? It doesn't seem like the `identifier` param is checked in the reducer so how would the url, as a param, affect it?

src/components/EditableConfigList.tsx
```diff
@@ -31,14 +31,20 @@ export interface EditableConfigListProps<T> extends EditableConfigListStateProps
...
-  data: T;
-  disabled: boolean;
-  editItem: (data: FormData) => Promise<void>;
-  urlBase: string;
-  listDataKey: string;
+  data?: T;
+  disabled?: boolean;
+  editItem?: (data: FormData) => Promise<void>;
+  urlBase?: string;
+  listDataKey?: string;
```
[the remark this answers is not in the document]
EdwinGuzman: I'll undo. I previously made them optional when I was trying to figure out how to update and add the new optional component.

src/components/SelfTests.tsx
EdwinGuzman: Rename which one, `collection` or `item`? I guess `collection` since not only collections will have self test results.

src/reducers/collections.ts
aslagle: That's weird, I would think if they weren't here it would not show the loading message but it would still rerender the new data when it was loaded.

I think it's good to keep the isFetching and error info in the store separate for different types of requests. I'd probably remove these but have the component's mapStateToProps use isFetching and error from the self tests reducer as well as the collections reducer. Only the loaded data would be automatically in the collections state. The data is better to have in the reducer because it enables you to update results for multiple individual collections, but I don't think that applies for isFetching/error since there's only one of that in the state.

I'm not sure about this though, maybe it is better to handle as much as possible in reducers.

src/actions.ts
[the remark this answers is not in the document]
aslagle: Because there are separate reducers for each type of external integration, and they will all need to handle the action. You've got it in the collections reducer right now. I guess you could make another createReducer method that creates a fetchEdit reducer with the extra action and use it for multiple things.

src/components/SelfTests.tsx
aslagle: yeah, `collection`

EdwinGuzman: I'm setting the `SelfTestsStateProps` to `ServiceData`. Will that be a problem later for another object/integration that is not a collection?

aslagle: I think that's fine, ServiceData is already used for other external integrations as well.

src/components/SelfTests.tsx
EdwinGuzman: Cool. You're right, I don't need this anymore. I still don't fully understand how this works so I'll redo some of action/reducers/data work for this component.

src/components/SelfTests.tsx
```diff
@@ -0,0 +1,166 @@
...
+      <div className="integration-selftests">
+        <div>
+          {resultIcon}
+          <p className="description">Tests last ran on {startDate} {startTime} and lasted {duration}s.</p>
+          <button onClick={this.toggleView} className="btn btn-default">{resultsLabel} Results</button>
+        </div>
+        <div className={`results collapse ${expandResultClass}`}>
+          <h3>Self Test Results</h3>
+          {isFetching &&
+            <span>Running new self tests</span>
```
EdwinGuzman: @aslagle let me know what you think of this message as well as the `colorResultClass` below which turns the integration's list of self tests gray while new tests and results are being fetched before updating.

src/reducers/collections.ts
EdwinGuzman: Now I'm deep copying the state and returning the new object.

src/components/Collections.tsx
```diff
@@ -18,13 +19,9 @@ export interface CollectionsProps extends CollectionsStateProps, CollectionsDisp
 
 class CollectionEditForm extends ServiceWithRegistrationsEditForm<CollectionsData> {};
 
-/** Right panel for collections on the system configuration page.
-    Shows a list of current collections and allows creating a new
-    collection or editing or deleting an existing collection. 
-    Also allows registering libraries with the collection when
-    the collection supports it. */
```
aslagle: Why did you take out the comment?

src/components/EditableConfigList.tsx
```diff
@@ -62,8 +69,9 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
 
   render(): JSX.Element {
     let EditForm = this.EditForm;
+    let AdditionalContent = this.AdditionalContent || null;
     return (
-      <div>
+      <div className={`admin-${this.itemTypeName}`}>
```
aslagle: Might want to use a more generic class like `has-additional-content` to make it easier when you have self tests for other things.

src/components/SelfTests.tsx
aslagle: This looks good.

src/reducers/selfTests.ts
```diff
@@ -0,0 +1,30 @@
...
+  return Object.assign({}, state, {
+    fetchError: null,
+    isFetching: false,
+  });
+};
+
+const extraActions = {
+  [`${ActionCreator.GET_SELF_TESTS}_${ActionCreator.REQUEST}`]: requestCB,
+  [`${ActionCreator.GET_SELF_TESTS}_${ActionCreator.SUCCESS}`]: successCB,
+};
```
aslagle: I feel like you don't need extra actions in this reducer. Doesn't createFetchEditReducer already set up everything you need?

src/components/Collections.tsx
EdwinGuzman: I didn't mean to. I'll add it back in.

src/components/SelfTests.tsx
```diff
@@ -0,0 +1,166 @@
...
+      runTests: this.state.runTests,
+    });
+  }
+
+  async runSelfTests(e) {
+    e.preventDefault();
+    this.setState({
+      expand: this.state.expand,
+      runTests: true,
+    });
```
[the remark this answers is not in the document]
EdwinGuzman: No. That's a good point and I'll add that in.

src/reducers/selfTests.ts
EdwinGuzman: Yeah, it does give me those already.

src/reducers/collections.ts
EdwinGuzman: There are tests for this now in `src/reducers/__tests__/collections-tests.ts`.

src/components/SelfTests.tsx
```diff
@@ -0,0 +1,170 @@
...
+        </div>
+        <div className={`results collapse ${expandResultClass}`}>
+          <h3>Self Test Results</h3>
+          {isFetching &&
+            <span>Running new self tests</span>
+          }
+          <button
+            onClick={(e) => this.runSelfTests(e)}
+            className="btn btn-default runSelfTests"
+            disabled={this.props.isFetching}
```
EdwinGuzman: Disabled all run test buttons when one is currently running. It should cover your comment about simultaneously running tests.

src/actions.ts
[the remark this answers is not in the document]
EdwinGuzman: Updated. I only updated the actions and not the reducers and everything seemed to work the same as before.

src/components/SelfTests.tsx
```diff
@@ -0,0 +1,183 @@
...
+    let duration;
+
+    if (integration.self_test_results) {
+      date = new Date(integration.self_test_results.start);
+      startDate = date.toDateString();
+      hours = ("0" + date.getHours()).slice(-2);
+      minutes = ("0" + date.getMinutes()).slice(-2);
+      seconds = ("0" + date.getSeconds()).slice(-2);
+      startTime = `${hours}:${minutes}:${seconds}`;
+      results = integration.self_test_results && integration.self_test_results.results;
```
aslagle: You don't need to check `integration.self_test_results` here now that you have the `if`.

src/components/SelfTests.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+    return (
+      <div className="integration-selftests">
+        <div>
+          {results.length ? resultIcon : null}
+          <p className="description">{testDescription}</p>
+          <button onClick={this.toggleView} className="btn btn-default">{resultsLabel} Results</button>
+        </div>
+        <div className={`results collapse ${expandResultClass}`}>
+          <h3>Self Test Results</h3>
```
aslagle: I would make this smaller - it's bigger than the collection names above it.

src/components/SelfTests.tsx
```diff
@@ -0,0 +1,203 @@
...
+          error: null,
+        });
+      })
+      .catch(error => {
+        this.setState({
+          expand: this.state.expand,
+          runTests: false,
+          error: error,
+        });
+      });
```
aslagle: You can use the `await` syntax inside a regular try-catch instead of callbacks like this.

EdwinGuzman: Updated. Thanks!

PR #166, opened 2018-08-23
Manage patrons admin

adriana-alter:


src/components/PatronInfo.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from "react";
+import { PatronData } from "../interfaces";
+
+export interface PatronInfoProps {
+  patron: PatronData;
+}
+
+export default class PatronInfo extends React.Component<PatronInfoProps, void> {
```
EdwinGuzman: @adriana-alter you don't need `OwnProps` if it's a simple component like this. The state is also `void` instead of `{}`.

src/components/ResetAdobeId.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+export interface PatronActionsListOwnProps {
+  store?: Store<State>;
+  csrfToken?: string;
+  patron: PatronData;
+}
+
+export interface PatronActionsListProps extends React.Props<PatronActionsListProps>, PatronActionsListStateProps, PatronActionsListDispatchProps, PatronActionsListOwnProps {}
+
+export interface PatronActionsListState {
```
EdwinGuzman: This is the file where we didn't update the name of these interfaces to reflect that it's now `ResetAdobeId`. That should be done.

src/components/__tests__/ManagePatronsForm-test.tsx
```diff
@@ -0,0 +1,150 @@
...
+        />
+      );
+
+      let errorMessage = wrapper.find(ErrorMessage);
+      expect(errorMessage.length).to.equal(0);
+
+      let button = wrapper.find("button");
+
+      // TODO: make this work
+      // await button.simulate("submit");
```
EdwinGuzman: @aslagle we had trouble trying to figure out why this did not work. So instead of simulating the submit, we set the props. There's another similar test where we did set the state instead of the props, but still the same idea of not getting `simulate` to work. Can you help us figure it out?

adriana-alter: Note: I just spent about an hour Googling and experimenting.  I have discovered that 1) this is a really common problem--maybe even a bug in this version of Enzyme--and 2) none of the solutions proposed on Stack Overflow etc., even the ones that sound really plausible (e.g. mocking preventDefault), work.  I'm not sure how worthwhile it is to sink a lot more time into this, particularly given that the way we're currently doing it does work...

src/actions.ts
```diff
@@ -171,7 +182,11 @@ export default class ActionCreator extends BaseActionCreator {
             if (response.text) {
               response.text().then(text => {
                 dispatch(this.load<string>(type, text));
-                resolve(response);
+                if (returnType === "JSON") {
+                  resolve({ response, text: JSON.parse(text) });
```
aslagle: You can also call `response.json` instead of `response.text` if you're expecting JSON. I'm not sure if that would have any advantages here or not. 

It would be nice if the JSON and text return types were handled consistently though, so it would probably be better for the load event to get the parsed JSON instead of the text, and for the arguments to resolve to match.

src/components/Header.tsx
```diff
@@ -115,6 +115,11 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
               <li>
                 <Link to={"/admin/web/dashboard/" + this.context.library()}>Dashboard</Link>
               </li>
+              { this.context.admin.isSystemAdmin() &&
```
aslagle: Is this only meant for system admins? I think at least library managers should be able to use it.

src/components/ManagePatrons.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Store } from "redux";
+import { connect } from "react-redux";
+import { State } from "../reducers/index";
+import ActionCreator from "../actions";
+import { FetchErrorData } from "opds-web-client/lib/interfaces";
+import { PatronData } from "../interfaces";
+import ButtonForm from "./ButtonForm";
+import EditableInput from "./EditableInput";
+import ErrorMessage from "./ErrorMessage";
+import { Alert } from "react-bootstrap";
```
aslagle: You have some unused imports here.

src/components/Header.tsx
EdwinGuzman: Hmm we've been using "Library Admin" as the term when discussing it and that may not be a term that maps to what we are using here. Library managers sounds correct, too.

aslagle: Everyone who can use the admin interface at all is a "library admin". The terms I've been using for admin types are "system admin", "library manager", and "librarian". Whatever's done here should also be enforced on the server.

src/components/ManagePatronsTabContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+  context: ManagePatronsTabContainerContext;
+  static contextTypes: React.ValidationMap<ManagePatronsTabContainerContext> = {
+    router: React.PropTypes.object.isRequired,
+    pathFor: React.PropTypes.func.isRequired,
+    admin: React.PropTypes.object.isRequired
+  };
+
+  tabs() {
+    const tabs = {};
+    if (this.context.admin.isSystemAdmin()) {
```
aslagle: This also should probably be available to library managers or even librarians.

src/components/ResetAdobeId.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { connect } from "react-redux";
+import { State } from "../reducers/index";
+import ActionCreator from "../actions";
+import { FetchErrorData } from "opds-web-client/lib/interfaces";
+import { PatronData } from "../interfaces";
+import { Alert } from "react-bootstrap";
+import EditableInput from "./EditableInput";
+import PatronInfo from "./PatronInfo";
+
+export interface PatronError {
```
aslagle: Doesn't look like this is being used.

src/components/ResetAdobeId.tsx
```diff
@@ -0,0 +1,138 @@
...
+  error: FetchErrorData;
+  success: boolean;
+  checked: boolean;
+}
+
+export class ResetAdobeId extends React.Component<ResetAdobeIdProps, ResetAdobeIdState> {
+  constructor(props) {
+    super(props);
+    this.state = {
+      error: { status: 200, response: "", url: "" },
```
aslagle: I'd make the error start as null.

src/components/ResetAdobeId.tsx
```diff
@@ -0,0 +1,138 @@
...
+    this.verifiedToReset = this.verifiedToReset.bind(this);
+  }
+
+  async resetAdobeId(e) {
+    e.preventDefault();
+    const data = new (window as any).FormData();
+    data.append("identifier", this.props.patron.authorization_identifier);
+
+    try {
+      const response = await this.props.resetAdobeId(data);
```
aslagle: Normally we'd use redux for this instead of waiting for the response and setting state in the component.

src/components/ResetAdobeId.tsx
```diff
@@ -0,0 +1,138 @@
...
+    catch (error) {
+      this.setState({
+        success: false,
+        error: error,
+        checked: false
+      });
+    }
+  }
+
+  verifiedToReset() {
```
aslagle: Since this method can also be used to uncheck the box I would change the name to something like 'toggle'.

src/components/ResetAdobeId.tsx
```diff
@@ -0,0 +1,138 @@
...
+    return (
+      <div className="patron-actions">
+        <h4>Reset Adobe ID</h4>
+        { this.state.error.status > 200 &&
+          <Alert bsStyle="danger">Error: failed to reset Adobe ID for patron {patron.authorization_identifier}</Alert>
+        }
+        { this.state.success &&
+          <Alert bsStyle="success">Adobe ID for patron {patron.authorization_identifier} has been reset.</Alert>
+        }
+        <p>This feature allows you to delete the existing Adobe ID for an individual patron; a new Adobe ID will be assigned automatically when the patron logs in again. This step is necessary when patrons reach their device installation limit. Please be sure to inform patrons that resetting their Adobe ID will cause them to lose any existing loans or reserves.</p>
```
aslagle: I don't think "reserves" is a term we use anywhere else. Should that be "reservations", or "holds"?

src/components/ResetAdobeId.tsx
```diff
@@ -0,0 +1,138 @@
...
+              <p className="patron-warning"><b>Patron {patron && (patron.username || patron.personal_name)} will lose any existing loans or reserves when the Adobe ID is reset.</b></p>
+              <EditableInput
+                type="checkbox"
+                name="resetAdobeId"
+                checked={this.state.checked}
+                label="Patron has been informed about this change"
+                onChange={this.verifiedToReset}
+                />
+              <button
+                className="btn btn-default"
```
aslagle: This could be `btn-danger` since it's a destructive action.

src/components/__tests__/ManagePatronsForm-test.tsx
aslagle: I think this probably doesn't work because you mocked the patronLookup function, and that's what normally handles the error and leads to the prop being set. Nothing is actually handling the error from patronLookup itself.

But I actually think it's correct to set props to check that the fetchError is handled correctly. This component doesn't process the error, it just calls the patronLookup function and lets other code process it and pass down the prop. So for this component's unit tests I'd want one test to make sure the patronLookup stub is called when the button is clicked, and another test to make sure that the appropriate thing is displayed when fetchError is set.

src/components/ResetAdobeId.tsx
adriana-alter: I agree that it would make more sense for the error to start as null, but it breaks the beforeEach hooks in the tests, like so: "TypeError: Cannot read property 'status' of null."  Is there a workaround...?

aslagle: You can change the type of `error`. Either do `error?: FetchErrorData` so it's optional and leave it out entirely when there's no error, or set it to `error: FetchErrorData | null` and set it to `null`.

src/components/ManagePatronsTabContainer.tsx
adriana-alter: Should this conditional go on the parent component (ManagePatrons) instead?  There's no scenario in which someone should have access to the ManagePatrons interface but not to the tabs...right?

aslagle: I could go either way. Right now you're correct, but it's possible we'll add other patron management tools in the future which may be accessible to different admin types.

src/components/ResetAdobeId.tsx
adriana-alter: I just tried both of those, but it's still breaking.

src/components/ManagePatronsTabContainer.tsx
adriana-alter: Ok, I'll leave it on the tabs!

src/components/ResetAdobeId.tsx
aslagle: Ah, is that a JS error instead of a TypeScript error? You may need to change somewhere you're using error.status to check if error exists first.

src/components/Header.tsx
```diff
@@ -55,6 +55,11 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
   }
 
   render(): JSX.Element {
+
+    let isLibraryManager = this.context.library && this.context.admin.isLibraryManager(this.context.library());
+    let isSystemAdmin = this.context.admin.isSystemAdmin();
+    let canAccessPatronManager = isLibraryManager || isSystemAdmin;
```
aslagle: All system admins are automatically library managers of all libraries, so you don't need to check system admin status here.

src/components/ResetAdobeId.tsx
adriana-alter: Yes, that was it!

src/stylesheets/header.scss
```diff
@@ -37,6 +37,11 @@
   }
 }
 
+.pull-right {
```
EdwinGuzman: @adriana-alter not sure about this. It'll still be off on narrower viewports but instead of it being two rows it'll turn to three. This also doesn't seem to cover all the pages. It's also going to be three rows without the main nav on the config page which looks off.
screen shot 2018-08-27 at 2 53 41 pm

src/stylesheets/header.scss
```diff
@@ -37,11 +41,22 @@
   }
 }
 
+.header-link a {
+  padding-left: 5px !important;
+  padding-right: 5px !important;
```
EdwinGuzman: The linter doesn't like `!important` and that is something we try not to use. You can update this by adding more specificity to the anchor elements.

src/stylesheets/header.scss
```diff
@@ -37,11 +41,22 @@
...
+}
+
 .pull-right {
+  max-width: 50%;
+  min-width: 30%;
```
EdwinGuzman: The linter also prefers the `display` rule to go above the rest.

EdwinGuzman:
@aslagle updated the `postForm` action function to resolve the same response and call `response.json()` instead of `.text()` and thus not have to call `JSON.parse`.

Also updating ResetAdobeId to use more of the redux flow to not store the error in state and use the `fetchError` from the store. Also using the `editedIdentifier` property which returns `"Success"` after calling the `resetAdobeId` action. Is this fine to use?

aslagle:
It's fine to use but maybe we should rename the property. I'm not sure you really need the message from the server in this case though?

EdwinGuzman:
The "Success" message? or the error message? The messaged displayed is different, so we don't use either that's coming from the server. Rename the property in the reducer?...

aslagle:
If you’re not displaying the success message you probably shouldn’t be using it at all. If you need to know if the request succeeded it’s better to use the status code.

EdwinGuzman:
@adriana-alter the last commit I made is pretty large so let me know if you want to go over it. Basically, along with the last commit in [circ-1031](https://github.com/NYPL-Simplified/circulation/pull/1031), the server will present the successful alert message to the UI instead of letting the UI handle it.

I also added an action to clear the store whenever the ManagePatronsForm component unmounts for both the `fetchError` and `patron` data. This is because whenever I had an unsuccessful or successful state and I moved to another library or another page and then went back to the Patron Manager page, the previous state was still there and that wasn't good.

src/actions.ts
```diff
@@ -623,4 +640,18 @@ export default class ActionCreator extends BaseActionCreator {
   runSelfTests(url: string) {
     return this.postForm(ActionCreator.RUN_SELF_TESTS, url, null).bind(this);
   }
+
+  patronLookup(data: FormData) {
+    const url = "/admin/manage_patrons";
```
aslagle: Since you changed the server to use `library_route`, these URLs need to be changed to start with the library short name. This will work if the server only has one library, but if you have multiple libraries it will use the default library even if you're in another library.

src/actions.ts
```diff
@@ -140,6 +141,10 @@ export default class ActionCreator extends BaseActionCreator {
   static readonly GET_SELF_TESTS = "GET_SELF_TESTS";
   static readonly RUN_SELF_TESTS = "RUN_SELF_TESTS";
 
+  static readonly PATRON_LOOKUP = "PATRON_LOOKUP";
+  static readonly RESET_PATRON_DATA = "RESET_PATRON_DATA";
```
aslagle: I'd like to call this 'clear' instead of 'reset' to be consistent with other actions that do something similar. There aren't any in here but there are some in opds-web-client.

src/actions.ts
```diff
@@ -168,7 +180,12 @@ export default class ActionCreator extends BaseActionCreator {
         }).then(response => {
           if (response.status === 200 || response.status === 201) {
             dispatch(this.success(type));
-            if (response.text) {
+            if (response.json && returnType === "JSON") {
+              response.json().then(data => {
+                dispatch(this.load<string>(type, data));
```
aslagle: Why is the type `string`? Shouldn't it be `any` since it's arbitrary JSON?

src/actions.ts
```diff
@@ -623,4 +640,18 @@ export default class ActionCreator extends BaseActionCreator {
...
+    return this.postForm(ActionCreator.PATRON_LOOKUP, url, data, "POST", "", "JSON").bind(this);
+  }
+
+  resetAdobeId(data: FormData) {
+    const url = "/admin/manage_patrons/reset_adobe_id";
+    return this.postForm(ActionCreator.RESET_ADOBE_ID, url, data).bind(this);
+  }
+
+  resetPatronData() {
+    return (dispatch) => dispatch(this.load<string>(ActionCreator.RESET_PATRON_DATA, null));
```
aslagle: Also why is this type `string`?

src/components/ManagePatronsForm.tsx
```diff
@@ -0,0 +1,100 @@
...
+    this.props.resetPatronData();
+  }
+
+  render(): JSX.Element {
+    const { patron, fetchError } = this.props;
+    const patronExists = !!(patron);
+
+    return (
+      <div className="manage-patrons-form">
+        { (fetchError && fetchError.status !== 200 && !patronExists) &&
```
aslagle: I think you can take out the status check now.

aslagle: What if the patron exists and there's a fetchError? Can that happen?

src/components/ManagePatronsForm.tsx
```diff
@@ -0,0 +1,100 @@
...
+          <ErrorMessage error={fetchError} />
+        }
+        { (!fetchError && patronExists) &&
+          <Alert bsStyle="success">Patron found: {patron.authorization_identifier}</Alert>
+        }
```
aslagle: I would put this below the form and I'd put the patron info there as well instead of in the reset adobe id section.

src/components/ResetAdobeId.tsx
```diff
@@ -0,0 +1,121 @@
...
+      patron,
+      fetchError,
+      responseBody,
+    } = this.props;
+    const patronExists = !!(patron && patron.authorization_identifier);
+
+    return (
+      <div className="patron-actions">
+        <h4>Reset Adobe ID</h4>
+        { (fetchError && fetchError.status > 200 && patronExists) &&
```
aslagle: you can also remove the status check here.

src/components/ManagePatronsForm.tsx
aslagle: I see, that's handled in the adobe section.

src/components/ResetAdobeId.tsx
```diff
@@ -0,0 +1,121 @@
...
+          <Alert bsStyle="danger">Error: failed to reset Adobe ID for patron {patron.authorization_identifier}</Alert>
+        }
+        { responseBody &&
+          <Alert bsStyle="success">{responseBody}</Alert>
+        }
+        <p>This feature allows you to delete the existing Adobe ID for an individual patron; a new Adobe ID will be assigned automatically when the patron logs in again. This step is necessary when patrons reach their device installation limit. Please be sure to inform patrons that resetting their Adobe ID will cause them to lose any existing loans or holds.</p>
+        { patron ?
+            <div>
+              <PatronInfo patron={patron} />
+              <p className="patron-warning"><b>Patron {patron && (patron.username || patron.personal_name)} will lose any existing loans or holds when the Adobe ID is reset.</b></p>
```
aslagle: I'd add authorization identifier too just in case the patron has no username and no personal name.

src/components/ManagePatronsForm.tsx
EdwinGuzman: The second suggestion sounds like a good change to make.

EdwinGuzman:
@adriana-alter one more thing, the version should be updated for this PR.

Let's also start a `CHANGELOG.md` file to keep track of changes in each version. Maybe something like this: https://github.com/NYPL/dgx-homepage/blob/master/CHANGELOG.md

EdwinGuzman:
Thanks. One more (last?) look @aslagle ?

src/components/PatronInfo.tsx
```diff
@@ -0,0 +1,32 @@
...
+          { patron.username &&
+            <li><label>Username:</label><p>{patron.username}</p></li>
+          }
+          { patron.personal_name &&
+            <li><label>Personal Name:</label><p>{patron.personal_name}</p></li>
+          }
+          { patron.email_address &&
+            <li><label>Email Address:</label><p>{patron.email_address}</p></li>
+          }
+          <li><label>Identifier:</label><p>{patron.authorization_identifier}</p></li>
```
aslagle: The terminology is a little inconsistent - in the search box it says enter the *barcode* and here it says *identifier*.

src/components/ResetAdobeId.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    return (
+      <div className="patron-actions">
+        <h4>Reset Adobe ID</h4>
+        { (fetchError && patronExists) &&
+          <Alert bsStyle="danger">Error: failed to reset Adobe ID for patron {patron.authorization_identifier}</Alert>
+        }
+        <p>This feature allows you to delete the existing Adobe ID for an individual patron; a new Adobe ID will be assigned
+          automatically when the patron logs in again. This step is necessary when patrons reach their device installation limit.
+          Please be sure to inform patrons that resetting their Adobe ID will cause them to lose any existing loans or holds.</p>
```
aslagle: I don't think this can cause you to lose holds, only loans that you have already downloaded in SimplyE.

src/index.tsx
```diff
@@ -67,6 +68,7 @@ class CirculationWeb {
             <Route path="/admin/web/dashboard(/:library)" component={DashboardPage} />
             <Route path="/admin/web/config(/:tab)(/:editOrCreate)(/:identifier)" component={ConfigPage} />
             <Route path="/admin/web/account" component={AccountPage} />
+            <Route path="/admin/web/patrons(/:library)(/:tab)" component={ManagePatrons} />
```
aslagle: Why is the library optional?

src/stylesheets/manage_patrons.scss
```diff
@@ -0,0 +1,88 @@
...
+          }
+        }
+      }
+    }
+
+    .tab-container {
+      min-height: 450px;
+      .nav-tabs {
+        min-height: 400px;
+        border-left: 1px solid #DDD;
```
aslagle: This should use one of the color variables.

src/components/Header.tsx
```diff
@@ -80,51 +82,56 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
...
                   <Link to={"/admin/web/lanes/" + this.context.library()}>Lanes</Link>
                 </li>
               }
-              <li>
+              <li className="header-link">
                 <Link to={"/admin/web/dashboard/" + this.context.library()}>Dashboard</Link>
               </li>
+              { isLibraryManager &&
+                <li className="header-link">
+                  <Link to={"/admin/web/patrons/" + this.context.library()}>Patron Manager</Link>
```
aslagle: The header is getting pretty cramped. Maybe we could call this "Patrons" in the header? Lists and Lanes are similar - the header says "Lists" and "Lanes" but on the actual pages it says "List/Lane Manager".

src/actions.ts
```diff
@@ -623,4 +639,18 @@ export default class ActionCreator extends BaseActionCreator {
...
+    return this.postForm(ActionCreator.PATRON_LOOKUP, url, data, "POST", "", "JSON").bind(this);
+  }
+
+  resetAdobeId(data: FormData, library: string) {
+    const url = "/" + library + "/admin/manage_patrons/reset_adobe_id";
+    return this.postForm(ActionCreator.RESET_ADOBE_ID, url, data).bind(this);
+  }
+
+  clearPatronData() {
+    return (dispatch) => dispatch(this.load<any>(ActionCreator.CLEAR_PATRON_DATA, null));
```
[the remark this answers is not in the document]
EdwinGuzman: Is that because we're not expecting any data back from this call?

src/components/Header.tsx
EdwinGuzman: That seems good for consistency purposes. I believe Kim wants to have a rearchitecture of the navigation menu. _If_ that ever happens, I'll keep you in the loop for that.

src/actions.ts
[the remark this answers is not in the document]
aslagle: Right

src/components/Header.tsx
aslagle: I was hoping to eventually remove some of the links (catalog, complaints, hidden), and change the main page to show an admin-specific OPDS feed that has those as lanes instead. I'd also like to combine lists and lanes somehow since I've gotten feedback that the distinction is unclear.

src/components/__tests__/ManagePatronsForm-test.tsx
```diff
@@ -0,0 +1,154 @@
...
+    it("displays an error message if there is an error", () => {
+      let errorMessage = wrapper.find(ErrorMessage);
+      const fetchError = {status: 400, response: "", url: ""};
+      expect(errorMessage.length).to.equal(0);
+
+      wrapper.setProps({ fetchError });
+
+      errorMessage = wrapper.find(ErrorMessage);
+      expect(errorMessage.length).to.equal(1);
+      expect(errorMessage.hasClass("alert-danger")).to.equal(true);
```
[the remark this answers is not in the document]
adriana-alter: In most cases, I am relying on the classes for styling...  (But I'll get rid of the tests for the ones where I'm not!)

PR #167, opened 2018-08-24
Library registration type

EdwinGuzman:
This helps fix [SIMPLY-1141](https://jira.nypl.org/browse/SIMPLY-1141).

It adds a select dropdown with "Testing" and "Production" stage values used to register a library. If a library chooses "Production", it should not appear again because there's no going back after going live.


src/components/__tests__/ServiceWithRegistrationsEditForm-test.tsx
```diff
@@ -137,5 +200,32 @@ describe("ServiceWithRegistrationsEditForm", () => {
...
+      let bpl = libraryRegistrationInfo.at(1);
+      let bplEditableInput = bpl.find(EditableInput);
+
+      expect(wrapper.state().registration_stage.bpl).to.equal(undefined);
+      expect(bplEditableInput.props().value).to.equal("testing");
+
+      bplEditableInput.props().onChange({ "short_name": "bpl", name: "Brooklyn Public Library" });
+
+      expect(wrapper.state().registration_stage.bpl).to.equal("testing");
+      expect(bplEditableInput.props().value).to.equal("testing");
```
EdwinGuzman: @aslagle this half works but I wanted to test selecting "production and then calling the `onChange` method but that didn't work. I noticed in another test the `onChange` prop itself was called and that helped more than `simulate.change` but things still didn't work fully.

src/components/ServiceEditForm.tsx
```diff
@@ -22,6 +22,7 @@ export interface ServiceEditFormState {
   libraries: LibraryWithSettingsData[];
   expandedLibraries: string[];
   selectedLibrary: string | null;
+  registration_stage?: { [key: string]: string } | null;
```
aslagle: Could this go in ServiceWithRegistrationsEditForm? It doesn't seem like it would be relevant to any other forms.

src/components/ServiceWithRegistrationsEditForm.tsx
```diff
@@ -1,69 +1,99 @@
...
-                        Not registered
-                      </span>
-                      <button
-                        type="button"
-                        className="btn btn-default"
-                        disabled={this.props.disabled}
-                        onClick={() => this.context.registerLibrary(library)}
-                        >Register</button>
+            { this.props.data.allLibraries.map(library => {
+                const setRegistryStage = this.getLibraryProp(library, "stage") || "testing";
```
aslagle: It could be misleading to say "Current stage: testing" when the library isn't registered at all yet.

src/components/__tests__/ServiceWithRegistrationsEditForm-test.tsx
aslagle: Here's a test that does something similar: https://github.com/NYPL-Simplified/circulation-web/blob/master/src/components/__tests__/ClassificationsForm-test.tsx#L149

src/components/ServiceEditForm.tsx
```diff
@@ -26,7 +26,7 @@ export interface ServiceEditFormState {
 
 /** Form for editing service configuration based on protocol information from the server.
     Used on most tabs on the system configuration page. */
-export default class ServiceEditForm<T extends ServicesData> extends React.Component<ServiceEditFormProps<T>, ServiceEditFormState> {
+export default class ServiceEditForm<T extends ServicesData, ServiceEditFormState> extends React.Component<ServiceEditFormProps<T>, any> {
```
aslagle: This looks pretty good but I'm not sure what the `ServiceEditFormState` parameter is doing here?

Why not have `U extends ServiceEditFormState` and then `U` instead of `any`?

src/components/ServiceWithRegistrationsEditForm.tsx
```diff
@@ -1,69 +1,107 @@
...
-                      </span>
-                      <button
-                        type="button"
-                        className="btn btn-default"
-                        disabled={this.props.disabled}
-                        onClick={() => this.context.registerLibrary(library)}
-                        >Register</button>
+            { this.props.data.allLibraries.map(library => {
+                const libraryRegistrationStatus = this.getLibraryProp(library, "status");
+                const setRegistryStage = this.getLibraryProp(library, "stage");
```
aslagle: I found this variable name a little confusing - does this mean the previously set registry stage on the server?

src/components/ServiceEditForm.tsx
EdwinGuzman: I had trouble getting the class that extended `ServiceEditForm` to be able to extend the state and this seemed to work. I think I tried `U extends ServiceEditFormState` and that didn't work, but I'll try it again.

src/components/ServiceWithRegistrationsEditForm.tsx
EdwinGuzman: Yes, I guess `currentRegistryStage` would be more helpful.

src/components/ServiceWithRegistrationsEditForm.tsx
```diff
@@ -1,69 +1,107 @@
...
+                return (
+                  <div className="service-with-registrations-library" key={library.short_name}>
+                    <div className="library-name">{ library.name }</div>
+                    <div className="library-registration-info">
+                      <div className="current-stage">
+                        { (setRegistryStage && libraryRegistrationStatus !== "failure") ?
+                          <span>Current Stage: {setRegistryStage}</span> :
+                          <span>No current stage</span>
+                        }
+                        { setRegistryStage !== "production" &&
```
aslagle: This is more of a question about the registration process rather than this pr, but if you're registering for the first time I'm not sure you should have the option to go straight to production. It probably doesn't make much difference though.

src/components/ServiceWithRegistrationsEditForm.tsx
```diff
@@ -1,69 +1,107 @@
...
+                const registration_stage =
+                  (this.state.registration_stage && this.state.registration_stage[library.short_name]) || "testing";
+
+                return (
+                  <div className="service-with-registrations-library" key={library.short_name}>
+                    <div className="library-name">{ library.name }</div>
+                    <div className="library-registration-info">
+                      <div className="current-stage">
+                        { (setRegistryStage && libraryRegistrationStatus !== "failure") ?
+                          <span>Current Stage: {setRegistryStage}</span> :
```
aslagle: I think this could use a bit more explanation, since you would expect if you select production and register then your current stage would then be production, but in fact it will be testing if the registry hasn't approved you yet.

src/components/ServiceWithRegistrationsEditForm.tsx
```diff
@@ -1,69 +1,107 @@
...
+                        <div>
+                          <span className="bg-success">
+                            Registered
+                          </span>
+                          <button
+                            type="button"
+                            className="btn btn-default"
+                            disabled={this.props.disabled}
+                            onClick={() => this.context.registerLibrary(library, registration_stage)}
+                            >Update registration</button>
```
aslagle: I know this came from the old code but these buttons are identical except the text. You could define a variable for the text and reduce the repetition here.

src/components/ServiceWithRegistrationsEditForm.tsx
leonardr: You don't go straight to production. This value becomes `Library.library_stage`, which is the library's opinion about what stage it should be in. A library doesn't go into production until both `library_stage` and `registry_stage` are set to "production", i.e. until the library and the registry agree it should be.

aslagle: That's true. I guess I think we should encourage libraries to do their own testing before going to production rather than relying on the registry admins to catch problems.

src/components/ServiceWithRegistrationsEditForm.tsx
EdwinGuzman: How long would you know until you've been approved?

src/components/ServiceWithRegistrationsEditForm.tsx
EdwinGuzman: Updated.

src/components/ServiceWithRegistrationsEditForm.tsx
EdwinGuzman: Updated.

src/components/ServiceWithRegistrationsEditForm.tsx
leonardr: I agree that more explanation would be useful here. We currently don't have a system (software or procedural) for approving libraries, or for communicating approval decisions to the libraries. We do it all via Slack and email.

src/components/ServiceEditForm.tsx
aslagle: I looked at this a little more and I don't think there's a clean way to do this - if you let the subclass change the state type you can't set the state in the parent class without getting type errors or using `any` or casting. 

This seems like a case where it would be better to use composition instead of inheritance. None of the stuff that's added by this component reuses anything from the parent component, so it would be better to just define a new component for that stuff so it can have its own state. This component can still exist and can be responsible for rendering the new component and its parent.

src/components/DiscoveryServices.tsx
```diff
@@ -35,10 +35,12 @@ export class DiscoveryServices extends GenericEditableConfigList<DiscoveryServic
 
   getChildContext() {
     return {
-      registerLibrary: (library: LibraryData) => {
+      registerLibrary: (library: LibraryData, registration_stage: string) => {
+        console.log(library, registration_stage);
```
aslagle: you left a console.log in here

PR #168, opened 2018-08-27
Fix logging url

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1155


EdwinGuzman:
There's no version bump for this. I think it's fine and we can bump the version for the bigger features when it gets approved and merged.

We should also start thinking about adding a changelog to keep track of all the changes and updates we do.

PR #169, opened 2018-08-28
Enable ErrorMessage to parse non-JSON problem detail strings

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1150

ErrorMessage can now parse problem detail strings so as to show all of the relevant information, rather than a) showing a less-helpful default error message or b) showing the entire problem detail document in a hard-to-read format.

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -185,8 +185,7 @@ export default class ActionCreator extends BaseActionCreator {
               };
               dispatch(this.failure(type, err));
               reject(err);
-            })
-            .catch(parseError => {
+            }).catch(parseError => {
               err = {
                 status: response.status,
                 response: defaultErrorMessage || "Failed to save changes",
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -185,7 +185,8 @@ export default class ActionCreator extends BaseActionCreator {
               };
               dispatch(this.failure(type, err));
               reject(err);
-            }).catch(parseError => {
+            })
+            .catch(parseError => {
               err = {
                 status: response.status,
                 response: defaultErrorMessage || "Failed to save changes",
--- a/src/components/ErrorMessage.tsx
+++ b/src/components/ErrorMessage.tsx
@@ -15,6 +15,7 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, voi
     let status = this.props.error.status;
     let errorMessageHeader;
     let errorMessageText;
+
     if (status === 401) {
       return (
         <Alert bsStyle="danger">
@@ -32,14 +33,18 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, voi
         response = JSON.parse(this.props.error.response).detail;
       } catch (e) {
         response = this.props.error.response;
-        if (this.isProblemDetail(response)) {
-          response = this.parseProblemDetail(response);
-          errorMessageHeader = response.title;
-          errorMessageText = response.description + response.status + ": " + response.detail;
+        // The response might be a problem detail document encoded as a string rather than as JSON;
+        // if so, we need to parse it and display the relevant information from it
+        // (rather than just displaying the entire string, which is hard to read)
+        let pdString = "Remote service returned a problem detail document";
+        if (this.isProblemDetail(response, pdString)) {
+          response = this.parseProblemDetail(response, pdString);
+          errorMessageHeader = response.title ? response.title : "Error";
+          errorMessageText = `${response.description}${response.status}: ${response.detail}`;
         }
       }
       if (!errorMessageText) {
-        errorMessageText = "Error: " + response;
+        errorMessageText = `Error: ${response}`;
       }
       return (
         <Alert bsStyle="danger">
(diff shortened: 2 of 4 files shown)


src/actions.ts
```diff
@@ -185,7 +185,8 @@ export default class ActionCreator extends BaseActionCreator {
               };
               dispatch(this.failure(type, err));
               reject(err);
-            }).catch(parseError => {
+            })
+            .catch(parseError => {
```
EdwinGuzman: The catch on the previous line is preferable to me for visual syntax style guide, but we don't have anything to say otherwise.

src/components/ErrorMessage.tsx
```diff
@@ -44,6 +57,27 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, voi
     }
   }
 
+  isProblemDetail(response) {
+    if (response.split(":")[0]) {
+      return response.split(":")[0] === "Remote service returned a problem detail document";
```
EdwinGuzman: Where are you getting this string from? Is the server always going to return this string for a problem detail?

If the server changes that string, this won't work, so maybe there's a better way to check this?

src/components/ErrorMessage.tsx
```diff
@@ -29,12 +31,23 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, voi
...
-        // response wasn't json
         response = this.props.error.response;
+        if (this.isProblemDetail(response)) {
+          response = this.parseProblemDetail(response);
+          errorMessageHeader = response.title;
+          errorMessageText = response.description + response.status + ": " + response.detail;
+        }
+      }
+      if (!errorMessageText) {
+        errorMessageText = "Error: " + response;
```
EdwinGuzman: Use string interpolation if possible `Error: ${response}`.

src/actions.ts
adriana-alter: Ok, changed it!

src/components/ErrorMessage.tsx
adriana-alter: Server is always going to return it.  (But if you can think of a better way to check this, I'd love to hear it; this was all I could think of...)

src/components/ErrorMessage.tsx
```diff
@@ -44,6 +57,27 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, voi
...
+  isProblemDetail(response) {
+    if (response.split(":")[0]) {
+      return response.split(":")[0] === "Remote service returned a problem detail document";
+    }
+    return false;
+  }
+
+  parseProblemDetail(response) {
+    let [status, detail, title] = this.extractProperty(response, ["\"status\": ", "\"detail\": ", "\"title\": "]);
+    let description = "Remote service returned a problem detail document with status ";
```
EdwinGuzman: From the test, it seems that the response could also contain this string? Maybe it can also be parsed from the response instead of hardcoding it here?

src/components/ErrorMessage.tsx
```diff
@@ -44,6 +57,27 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, voi
...
+    return {description, status, detail, title};
+  }
+
+  extractProperty(response, propertyStrings) {
```
EdwinGuzman: Comments on these functions would be helpful, especially because the splitting you're doing is confusing at first glance.

src/components/__tests__/ErrorMessage-test.tsx
```diff
@@ -46,6 +46,23 @@ describe("ErrorMessage", () => {
     expect(alert.text()).to.contain("response");
   });
 
+  it("parses non-JSON problem detail string", () => {
+    let pd = "Remote service returned a problem detail document: {\"status\": 502, \"detail\": problem text, \"title\": problem title}";
```
EdwinGuzman: If you remove `\"status\": 502` from this string, an `Cannot read property 'split' of undefined` error will be thrown. There's definitely no way to add that to a new property instead of embedding it in an existing string? If the server misses a property, the app will crash. Unless if you add more logic checks in `extractProperty`.

EdwinGuzman:
You can also bump up the version for the package for this PR. Manage patrons is `0.0.75`, and the library registration PR is `0.0.76`... maybe this can be `0.0.77` but it also depends when we merge PRs in.

src/components/__tests__/ErrorMessage-test.tsx
adriana-alter: There definitely isn't...but I don't think there's any way the server could miss a property.  I'll add more logic checks just to be safe!

src/components/ErrorMessage.tsx
EdwinGuzman: From my comment in the `isProblemDetail` function, if the server will always return that, you should reuse it here and add `with status` instead of creating it again.

src/components/__tests__/ErrorMessage-test.tsx
EdwinGuzman: It seems a bit weird to me to add a json document in a string from the server so the UI can parse it... but if there's no way around it.

PR #171, opened 2018-09-12
Custom list opds render

EdwinGuzman:
Related JIRA ticket: [SIMPLY-459](https://jira.nypl.org/browse/SIMPLY-459)

This PR is working off [Circ #1043](https://github.com/NYPL-Simplified/circulation/pull/1043) (still being worked on) and [server-core](https://github.com/NYPL-Simplified/server_core/pull/940).

The List page in the admin will now read an OPDS feed instead of a large JSON file and only fetch 50 entries at a time. Clicking on the "Load more" button will have the same behavior as the existing "Load more" button which is on the "Search Results" column.

A few new things:
- Added @adriana-alter 's refactor of the Load More button component.
- Added an action/reducer to concatenate the books from the OPDS feeds that are being read for each new custom list page.
- Updated `identifier_urn` to `id`.
- Keeping track of all the entries that have been removed in the `CustomListEntriesEditor`'s state and sending that to the server (easier to find the difference).

screen shot 2018-09-12 at 4 22 02 pm


aslagle:
You should hide the "load more" button when there are no more books to load (we assume there are no more if you see an empty page in the feed).

aslagle:
I don't think you need to show the page-level "loading" overlay when loading the next page - it's enough to change the "load more" button to indicate that.

aslagle:
When you switch to a different list, it should reset the scroll position - right now if I load more in a long list, and then switch, I end up scrolled down in the list I switched to.

aslagle:
When you add a book to a list, it doesn't update the total count above the list until you save (where it says "Displaying 1-100 of 118 books").

aslagle:
Before the cancel changes would only apply to unsaved changes. It looks like now it still shows after you save and will undo changes that you already saved?

src/actions.ts
```diff
@@ -581,7 +583,11 @@ export default class ActionCreator extends BaseActionCreator {
 
   fetchCustomListDetails(library: string, id: string) {
     const url = "/" + library + "/admin/custom_list/" + id;
-    return this.fetchJSON<CustomListDetailsData>(ActionCreator.CUSTOM_LIST_DETAILS, url).bind(this);
+    return this.fetchOPDS<CustomListDetailsData>(ActionCreator.CUSTOM_LIST_DETAILS, url).bind(this);
```
aslagle: I'm not sure why this is still using `CustomListDetailsData` - that seems like it would be obsolete now and we could delete that interface.

src/components/CustomListEditor.tsx
```diff
@@ -4,30 +4,42 @@ import {
...
+import { CollectionData, BookData } from "opds-web-client/lib/interfaces";
 import TextWithEditMode from "./TextWithEditMode";
 import EditableInput from "./EditableInput";
 import CustomListEntriesEditor from "./CustomListEntriesEditor";
 import XCloseIcon from "./icons/XCloseIcon";
 import SearchIcon from "./icons/SearchIcon";
 
+export interface List extends CollectionData {
+  collections?: AdminCollectionData[];
+}
```
aslagle: Is this for the collections that automatically add new books to the list? It seems weird since AdminCollectionData is not really related to lists. I'd probably make these separate props.

src/components/CustomListEditor.tsx
```diff
@@ -4,30 +4,42 @@ import {
...
   loadMoreSearchResults: (url: string) => Promise<CollectionData>;
+  loadMoreEntries: (url: string) => Promise<CollectionData>;
   isFetchingMoreSearchResults: boolean;
+  isFetchingMoreCustomListEntries: boolean;
   entryPoints?: string[];
+  entry_count?: string;
+}
+
+export interface Entry extends BookData {
+  medium?: string;
```
aslagle: You could import this from CustomListEntriesEditor.

src/components/CustomListEditor.tsx
```diff
@@ -4,30 +4,42 @@ import {
...
   responseBody?: string;
   searchResults?: CollectionData;
   editCustomList: (data: FormData, listId?: string) => Promise<void>;
   search: (url: string) => Promise<CollectionData>;
   loadMoreSearchResults: (url: string) => Promise<CollectionData>;
+  loadMoreEntries: (url: string) => Promise<CollectionData>;
   isFetchingMoreSearchResults: boolean;
+  isFetchingMoreCustomListEntries: boolean;
   entryPoints?: string[];
+  entry_count?: string;
```
aslagle: This prop has different casing than everything else - because it came that way from the server?

src/components/CustomListEntriesEditor.tsx
```diff
@@ -182,6 +206,12 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
                 </ul>
               )}
             </Droppable>
+            { nextPageUrl &&
```
aslagle: It seems fine to use nextPageUrl to determine whether to show the button but doesn't seem like it's working - maybe the server is putting in the url in even when there's no next page?

src/components/LoadButton.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from "react";
+import MoreDotsIcon from "./icons/MoreDotsIcon";
+
+export interface LoadButtonProps {
+  isFetching: boolean;
+  loadMore: () => any;
+}
+
+export default class LoadButton extends React.Component<LoadButtonProps, void> {
```
aslagle: I don't see a test for this component, though it is a pretty simple component..

src/components/CustomListEntriesEditor.tsx
EdwinGuzman: This is fixed with https://github.com/NYPL-Simplified/server_core/pull/950/commits/1b24ba91d77839fece0dee5e904d3cf61963e36a#diff-eb69c9549013acc45b217e06734d0ec9R756

src/actions.ts
EdwinGuzman: This can be `CollectionsData`, right?

src/components/CustomListEditor.tsx
EdwinGuzman: Yes, I'll update.

src/components/CustomListEditor.tsx
EdwinGuzman: Each collection in the list was part of the dictionary when it was returned. The same thing with `protocol` but it seems that's not being used in the front-end. Since now the server returns an OPDS feed, there is no `collections` property in the data. Is there another way to retrieve this?

EdwinGuzman: I guess the only reason this is here is because there are calls to `this.props.list.collection` but that always returns `undefined` since nothing is coming back from the server.

src/reducers/customListDetails.ts
```diff
@@ -4,7 +4,7 @@ import createFetchEditReducer from "./createFetchEditReducer";
 
 const loadRequest = (state, action) => {
   return Object.assign({}, state, {
-    isFetching: true,
+    isFetchingEntries: true,
```
EdwinGuzman: @aslagle re:
>I don't think you need to show the page-level "loading" overlay when loading the next page - it's enough to change the "load more" button to indicate that.

Is it safe to just add a property like this in the returned state even though it's not defined in the main reducer's interface?

src/components/CustomListEditor.tsx
aslagle: This will probably need to be refactored since we do still need that info. Maybe you could include it when you get all the lists from the server?

src/reducers/customListDetails.ts
aslagle: Probably not, I would expect a type error there.

EdwinGuzman: No errors come up.

src/components/CustomListEditor.tsx
EdwinGuzman: I have added collections to different libraries but none of the libraries return the collections they have. I can refactor but this also seems to be a bigger issue. All the collections here are empty: https://github.com/NYPL-Simplified/circulation/blob/master/api/admin/controller.py#L1501

src/components/LoadButton.tsx
EdwinGuzman: Added tests for this.

EdwinGuzman:
>Before the cancel changes would only apply to unsaved changes. It looks like now it still shows after you save and will undo changes that you already saved?

I can't replicate this.

EdwinGuzman:
I think I fixed the above for canceling changes. I wasn't correctly updating the deleted entries list in some instances. Fixed the scrolling to the top when switching lists, and updated the total count of entries in a list.

I'm not sure about the loading layer since I'm using the `isFetching` prop for the load more button animation/disabled status.

src/components/CustomListEditor.tsx
EdwinGuzman: It looks like, previously, the `custom_lists` endpoint returned an empty array for collections for all lists.

EdwinGuzman:
@aslagle Alright, fixed the issue with the collections for individual lists. All issues should be fixed now.

src/reducers/customListDetails.ts
aslagle: That's probably fine then.

src/components/CustomListEditor.tsx
```diff
@@ -52,20 +53,22 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
+    const listName = this.props.list && this.props.list.title ? this.props.list.title : "";
+    const nextPageUrl = this.props.list && this.props.list.nextPageUrl;
     const opdsFeedUrl = `${this.props.library}/lists/${listName}/crawlable`;
     return (
       <div className="custom-list-editor">
         <div className="custom-list-editor-header">
           <div>
             <TextWithEditMode
-              text={this.props.list && this.props.list.name}
+              text={this.props.list && this.props.list.title}
```
aslagle: There seems to be a bug related to this - when I edit an existing list I'm not seeing the name.

aslagle:
This may have been an existing bug: It's possible to save a list with an empty string as the title. Then if you edit a different list with a name, and go back to the empty string list, it shows the other list's title and you can't save.

aslagle:
I'm seeing the "cancel changes" link before I make any edits.

aslagle:
It also still seems to be possible to cancel changes after saving them.

src/components/CustomListEntriesEditor.tsx
```diff
@@ -46,16 +58,51 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
+    if (entries.length) {
+      let entriesCount;
+      // Simple case where the total is less than 50.
+      if (entryCount <= TOTALPERFEED) {
+        entriesCount = entries.length;
+      } else {
+        // If adding to a list makes the entries total over 50, then we just want to
+        // keep track of new additions.
+        if (entries.length > TOTALPERFEED) {
+          entriesCount = entryCount + (entries.length % TOTALPERFEED);
```
aslagle: This doesn't work if the last page of the feed has less than 50 books. For example: entryCount is 139, and I loaded all entries, so entries.length is also 139. It should say "Displaying 1-139 of 139 books", but it says "Displaying 1-139 of 178 books".

src/components/CustomListEntriesEditor.tsx
```diff
@@ -46,16 +58,51 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
+        } else {
+          // Otherwise, we want to subtract any deletions.
+          entriesCount = entryCount - (TOTALPERFEED - entries.length);
```
aslagle: What if there are both additions and deletions at the same time?

aslagle: I haven't thought this through, but I'd consider keeping track of additions separately like you're doing with deletions. It would be good if you could get it so you don't have to hard code the page size.

EdwinGuzman:
Both
>This may have been an existing bug: It's possible to save a list with an empty string as the title. Then if you edit a different list with a name, and go back to the empty string list, it shows the other list's title and you can't save.
and 
> I'm seeing the "cancel changes" link before I make any edits.

are existing bugs. I can try to fix them in this PR.

src/components/CustomListEntriesEditor.tsx
EdwinGuzman: Keeping track of additions now made it easier to keep track of count. Thanks for that suggestion.

src/components/CustomListEntriesEditor.tsx
EdwinGuzman: Keeping track of additions now so this should work now.

EdwinGuzman:
@aslagle and @adriana-alter can you give this another look? The "cancel changes" button shouldn't display right away and canceling changes after saving them shouldn't happen. You also can't save a list without a name.

src/components/CustomListEntriesEditor.tsx
```diff
@@ -62,39 +63,35 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
       nextPageUrl,
+      entryCount
     } = this.props;
     let entryListDisplay = "No books in this list";
+    let totalEntries = parseInt(entryCount, 10);
+
+    // This is for the case where there can be 55 total entries and the user
+    // clicks on the "Delete all" button. Now there are no visible entries
+    // but there are still 5 more entries in the entire list.
+    if (!entries.length && totalEntries && (totalEntries - deleted.length !== 0)) {
```
adriana-alter: I would imagine that users would expect/want to be able to clear the entire list, not just the visible entries, by clicking "Delete all."  (That would certainly be my expectation if I saw a "delete all" button...)  No?

src/components/CustomListEntriesEditor.tsx
```diff
@@ -62,39 +63,35 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
+      let notDisplaying = totalEntries - deleted.length;
+      entryListDisplay = `Currently not displaying any entries of ${notDisplaying} Books in this list`;
```
adriana-alter: This wording seems potentially confusing to me (especially the distinction between entries and books).  How about "Currently displaying 0 of ${notDisplaying} books"?

src/components/CustomListEditor.tsx
```diff
@@ -104,6 +104,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
             <button
               className="btn btn-default save-list"
               onClick={this.save}
+              disabled={!this.state.name && !this.hasChanges()}
```
adriana-alter: Should this be ||, not &&?  (If the list doesn't have a name, then there's no situation in which you should be able to save it; it doesn't matter whether it has changes.)

src/components/CustomListEditor.tsx
```diff
@@ -143,25 +144,38 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
         entries: (nextProps.list && nextProps.list.books) || [],
         collections: (nextProps.list && nextProps.listCollections) || []
       });
     }
     else if ((!this.props.list || !this.props.listCollections) && nextProps.list && nextProps.listCollections) {
+      let name = this.state.name;
+      let entries = this.state.entries;
+      let collections = this.state.collections;
+      if (nextProps.list.title !== name) {
+        name = nextProps.list.title;
```
adriana-alter: For the sake of readability: can we be more consistent in how we refer to a list's name (call it either "name" or "title", rather than using the two words interchangeably)?

src/components/CustomListEntriesEditor.tsx
```diff
@@ -237,12 +234,44 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
+              nextProps.entries.splice(i, 1);
+            }
+          });
+        });
+      }
+
+      // If there are any added entries and the user loads more entries,
+      // we want to added them back to the entire combined list.
+      if (this.state.added.length) {
+        nextProps.entries.concat(this.state.added);
```
adriana-alter: I'm pretty sure that .concat in JS creates a new array rather than modifying the existing one; I think you have to do `nextProps = nextProps.entries.concat(this.state.added)` in order to make nextProps.entries actually change...

src/components/CustomListEditor.tsx
EdwinGuzman: The `title` is from the OPDS spec I think. May be better to update name to be title instead.

src/components/CustomListEntriesEditor.tsx
EdwinGuzman: Yes, that was actually the previous behavior so I'll keep it.

src/components/CustomListEntriesEditor.tsx
EdwinGuzman: Yep.

src/components/CustomListEditor.tsx
```diff
@@ -231,15 +254,22 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
     if (this.props.list) {
-      data.append("id", this.props.list.id);
+      data.append("id", this.props.listId);
     }
-    let name = (this.refs["listName"] as TextWithEditMode).getText();
-    data.append("name", name);
+    let title = (this.refs["listTitle"] as TextWithEditMode).getText();
+    data.append("name", title);
     let entries = (this.refs["listEntries"] as CustomListEntriesEditor).getEntries();
     data.append("entries", JSON.stringify(entries));
```
aslagle: You don't need to change this, but I realized you could send the added entries instead of all the entries now that you're keeping track of additions. I think you wouldn't even need to change any server code.

src/components/CustomListEntriesEditor.tsx
aslagle: +1 to Adriana's suggestion and also I'm glad you found this edge case :)

src/components/CustomListEntriesEditor.tsx
aslagle: The comment should be updated if you're keeping that behavior, but this code is still needed because the user might delete all 50 of the entries one at a time.

src/components/CustomListEntriesEditor.tsx
```diff
@@ -182,15 +221,64 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
+              nextProps.entries.splice(i, 1);
+            }
+          });
+        });
+      }
+
+      // If there are any added entries and the user loads more entries,
+      // we want to added them back to the entire combined list.
+      if (this.state.added.length) {
+        nextProps.entries = nextProps.entries.concat(this.state.added);
```
aslagle: Does this move the added entries to the end instead of the beginning? Maybe it should be flipped?

src/components/CustomListEntriesEditor.tsx
EdwinGuzman: Yes, I'm rewriting the whole display text since I have to update the "delete all" functionality, too.

EdwinGuzman:
The code is not fully complete but I believe the counting, keeping track of changes, not canceling changes after saving, and list name issues have all been resolved. Can that be reviewed while I give another code update soon?

EdwinGuzman:
@aslagle It should all be done now. There is ONE thing that I just realized but I will need your help with. Let's say there are a total of 60 entries and you click on "delete all" but then decided to cancel those changes. Now the list "reverts" back to the 50 entries that were visible before but only those 50 and not the 10 that are still in the list but not shown. This is a simplified example of 101 or more since that could possibly mean more than one request to get the rest of the entries from the server. Any thoughts on this?

aslagle:
In that case yes, it should go back to the 50 entries and say it's showing 50 of 60, or if you had loaded multiple pages before deleting all it would show however many you had loaded. I think it should be possible to do that by using the props to restore the original state.

EdwinGuzman:
@adriana-alter @aslagle Alright, one more push. Instead of deleting all the entries in a list, the "delete" button will only delete all entries in the list that are _currently visible_.

screen shot 2018-10-16 at 2 53 50 pm

This is to "fix" the issue mentioned above.

aslagle:
Wait, how does this work on the server then? Are you still sending 'deleteAll'?

EdwinGuzman:
I'm not sending the "deleteAll" flag anymore. When all the entries that are currently visible get deleted, they all go to the `deleted` array and those get sent over.

EdwinGuzman:
Yes, I don't see why we would need it anymore so I can remove that from the server. I've tested various scenarios of deleting and adding but more thorough tests would be great.

src/components/CustomListEditor.tsx
```diff
@@ -101,6 +104,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
             <button
               className="btn btn-default save-list"
               onClick={this.save}
+              disabled={!this.state.title && !this.hasChanges()}
```
aslagle: I think this should be an or instead of and - you shouldn't be able to save in either of these conditions.

src/components/CustomListEditor.tsx
```diff
@@ -130,55 +134,65 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
-        name: nextProps.list && nextProps.list.name,
-        entries: (nextProps.list && nextProps.list.entries) || [],
-        collections: (nextProps.list && nextProps.list.collections) || []
+        title: nextProps.list && nextProps.list.title,
+        entries: (nextProps.list && nextProps.list.books) || [],
+        collections: (nextProps.list && nextProps.listCollections) || []
       });
-    }
-    else if ((!this.props.list || !this.props.list.collections) && nextProps.list && nextProps.list.collections) {
+    } else if (nextProps.list && nextProps.list.books.length !== this.state.entries) {
```
aslagle: Should this last comparison be `this.state.entries.length`? Looks like it's comparing a number with an array so that would never be equal.

src/components/CustomListEditor.tsx
```diff
@@ -130,55 +134,65 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
+      this.setState({
+        title: nextProps.list.title,
+        entries: nextProps.list.books,
+        collections: this.state.collections,
+      });
+    } else if ((!this.props.list || !this.props.listCollections) && nextProps.list && nextProps.listCollections) {
       this.setState({
-        name: this.state.name,
+        title: this.state.title,
```
aslagle: These conditions are a little confusing. Why isn't this one using `nextProps.list.title`?

src/components/CustomListEditor.tsx
```diff
@@ -231,15 +245,21 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
     let entries = (this.refs["listEntries"] as CustomListEntriesEditor).getEntries();
     data.append("entries", JSON.stringify(entries));
+    let deletedEntries = (this.refs["listEntries"] as CustomListEntriesEditor).getDeleted();
+    data.append("deletedEntries", JSON.stringify(deletedEntries));
     let collections = this.state.collections.map(collection => collection.id);
     data.append("collections", JSON.stringify(collections));
-    this.props.editCustomList(data, this.props.list && String(this.props.list.id)).then(() => {
+    (this.refs["listEntries"] as CustomListEntriesEditor).clearState();
+
+    this.setState({ title: this.state.title, entries });
```
aslagle: Should these two lines go in the `then` after the edit finishes?

src/components/CustomListEditor.tsx
aslagle: There's a bug with the cancel changes button that might be related to this. When you have a list with over 100 books and you delete all and then load more, the cancel changes button should be displayed but it's not.

src/components/CustomListEntriesEditor.tsx
```diff
@@ -46,16 +59,63 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
+        displayTotal = `0 - 0 of ${currentDisplay}`;
+        entryListDisplay = `Displaying ${displayTotal} Books`;
+      }
+    } else {
+      // No existing entries in a list so it's a new list or all entries
+      // were recently deleted.
+      if (entries && entries.length) {
+        currentDisplay = !added.length ? entries.length : added.length;
+        entriesCount = currentDisplay - deleted.length;
+        displayTotal = `1 - ${currentDisplay} of ${entriesCount}`;
```
aslagle: I found a couple bugs around this:

Save a new list. Add 10 books to the list, delete all, add one back. Result: "Displaying 1 - 11 of 1 books"
Save a new list with 3 books. Delete all, add the 3 books back, delete all again. Result: "Displaying 0 - 0 of -3 books"

src/components/CustomListEntriesEditor.tsx
```diff
@@ -232,25 +342,51 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
+  }
+
   reset() {
+    let entries = [];
+    if (!this.state.deleted.length) {
+      entries = this.props.entries || [];
+    } else if (this.state.added) {
+      entries = this.state.deleted.concat(this.removeAdded());
+    } else {
+      entries = this.state.deleted.concat(this.state.entries);
```
aslagle: I don't understand why you need this code. Won't you end up with the same thing as `this.props.entries` in all these cases?

src/components/CustomLists.tsx
```diff
@@ -273,6 +292,9 @@ export class CustomLists extends React.Component<CustomListsProps, CustomListsSt
   async editCustomList(data: FormData, listId?: string): Promise<void> {
     await this.props.editCustomList(data, listId);
     this.props.fetchCustomLists();
+    // if (listId) {
+    //   this.props.fetchCustomListDetails(listId);
+    // }
```
aslagle: Should this be removed?

src/components/CustomListEntriesEditor.tsx
EdwinGuzman: I thought so too but it doesn't seem like the original set of entries are in the `this.props.entries`. So this keeps track of anything that was removed and adds it back in, or added and removes it.

aslagle: That might be because of the other thing I pointed out in CustomListEditor.componentWillReceiveProps?

src/components/CustomLists.tsx
EdwinGuzman: Yes.

src/components/CustomListEntriesEditor.tsx
EdwinGuzman: Perhaps. I'll update that first then.

src/components/CustomListEditor.tsx
EdwinGuzman: I'm updating it so that it only checks `this.hasChanges`.

src/components/CustomListEditor.tsx
EdwinGuzman: That's how it was before I think. I just updated `name` to `title`.

src/components/CustomListEditor.tsx
aslagle: Was this to prevent saving a list with no title though? We still want that too.

src/components/CustomListEditor.tsx
EdwinGuzman: Yes, moving them.

src/components/CustomListEditor.tsx
EdwinGuzman: I think that's because when you delete 100 and load more, the component thinks it's at a new state in  the list. So if you then go and delete whatever is loaded and undo, you'll only get back what you _just_ deleted but not the previous 100 entries. Do you think this could be address in a separate issue so this PR doesn't keep being prolonged?

I'd also like to address some of the many `setStates` statements that go on when saving and clearing changes. There are about 3 different ones and since they're all asynchronous, it doesn't seem like all the changes propagate through very well.

src/components/CustomListEditor.tsx
EdwinGuzman: Yes, I covered that case.

src/components/CustomListEditor.tsx
```diff
@@ -130,55 +134,70 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
-        collections: nextProps.list.collections
+        collections: nextProps.listCollections,
       });
     }
   }
 
   hasChanges(): boolean {
-    const nameChanged = (this.props.list && this.props.list.name !== this.state.name);
+    const titleChanged = (this.props.list && this.props.list.title !== this.state.title) ||
+      !!(!this.props.list && this.state.title);
```
aslagle: Moving this empty title check here means it won't prevent you from saving if the title is empty but you've added entries or collections to the list. I think it was better to have this separate from `hasChanges` since it's more of a validation thing.

aslagle: Hmm, this does seem to work when creating a new list, but I don't understand how...

aslagle: I see, for a new list, `entriesChanged` and `collectionsChanged` are always false so it's only the title that determines whether you can save. I think this is fine but I might rearrange this method so it's easier to understand what's going on. Maybe put an `if (!this.props.list) { return !!this.state.title; }` at the top so the rest of the code is for when there's a list.

src/components/CustomListEntriesEditor.tsx
```diff
@@ -279,71 +418,117 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
       }
     }
-    this.setState({ draggingFrom: null, entries });
+    let added = this.state.added.filter(entry => entry.id !== id);
+    let deleted = this.state.deleted.filter(entry => entry.id !== id);
+    this.setState({
+      draggingFrom: null,
+      entries,
+      deleted,
+      added: added.concat([entry]),
```
aslagle: There is a subtle bug here when you delete a book and then add it back. In that case, the book should go back into entries and be removed from deleted, but should not but put in added since it was already in the entries originally.

EdwinGuzman: Yeah, it's more than subtle. I wasn't removing any entries that were deleted and added back from the deleted array and vice versa for new additions that were deleted. Same idea for adding all and removing all. I fixed this last night before I saw your comments. but I'll fix the other name changes you mentioned.

src/components/CustomListEditor.tsx
EdwinGuzman: Sounds good.

src/components/CustomListEditor.tsx
```diff
@@ -248,14 +275,16 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
-    (this.refs["listName"] as TextWithEditMode).reset();
+    (this.refs["listTitle"] as TextWithEditMode).reset();
     (this.refs["listEntries"] as CustomListEntriesEditor).reset();
-    this.setState({
-      name: this.state.name,
-      entries: this.state.entries,
-      collections: (this.props.list && this.props.list.collections) || [],
-      entryPointSelected: "all",
-    });
+    setTimeout(() => {
```
aslagle: What's this for?

EdwinGuzman: Reverting the list name back doesn't always get propagated to the state and the `setTimeout` gives it some time for it to update properly.

aslagle: Oh I see - calling reset on the other components triggers a `setState`, but `setState` doesn't execute right away. That's annoying but I don't have a better idea.

EdwinGuzman: Right, there are two `setState`s and since they're both async they don't always finish before this last setState. I tried other ways but this seems more straight forward than the others.

I mentioned this in a separate comment, there seems to be a few setStates going on at the same time which could be readdressed later.

EdwinGuzman:
Alright. I don't seem to get negative numbers for the count for any weird scenarios, for new or existing lists. I hope.

aslagle:
I'm still getting weird numbers in some cases :(
Not negative at least, though.

One example: I have a list with two books. Add one book, save list, delete all => Displaying 0 - 0 of 1 books.

I think it might be good to write more unit tests around all this logic?

EdwinGuzman:
Hmm alright. I'll write some more.

EdwinGuzman:
@aslagle check it out. I am now retrieving the list after a save so that the prop entries get updated. This makes it easier to keep track of the count. The only issue, which I don't know how big of an issue it is, is that if you have 200 entries and delete all the currently visible (50) and add a new one in, after saving, since it's retrieving the new saved list, it will load the newly added one but also 49 more since that's the new list from the server. I think this is fine and the display also says it's displaying 1-50 still. Let me know what you think of that.

src/components/CustomListEditor.tsx
```diff
@@ -130,55 +134,72 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
+        collections: nextProps.listCollections,
       });
     }
   }
 
   hasChanges(): boolean {
-    const nameChanged = (this.props.list && this.props.list.name !== this.state.name);
+    let titleChanged = (this.props.list && this.props.list.title !== this.state.title);
+    if (!this.props.list) {
+      titleChanged = !!this.state.title;
```
aslagle: A weird side effect of this is that the cancel changes button shows up for a new list when you put in a title. I think that's ok - cancel changes is basically a reset form button in that case.

PR #172, opened 2018-09-25
Success message on save

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1185


src/components/SaveButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+  form: any;
+}
+
+export interface SaveButtonDispatchProps {
+  handleSubmit?: (data: FormData) => void;
+}
+
+export interface SaveButtonProps extends React.Props<SaveButtonProps>, SaveButtonOwnProps, SaveButtonDispatchProps {}
+
+export default class SaveButton extends React.Component<SaveButtonProps, void> {
```
EdwinGuzman: Since this component doesn't connect to the Store, you can have all the prop values into just one interface, in this case it'll be the `SaveButtonProps`.

src/components/__tests__/IndividualAdminEditForm-test.tsx
```diff
@@ -1,16 +1,17 @@
 import { expect } from "chai";
-import { stub } from "sinon";
+import { stub, useFakeTimers } from "sinon";
```
EdwinGuzman: You imported `useFakeTimers` but you're not using it?

src/components/EditableConfigList.tsx
```diff
@@ -72,14 +76,16 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
     let AdditionalContent = this.AdditionalContent || null;
     return (
       <div className={AdditionalContent ? "has-additional-content" : ""}>
-        <h2>{this.itemTypeName.slice(0, 1).toUpperCase() + this.itemTypeName.slice(1)} configuration</h2>
+        <h2>{this.getItemType()} configuration</h2>
+        { this.props.responseBody && this.props.editOrCreate &&
```
EdwinGuzman: Why do you need the `this.props.editOrCreate` check here?

src/components/EditableConfigList.tsx
```diff
@@ -178,11 +195,49 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
     await this.props.editItem(data);
     this.props.fetchData();
   }
 
+  clearForm() {
+    let form = (this.refs["edit-form"] as any).refs.form;
+    if (form) {
+      form.reset();
+      for (const el of form.children){
+        if (el.className === "form-group" || (el.children[0] && el.children[0].className === "form-group")) {
```
EdwinGuzman: I don't have a solution or suggestion for this yet, but this is very coupled to the HTML markup and I don't think it should be.

src/components/SaveButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  render(): JSX.Element {
+
+   return (
+     <button
+       type="submit"
+       className="btn btn-default"
+       disabled={this.props.disabled}
+       onClick={this.handleSubmit}
+    >Submit</button>
```
EdwinGuzman: Maybe the component name should be `SubmitButton` or you could make this label be a prop in case you want to have a `SaveButton` that says `Save`.

src/components/EditableConfigList.tsx
adriana-alter: So that the success message will only show when the edit form or the create form is showing (i.e. will go away when you go back to the list of things).

src/components/SaveButton.tsx
adriana-alter: Yes, good idea!

src/components/__tests__/SaveButton-test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    wrapper.setProps({ handleData: handleData });
+    wrapper.simulate("click");
+    expect(handleData.callCount).to.equal(1);
+  });
+
+  it("should do nothing if disabled", () => {
+    wrapper.setProps({ disabled: true, handleData: stub() });
+    wrapper.simulate("click");
+    expect(save.callCount).to.equal(0);
+    expect(handleData.callCount).to.equal(0);
```
EdwinGuzman: You're using the wrong `handleData` stub here since you explicitly passed `stub()` as a prop.

src/components/__tests__/EditableConfigList-test.tsx
```diff
@@ -99,6 +103,25 @@ describe("EditableConfigList", () => {
     wrapper.setProps({ fetchError });
     error = wrapper.find(ErrorMessage);
     expect(error.length).to.equal(1);
+    expect(error.text()).to.equal("Error: test error");
+  });
+
+  it("shows success message on create", () => {
+    let success = wrapper.find(Alert);
+    expect(success.length).to.equal(0);
+    wrapper.setProps({ responseBody: "itemType", editOrCreate: "create"});
```
EdwinGuzman: Should this be `successMessage` instead of `responseBody`?

adriana-alter: No, the prop in EditableConfigList is still called `responseBody`.  (Its value is determined by the `successMessage` prop of the individual edit forms.)

src/components/EditableConfigList.tsx
adriana-alter: Agreed.  But the only solution I can see--and this would only be a partial solution anyway--would be to change the HTML such that this function can look for form elements by tagname rather than classname--i.e. have fieldset elements rather than div elements with class "form-group."  And that seems like a job for a different branch...no?

EdwinGuzman: Yes, it does seem like a task for different branch, or you could fix it now. Would take longer but it's relevant.

EdwinGuzman:
This looks good to me. Waiting on @aslagle .

src/components/EditableConfigList.tsx
adriana-alter: Do you think changing the divs to fieldsets is the right solution, or do you have a better idea?

EdwinGuzman: Which `div`s? Using `fieldsets` may be better for accessibility and breaking down long forms but they may not be 100% necessary for us. Hopefully Willa can give this a look over one day...

src/components/EditableConfigList.tsx
```diff
@@ -12,12 +13,14 @@ export interface EditableConfigListStateProps<T> {
   fetchError?: FetchErrorData;
   isFetching?: boolean;
   responseBody?: string;
+  successMessage?: string;
```
aslagle: This component has both responseBody and successMessage?

src/components/__tests__/IndividualAdminEditForm-test.tsx
```diff
@@ -525,28 +529,18 @@ describe("IndividualAdminEditForm", () => {
...
+      expect(save.callCount).to.equal(1);
 
-      wrapper.setProps({ responseBody: "newEmail" });
-      // Let the call stack clear so the callback after editItem will run.
-      const pause = (): Promise<void> => {
-          return new Promise<void>(resolve => setTimeout(resolve, 0));
-      };
-      await pause();
-      expect(window.location.href).to.contain("edit");
-      expect(window.location.href).to.contain("newEmail");
```
aslagle: I haven't completely figured out what this PR is doing yet, but I think it's important to test that the redirect when you save a new item works properly. Maybe that should be in a test of `save` instead of here now?

aslagle:
I'm going to need some explanation of what this PR is. Did you remove the redirect to the edit page for a new item when it's created? I like adding a success message but I think the redirect is still useful.

aslagle:
I would suggest showing the "You created a new X" success message on the edit page after you redirect to it rather than clearing the create form.

adriana-alter:
@aslagle Here's the JIRA ticket explaining the purpose of the PR: https://jira.nypl.org/browse/SIMPLY-1185

I took out the redirect because:
1)  it wasn't working anyway (as I recall, the `fetchData()` call was blocking it...?)
2)  I didn't see a way to have both a redirect and a success message (since the redirect takes the user away from the page that's displaying the success message) and, as between the two, the success message seemed more important.  I like your suggestion about displaying the success message on the edit page, though.  
3) Jack, the UX designer, thought that it would be a good idea to clear the create form and stay on the create page, with the rationale that users are more likely to want to create a bunch of new things in a row than they are to want to immediately edit the new thing that they've just finished making.  (I honestly don't know how realistic that is, though; opinions welcome!)

EdwinGuzman:
Also, the redirect https://github.com/NYPL-Simplified/circulation-web/blob/master/src/components/LibraryEditForm.tsx#L69 to the edit page relied on the `this.props.responseBody` value but that kept getting overridden by the `fetchData()` call to `null`. That's why I think the redirect didn't previously worked.

aslagle:
Thanks @adriana-alter that's helpful. I think when you're first setting up a circulation manager it would be more likely that you'd be creating a bunch of things in a row, but on later visits you'd probably be creating one thing or a few things. Hopefully you wouldn't need to edit the thing after (unless it's something like the registry where you have to create it and then register libraries from the edit page), but I don't know if it's helpful to see the newly created thing or not. I didn't know about the `fetchData` problem either.

christhepianist:
Hopping in to hopefully help on this:
"Hopefully you wouldn't need to edit the thing after (unless it's something like the registry where you have to create it and then register libraries from the edit page), but I don't know if it's helpful to see the newly created thing or not. "

Most of the frustration will be at the implementation stage, so after you create the thing, you'll want a blank form to create another. During on-going use of the service, you will be creating just one or two, so it won't matter as much where you leave them; a blank form is as good as the newly edited data. 

It would be nice, however, to have a link that is visible that allows the admin to see the thing that was just created if they want to do that.

aslagle:
Clearing the form isn't working after creating a library.

src/components/EditableConfigList.tsx
```diff
@@ -12,12 +13,14 @@ export interface EditableConfigListStateProps<T> {
...
 }
 
 export interface EditableConfigListDispatchProps<T> {
   fetchData?: () => Promise<T>;
   editItem?: (data: FormData) => Promise<void>;
   deleteItem?: (identifier: string | number) => Promise<void>;
+  save?: (data: any) => void;
```
aslagle: I'm not sure what this prop is for. It looks like this component is still using `editItem`.

src/components/EditableConfigList.tsx
```diff
@@ -178,11 +195,49 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
   }
 
+  save(data) {
+    this.editItem(data).then(() => {
+      // Scrolling to the top lets the user see the success message
+      window.scrollTo(0, 0);
+    });
+    setTimeout(() => {
+      this.refs["edit-form"] && this.clearForm();
+    }, 500);
```
aslagle: It's pretty important to *not* clear the form if there was an error - likely you just need to fix one field and you wouldn't want to have to re-enter the rest of the info. It looks like this would clear the form regardless of whether the save was successful.

src/components/EditableConfigList.tsx
```diff
@@ -164,6 +171,16 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
     return item[this.labelKey];
   }
 
+  getItemType() {
+    return this.itemTypeName.slice(0, 1).toUpperCase() + this.itemTypeName.slice(1);
+  }
+
+  successMessage() {
+    let verb = this.props.editOrCreate === "edit" ? "edited this" : "created a new";
+    let itemType = this.getItemType().toLowerCase();
```
aslagle: This won't work for "CDN".

src/components/EditableConfigList.tsx
```diff
@@ -178,11 +195,49 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
+    }, 500);
+  }
+
   async editItem(data: FormData): Promise<void> {
     await this.props.editItem(data);
     this.props.fetchData();
   }
 
+  clearForm() {
+    let form = (this.refs["edit-form"] as any).refs.form;
```
aslagle: I need to think about whether this can be improved, but first off I think most of this code should be moved to `ServiceEditForm`. I don't think a component should need to know about what's inside another component it renders.

src/components/EditableConfigList.tsx
aslagle: I would probably move the resetting down into the EditForm/ProtocolFormField components and rely more heavily on refs to find them. I'm not sure clearing inputs is enough, because in some cases (e.g. languages in the library edit form), the user may add inputs that need to be removed on reset, not just cleared.

src/components/LibraryEditForm.tsx
```diff
@@ -1,13 +1,14 @@
...
 import ProtocolFormField from "./ProtocolFormField";
+import SaveButton from "./SaveButton";
 import { LibrariesData, LibraryData } from "../interfaces";
 
 export interface LibraryEditFormProps {
   data: LibrariesData;
   item?: LibraryData;
   disabled: boolean;
-  editItem: (data: FormData) => Promise<void>;
+  save: (data: any) => void;
```
aslagle: Shouldn't the type of this be the same as `editItem`'s was before?

src/components/SaveButton.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from "react";
+
+export interface SaveButtonProps {
+  disabled: boolean;
+  save: (data: any) => void;
+  handleData?: (data) => void;
+  form: any;
+  text?: string;
+  handleSubmit?: (data: FormData) => void;
```
aslagle: Doesn't look like this last prop is used.

src/components/SaveButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+       disabled={this.props.disabled}
+       onClick={this.handleSubmit}
+    >{text}</button>
+   );
+ }
+
+ handleSubmit(event) {
+   event.preventDefault();
+   let data = new (window as any).FormData(this.props.form as any);
+   data = (this.props.handleData && this.props.handleData(data)) || data;
```
aslagle: Looking at one of the `handleData`s, you don't need to set `data = ...` here, you can just call `handleData` and it will modify its argument. It's working because of the `|| data` at the end.

src/components/__tests__/EditableConfigList-test.tsx
aslagle: This confused me too. Why not rename `responseBody` as well?

src/components/EditableConfigList.tsx
adriana-alter: Yes; the difference is that `responseBody` gets overwritten by the `fetchData()` call, but `successMessage` sticks around so that the success alert component can access it even after the data on the page has been refreshed.

aslagle: Could we fix `responseBody` rather than adding another thing?

src/components/SaveButton.tsx
adriana-alter: Could you explain this to me in more detail?  My thought process: if I just call `handleData`, then yes, it will modify its argument, but the version of `data` which I pass to `this.props.save` on line 34 will still be whatever it was originally set to on line 33, not the modified version...

src/components/LibraryEditForm.tsx
adriana-alter: I just tried it, and it results in a whole bunch of error messages.  But anyway, `save` isn't a renamed copy of `editItem`; it's a separate function which, in EditableConfigList, calls `editItem`...

src/components/EditableConfigList.tsx
adriana-alter: I don't think `responseBody` is broken; my understanding is that, everywhere else in the code, it's working exactly the way it's supposed to, so it made more sense to add another thing for this one specific use case than to make a global change that would affect (and probably break) a lot of other things which are currently fine.  That said, I agree that it would be much more appealing and less clunky if we didn't have to add another thing.  @EdwinGuzman, thoughts on this?

src/components/EditableConfigList.tsx
adriana-alter: `save` is calling `editItem`, not replacing it.  It seemed cleaner to me to have just one `save` function, defined in the parent component to all of the forms, which could then be passed down to the save button via each form (and which takes care of all of the save behavior--edits the item AND clears the form), rather than having to define a different save function for every single form by copying and pasting near-identical code a bunch of times.

src/components/SaveButton.tsx
aslagle: This concept is called "pass by reference" vs "pass by value". Different languages and data types work differently, but in JS, objects are passed by reference, so when `handleData` modifies its `data` argument that actually changes the original `data` variable from line 33 as well as the `data` variable inside the function, because these variables are referencing the same object in memory. Simple data types like numbers or strings are passed by value, so that doesn't happen.

src/components/EditableConfigList.tsx
```diff
@@ -177,8 +177,9 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
 
   successMessage() {
     let verb = this.props.editOrCreate === "edit" ? "edited this" : "created a new";
-    let itemType = this.getItemType().toLowerCase();
-    return `Successfully ${verb} ${itemType}`;
+    let itemType = this.getItemType();
+    let formattedItemType = itemType === "CDN" ? itemType : itemType.toLowerCase();
```
aslagle: Maybe we could check if the type is all upper case instead of checking for the specific value? I'd probably forget to update this if I added a new type that also needed it.

adriana-alter: Even better; will do :)

src/components/SaveButton.tsx
aslagle: And by the way, it would also be fine to have `handleData` return the modified data - that way the caller doesn't have to know if `handleData` will modify its argument or return something new based on it. That's not what you were doing though - `handleData` was returning nothing but your code worked because the `|| data` was setting it back to the same original object.

src/components/EditableConfigList.tsx
aslagle: That sounds good, but I don't think it should be a prop then. It's an instance method on the component, not something passed into it.

src/components/LibraryEditForm.tsx
aslagle: You're right. Looks like your `save` doesn't return anything so `void` is correct, but data should still be type `FormData` - you may need to add that type in `save`'s definition too.

adriana-alter: I changed it in EditableConfigList and in SaveButton, but when I try to change it in the edit forms, this happens... 
screen shot 2018-10-02 at 11 33 01 am

src/components/EditableConfigList.tsx
adriana-alter: Agreed.  I made it a prop for testing purposes; having it be a prop is the only way I know how to be able to pass a stubbed version of it to the wrapper in the test.  I'm guessing that's completely wrong anyway...?

src/components/LibraryEditForm.tsx
aslagle: Hmm, I think that's because you have it as the `onSubmit` prop for the form. Typescript is correctly telling you that won't work, because there won't be any form data passed to it when it's called from a form submit.

src/components/EditableConfigList.tsx
aslagle: Yeah, that doesn't sound like it would work since your code wouldn't use the prop. You can use sinon's stub to replace an existing function, that's probably what you're looking for.

src/components/LibraryEditForm.tsx
adriana-alter: That seems plausible.  In that case, do you think it's ok to leave it as-is in the forms?  (Or is there a workaround for the Typescript error?)

src/components/EditableConfigList.tsx
adriana-alter: To my knowledge, that's what I'm currently doing...no?  (It seems to me that there must be a different/better way to do this, but I haven't found one.)  If the functions that I'm replacing aren't props, this stops working.

`beforeEach(() => {
    fetchData = stub();
    editItem = stub().returns(new Promise<void>(resolve => resolve()));
    save = stub();
    deleteItem = stub().returns(new Promise<void>(resolve => resolve()));
    canCreate = true;
    canDelete = true;

    wrapper = mount(
      <ThingEditableConfigList
        data={thingsData}
        fetchData={fetchData}
        editItem={editItem}
        save={save}
        deleteItem={deleteItem}
        csrfToken="token"
        isFetching={false}
        />
    );
  });`

src/components/LibraryEditForm.tsx
aslagle: If you want to make `onSubmit` work (so you can submit the form by pressing enter instead of clicking save, for example), you would have to use a different function that gets the form data before calling `save`, like what you do in `SaveButton`. Otherwise, you might as well remove the `onSubmit` prop since it won't work, and that would hopefully fix the typescript error.

src/components/EditableConfigList.tsx
aslagle: I don't see how the `save` prop could be doing anything since you never use `this.props.save`. What happens if you remove it?

src/components/LibraryEditForm.tsx
adriana-alter: Hmm.  I definitely would like to make `onSubmit` work (because accessibility).  On the other hand, if I'm understanding correctly, that would entail going back to having a separate `save` function defined in each form, which is one of the things that I was trying to change with this branch.  What do you think?  Prioritize accessibility?  Am I the only one who's bothered by having the separate `save` functions anyway? :)

src/components/EditableConfigList.tsx
adriana-alter: This:
screen shot 2018-10-02 at 12 01 43 pm

src/components/LibraryEditForm.tsx
aslagle: No, I think it's great to be able to reuse that code. Maybe you could move `SaveButton`'s `handleSubmit` to be defined outside the component, so that you could import it in the forms and use it for `onSubmit` as well. You'd have to change what it does with `this.props.form` though.

src/components/EditableConfigList.tsx
aslagle: You would also need to remove the stub you're passing in to the component in the tests.

src/components/LibraryEditForm.tsx
adriana-alter: Could you explain in more detail?  (That sounds like a really good idea--I'm just unclear on how to implement it, since I don't think I've seen a pattern like that before...)

src/components/EditableConfigList.tsx
adriana-alter: Oh.  Yes, of course.  Thank you!!!  Removing it now.

src/components/LibraryEditForm.tsx
aslagle: You can export a function from a module like you would with an interface or a component (some of the reducers do this). Then import it like `import { handleSubmit }, SaveButton from "./SaveButton"`. I'm not sure exactly how you'd get the form, you'd probably need to bind handleSubmit to this when you use it in SaveButton, and then check if this.props exists, and if not, get the form from the event instead. The typings for that could be messy.

You could also use currying to pass in the form as an argument, but that's harder to explain.

adriana-alter: It turned out to be weirdly easy; apparently (magically...??) binding handleSubmit to this in SaveButton makes this.props.form available even when handleSubmit is being triggered by the form instead of by the button.  (I am extremely surprised and slightly confused that this works...but it definitely is working!)

aslagle:
There are some cases where you're only supposed to have one thing, for example Storage Services. Once you create one you can't create another. That might be the only case actually. So for that we would not want to clear the create form.

src/components/LibraryEditForm.tsx
aslagle: Wow - I am confused that this works too but I tested it and it does seem to work :)

adriana-alter:
@aslagle In the case(s) where you're only supposed to have one thing, what do you think should happen instead of clearing the form?  Redirect to the edit form?  Or redirect to the list of things?  Or stay on the create page but just don't clear the form?  Or...?

aslagle:
Redirecting to the edit form or the list of things would be fine. I don't think you should stay on the create page because it would let you try to save another new thing.

src/components/EditableConfigList.tsx
aslagle: It looks like you've edited `mapStateToProps` for all the subclasses of this to change `responseBody` to be `successMessage` from the store, but you haven't added the `successMessage` prop. So it seems like `this.props.successMessage` will never exist, except maybe in tests, and I don't see anything that's using it in this class. Do we actually need both props?

src/components/EditableConfigList.tsx
```diff
@@ -33,7 +35,7 @@ export interface EditFormProps<T, U> {
   item?: U;
   data: T;
   disabled: boolean;
-  editItem: (data: FormData) => Promise<void>;
+  save?: (data: any) => void;
```
aslagle: This should probably be `FormData` instead of `any` and all the edit forms should also have their `save` prop types updated.

src/components/EditableConfigList.tsx
```diff
@@ -178,11 +221,67 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
+  save(data: FormData) {
+    this.editItem(data).then(() => {
+      // Scrolling to the top lets the user see the success message
+      window.scrollTo(0, 0);
+      if (this.limitOne && this.props.editOrCreate === "create") {
+        // Wait for two seconds so that the user can see the success message,
+        // then go back to the previous page
+        setTimeout(() => {
+          window.location.href = this.urlBase;
+        }, 2000);
```
aslagle: I would redirect immediately and show the success message on the previous page, but if that's difficult it's not a huge deal.

src/components/EditableConfigList.tsx
```diff
@@ -178,11 +221,67 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
+  }
+
   async editItem(data: FormData): Promise<void> {
     await this.props.editItem(data);
     this.props.fetchData();
   }
 
+ clearForm(form) {
+    if (form) {
+      let inputs = form.elements;
```
aslagle: I still think it would be good to move most of this code to lower-level components, but if you'd rather keep it here it should have some thorough tests. I'd want to make sure a test fails when someone changes something in another component that breaks this code, since they might not realize this needs to be updated.

src/components/EditableConfigList.tsx
```diff
@@ -178,11 +221,67 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
+        let input = inputs[i];
+        input.disabled = false;
+        switch (input.type) {
+          case "text":
+            input.value = "";
+            break;
+          case "checkbox":
+            if (!input.dataset || input.dataset.is_default !== "true") {
+              input.checked = false;
+            }
```
aslagle: Do you also need to make sure it's checked if it is the default?

src/components/EditableConfigList.tsx
```diff
@@ -178,11 +221,67 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
+            input.selectedIndex = 0;
+            break;
+          case "file":
+            input.value = "";
+            break;
+        }
+      }
+      let removableItems = document.getElementsByClassName("with-remove-button");
+      for (let i = 0; i < removableItems.length; i++) {
+        removableItems[i].parentNode.removeChild(removableItems[i]);
```
aslagle: It's really surprising to me that this works, since the HTML is normally controlled by the state of `ProtocolFormField`. Is it possible the items have been removed from the HTML but are still in the state, and will be included in saved data even though they're not displayed?

src/components/EditableConfigList.tsx
adriana-alter: > This would involve changing a LOT of other things; certainly doable if necessary, but if you don't think it's a big deal, then I'd rather leave it alone...

src/components/EditableConfigList.tsx
adriana-alter: I'd rather keep it here; it makes more sense to me to have it all in one place, at the top level.  Open to suggestions, though; why do you think it should be moved?

I completely agree that it should be tested, but I spent a ridiculous number of hours trying to write tests for it, and couldn't get them to work--the tests apparently don't like manually changing the value of HTML elements.  (They don't throw error messages or anything; it's just that the elements retain their original value.)  I couldn't find anything helpful online or elsewhere in the codebase.  Do you have any suggestions?

src/components/EditableConfigList.tsx
adriana-alter: YES.  Good catch.  Thank you.

adriana-alter: Actually, it's already doing that; cool!

src/components/EditableConfigList.tsx
aslagle: If it was moved to component like `ProtocolFormField`, you could manipulate the state instead of changing the HTML values. That seems like a cleaner way to use React - maybe it does work to change the HTML, but I don't understand why it works. That would also make it easier to test, although it would probably be more code overall.

The other reason I'd like to move it is to increase the encapsulation of the other components. I see the HTML of other components as an internal detail that other parts of the code shouldn't know about. If any code can manipulate another component's HTML, then someone working on that component has to worry about breaking things somewhere else in the code, and it's hard to make changes. If you define a public API for the component and make sure the rest of the code uses that public API, that developer can change any internal details as long as they make sure the public API still works.

src/components/EditableConfigList.tsx
aslagle: Yeah, that is what I meant.

src/components/EditableConfigList.tsx
aslagle: As an example, I have a PR where I added a color picker to ProtocolFormField. If I was doing that after this branch had been merged, I would need code to reset the new field type, but since it's in a different file, I probably wouldn't realize I needed it. If `ProtocolFormField` itself had a reset method it's more likely I would notice I needed to update it.

adriana-alter: I changed it.  Writing tests now.

src/components/EditableConfigList.tsx
EdwinGuzman: @aslagle I see what you're saying and that seems right. The `successMessage` prop isn't being used and is not needed. The `responseBody` prop is being used but the `successMessage` attribute in the state is what is being mapped to it.

The above explanation regarding `responseBody` and `successMessage` is correct, but for it's not what it is used for in this component.

src/components/EditableConfigList.tsx
```diff
@@ -158,12 +166,47 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
   label(item): string {
     return item[this.labelKey];
   }
 
+  getItemType() {
+    return this.itemTypeName.slice(0, 1).toUpperCase() + this.itemTypeName.slice(1);
+  }
+
+  formatItemType() {
+    let itemType = this.getItemType();
```
EdwinGuzman: I'd prefer to see `const` instead of `let` for these since they don't change.

src/components/EditableConfigList.tsx
EdwinGuzman: User experience wise, that's a good idea, even if it's difficult to do.

src/components/EditableConfigList.tsx
```diff
@@ -178,11 +221,42 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
     await this.props.editItem(data);
     this.props.fetchData();
   }
 
+ clearForm(inputs) {
+   if (inputs) {
+     let keys = Object.keys(inputs);
+     for (let i = 0; i < keys.length; i++) {
+       let key = keys[i];
+       inputs[key].clear && inputs[key].clear();
```
EdwinGuzman: What is this line doing?

src/sharedFunctions.ts
```diff
@@ -0,0 +1,20 @@
+export function handleSubmit(event) {
```
EdwinGuzman: I don't think you have tests for these functions. You're also missing comments for this function.

EdwinGuzman:
I find it a bit jarring that the page redirects after a certain amount of time. I haven't tested this in a while and now I wasn't expecting that behavior. I was about to edit a form from something I created and the refresh was confusing.

EdwinGuzman:
Not necessary but we should think about having a success message on deleting a setting to be consistent.

EdwinGuzman:
Are you also working on the error messages in a form? If so, scrolling to the top of the form would be good for both success and error messages, especially in the long new library form.

EdwinGuzman:
I got this error when creating a `sysLog` logging service:
screen shot 2018-10-10 at 12 04 24 pm
seems to come from `ServiceEditForm`.

src/components/EditableConfigList.tsx
adriana-alter: Some of the types of input that the form might contain--specifically, ProtocolFormField and EditableInput--have their own `clear` method (it was already there).  This is checking whether this is one of the kinds of input that has a `clear` method, and, if so, calling `clear()` on it.

adriana-alter:
@EdwinGuzman Re: the page redirecting, that (hopefully!!) only happens for the things where the user is only allowed to create one; the point was just to avoid displaying the blank create form (like everything else now does on submit), because that will lead users to believe that they can go ahead and create another one.  I'll make it redirect to the edit form instead of the list, ok?

Re: success messages on delete, I don't really see the point; you can only delete things if you're on the list page, so you can tell that the delete has been successful because you can see that the thing you've deleted is no longer on the list.  No...?

src/components/EditableConfigList.tsx
adriana-alter: @aslagle @EdwinGuzman I've changed this so that it redirects to the edit page rather than the list page.  Do you still think the success message should be moved (i.e. displayed after the redirect)?

EdwinGuzman: The way it is now looks fine. You're redirected to a blank form, but you see the success message and have the option to go to one just created through the link.

EdwinGuzman:
Looks good. I think the logging system configuration doesn't redirect. It currently allows you to create more than one `sysLog` configuration. But when you try to do it you get an error that you can only create one. Seems like an existing bug.

I occasionally see messages on apps about successfully deleting things but it's not necessary for this.

I think we may be good to go for this PR.

src/components/EditableConfigList.tsx
```diff
@@ -178,11 +221,42 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
+  }
+
   async editItem(data: FormData): Promise<void> {
+    // Scrolling to the top lets the user see the success or error message
+    window.scrollTo(0, 0);
     await this.props.editItem(data);
     this.props.fetchData();
   }
 
+ clearForm(inputs) {
```
aslagle: This is looking a lot better but I would move it to ServiceEditForm and have it look up its own inputs instead of passing them in.

adriana-alter: Not everything inherits from the ServiceEditForm (unless I'm reading something very very wrong...?); the IndividualAdminEditForm, LibraryEditForm and SitewideSettingEditForm are all separate, so I thought it made more sense to have there just be one clearForm function in their shared parent component.  Do you think moving clearForm to sharedFunctions.ts (like I did with handleSubmit) would be a viable solution?

adriana-alter: Update: looking at it, it seems to me that the only way to do this would be to also move the `save` function down out of EditableConfigList and into the form (so as to preserve the functionality where it checks to make sure the save was successful before clearing the form).  Which should maybe happen anyway.  What do you think?  (Obviously, I would move it to sharedFunctions and call it from the form, not just copy/paste it a bunch of times.)

src/components/EditableConfigList.tsx
aslagle: I still think this should be addressed.

src/components/EditableConfigList.tsx
aslagle: You're right that those forms don't have a common parent class.

Is it true that adding this form clearing code adds a requirement that all the forms have refs for all their inputs? If so, I would like that to be clear from looking at the form classes themselves. I think moving `clearForm` to `sharedFunctions` is a good idea, and each of the other forms could have a function that gets their `refs` and then calls `clearForm`? I like having the `ref` access inside the component that defines the `refs`.

adriana-alter: That is true, but in almost all cases, the forms already had refs for all their inputs, which were being used for other stuff; I didn't add that many.  I'll move `clearForm` to `sharedFunctions`!  Does it seem ok to you to also move `save` to `sharedFunctions` (and give the forms a function that gets the `refs` and then calls `save`)?  I don't see how else to preserve the current behavior of predicating whether to call `clearForm` on the result of calling `save` (i.e. not clearing the form if there's an error message)...

src/components/EditableConfigList.tsx
adriana-alter: Sorry, thought I had; coming right up!

src/components/EditableConfigList.tsx
aslagle: I don't have a problem with that but I don't think it's necessary - you don't need the refs to check if there's an error.

adriana-alter: Okay, I've just committed an implementation in which `clearForm` lives in `sharedFunctions` and is called by the forms rather than by `EditableConfigList`...but I'm still working on having it only clear the form if there isn't an error.  Any suggestions on how to deal with that?  The things I've tried haven't been working; I'm kind of stuck.

EdwinGuzman: What have you tried?

adriana-alter: I tried passing the `fetchError` from `EditableConfigList` to the form as a prop, so that, inside `handleSubmit` in `sharedFunctions`, I would be able to check whether `form.props.fetchError` existed, and decide on that basis whether to clear the form.  The problem is (I think?) something to do with the sequence in which things are happening; the `fetchError` in `EditableConfigList` isn't accessible until the next time around.  As in, if I 1) submit a form with a field missing, and 2) submit a form the right way, after #1, `fetchError` is null, and after #2, `fetchError` is the error message from #1.  (I tried the same thing with `responseBody`; same problem.) The `fetchError` is in fact coming through when it's supposed to, as evidenced by the fact that the red error message still shows up at the top of the form...but I can't access it within the code.   Tried making various pieces of the process async, tried messing with the form's `componentWillReceiveProps`, etc.  Several hours later, still doesn't work...

src/components/IndividualAdminEditForm.tsx
```diff
@@ -290,27 +305,33 @@ export default class IndividualAdminEditForm extends React.Component<IndividualA
...
-        window.location.href = this.props.urlBase + "edit/" + this.props.responseBody;
-      }
-    });
+  submit(event) {
+    event.preventDefault();
+    handleSubmit(this);
+    // this.clearForm();
   }
+
+  // clearForm(err) {
```
EdwinGuzman: You also have some unneeded commented out code.

src/components/IndividualAdminEditForm.tsx
```diff
@@ -290,27 +304,25 @@ export default class IndividualAdminEditForm extends React.Component<IndividualA
...
-      // If we're setting up an admin for the first time, refresh the page
-      // to go to login.
-      if (this.context.settingUp) {
-        window.location.reload();
-        return;
-      }
+    // If we're setting up an admin for the first time, refresh the page
+    // to go to login.
+    if (this.context.settingUp) {
+     window.location.reload();
```
aslagle: This should not be part of `handleData` - I think this will cause the window to refresh before the data gets saved. You can test this by running the circ manager with a newly created, empty database and trying to create the first admin.

src/components/__tests__/SaveButton-test.tsx
aslagle: This is still true - the `handleData` stub isn't actually used by the component here.

src/sharedFunctions.ts
```diff
@@ -0,0 +1,40 @@
+// When a SaveButton is clicked or a form is otherwise submitted, collect and
```
aslagle: I might move this file under components/ since all the functions are related to components.

aslagle: I also don't see tests for this file? `findDefault` at least would be good to have tests for.

adriana-alter: Ok, wrote a test for findDefault.  The other two functions in that file are being tested within other tests.

src/components/IndividualAdminEditForm.tsx
adriana-alter: Yes.  Good catch; thank you.  I moved it.

src/components/IndividualAdminEditForm.tsx
```diff
@@ -290,27 +304,25 @@ export default class IndividualAdminEditForm extends React.Component<IndividualA
...
+  }
 
-      // If a new admin was created, go to its edit page.
-      if (!this.props.item && this.props.responseBody) {
-        window.location.href = this.props.urlBase + "edit/" + this.props.responseBody;
-      }
-    });
+  submit(event) {
+    event.preventDefault();
+    handleSubmit(this);
```
aslagle: I think you should do `await handleSubmit` here, so the reload happens after the request is successful. It's probably not too bad the way it is since the request has already been submitted, but if there's an error or for some reason the request takes a long time, you wouldn't want to reload and get the same initial setup form again.

EdwinGuzman:
Before you merge, you need to update the version in `package.json` and the updates in `changelog`.

PR #173, opened 2018-10-09
Added a color picker setting.

aslagle:
For now, this will be used for configuring background and foreground colors for the web catalog.

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -1,0 +1,50 @@
+import * as React from "react";
+import { CompactPicker } from "react-color";
+import { SettingData } from "../interfaces";
+
+export interface ColorPickerProps {
+  value?: string;
+  setting: SettingData;
+}
+
+export interface ColorPickerState {
+  value?: string;
+}
+
+export default class ColorPicker extends React.Component<ColorPickerProps, ColorPickerState> {
+  constructor(props) {
+    super(props);
+    this.state = { value: this.props.value };
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  render(): JSX.Element {
+    return (
+      <div className="color-picker">
+        <CompactPicker
+          color={this.state.value}
+          triangle="hide"
+          width="500px"
+          onChangeComplete={this.handleChange}
+          />
+        <input
+          type="hidden"
+          name={this.props.setting.key}
+          value={this.state.value}
+          />
+      </div>
+    );
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.setState({ value: nextProps.value });
+  }
+
+  handleChange(color, event) {
+    this.setState({ value: color.hex });
+  }
+
+  getValue(): string {
+    return this.state.value;
+  }
+}
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -36,10 +36,6 @@ export default class ColorPicker extends React.Component<ColorPickerProps, Color
     );
   }
 
-  componentWillReceiveProps(nextProps) {
-    this.setState({ value: nextProps.value });
-  }
-
   handleChange(color, event) {
     this.setState({ value: color.hex });
   }
--- a/src/components/ErrorMessage.tsx
+++ b/src/components/ErrorMessage.tsx
@@ -52,7 +52,7 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, voi
             { errorMessageHeader &&
               <p><b>{errorMessageHeader}</b>&nbsp;</p>
             }
-            {errorMessageText}<br />
+            <span dangerouslySetInnerHTML={{ __html: errorMessageText }} /><br />
             { this.props.tryAgain &&
               <a onClick={this.tryAgain.bind(this)}>Try again</a>
             }
--- a/src/components/ProtocolFormField.tsx
+++ b/src/components/ProtocolFormField.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import EditableInput from "./EditableInput";
 import WithRemoveButton from "./WithRemoveButton";
+import ColorPicker from "./ColorPicker";
 import { SettingData } from "../interfaces";
 
 export interface ProtocolFormFieldProps {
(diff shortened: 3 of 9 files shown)


src/components/ErrorMessage.tsx
```diff
@@ -52,7 +52,7 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, voi
             { errorMessageHeader &&
               <p><b>{errorMessageHeader}</b>&nbsp;</p>
             }
-            {errorMessageText}<br />
+            <span dangerouslySetInnerHTML={{ __html: errorMessageText }} /><br />
```
aslagle: I did this so that the server could send back a link to a contrast checker if the admin selected background and foreground colors that were too similar.

It might actually be dangerous though, since in some cases we might get an error message from a remote server and pass it straight through.

adriana-alter: Is there a way to make the check happen when the admin selects the second color, rather than on submit?  That way, if there's an error message, we know it must be related to the colors, and we would have the option of hard-coding it (to always just say something along the lines of "there isn't enough contrast between these colors") rather than passing it through.  Or (depending on how much freedom we want to give the admin) maybe even just disable the not-enough-contrast options for the second color once the admin has picked the first color.

aslagle: That would definitely be a much better user experience but also much more work to implement. The form fields are currently sent down from the server so the server would need some way to specify that these two fields are linked and have a particular relationship.

src/components/ProtocolFormField.tsx
```diff
@@ -164,6 +165,20 @@ export default class ProtocolFormField extends React.Component<ProtocolFormField
             />
         </div>
       );
+    } else if (setting.type === "color-picker") {
+      return (
+        <div>
+          <label>{setting.label + (setting.optional ? " (optional)" : "")}</label>
```
EdwinGuzman: I'd like to see explicitly something stating what the current value is besides what the color picker displays.

src/components/ColorPicker.tsx
```diff
@@ -0,0 +1,50 @@
...
+          type="hidden"
+          name={this.props.setting.key}
+          value={this.state.value}
+          />
+      </div>
+    );
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.setState({ value: nextProps.value });
```
EdwinGuzman: Do you need this because of the refresh after submitting the form?

src/components/ErrorMessage.tsx
EdwinGuzman: You could do a client side call before submitting the form to the server. You can try using webaim's API: https://webaim.org/resources/contrastchecker/, specifically `https://webaim.org/resources/contrastchecker/?fcolor=0000FF&bcolor=FFFFFF&api` (as stated at the bottom of the page).

Might be helpful if you want to give real-time feedback and avoid the `dangerouslySetInnerHTML`.

src/components/ProtocolFormField.tsx
```diff
@@ -164,6 +165,20 @@ export default class ProtocolFormField extends React.Component<ProtocolFormField
...
+          { setting.description &&
+            <span className="description" dangerouslySetInnerHTML={{__html: setting.description}} />
```
EdwinGuzman: Why does this need the `dangerouslySetInnerHTML` attribute?

aslagle: So it can contain links or other html tags for formatting. It's coming from a constant in the circ manager so it's safe in this case.

src/components/ErrorMessage.tsx
aslagle: The color library I'm using to generate the CSS variables in my other PR can also do contrast checking. The problem is I don't want to hard-code the relationship between these settings in the client, since otherwise all the setting information is configured on the server.

src/components/ProtocolFormField.tsx
EdwinGuzman: Yeah, but I missed it the first time after I selected it. It's not too obvious.

aslagle: I don't think I'd want to display it twice though - do you think I should use a simpler color picker but make a separate input for the hex value?

That library has a bunch of options, but the options aren't super-configurable: https://casesandberg.github.io/react-color/

adriana-alter: I agree that it would be nice to have the value more prominently displayed.  It looks like using CustomPicker instead of CompactPicker and putting in an EditableInput component might do the trick--under "Create Your Own" on the page you sent the link to.  (It lets you set the CSS styling for the EditableInput component, so you could make it as prominent as you want.)

aslagle: I tried the CustomPicker at first but it doesn't seem to let you make a grid of clickable colors like some of the non-custom ones, and I liked that UI better than sliders.

src/components/ColorPicker.tsx
EdwinGuzman: Hmm can the `component.update()` be used for tests to avoid this?

src/components/ProtocolFormField.tsx
adriana-alter: Agreed; it would be one thing if the CustomPicker were a viable option (it's strange and annoying that it won't let you make a grid; I would have expected grids to be way more popular than sliders!), but as it is, doesn't seem worth the potential complications of adding an extra input.

src/components/ColorPicker.tsx
adriana-alter: I'm pretty sure `component.update()` can only be used for the root instance; since the test needs to update the props for the child component, not for the wrapper, I don't think that would work...  (I don't see a problem with the way it's currently working, though!)

aslagle: Looks like this actually causes a bug where when you get an error it resets the colors to their original values instead of what you selected. I'm going to take it out.

aslagle: And I didn't actually need it for the tests anyway.

adriana-alter: Even better :)

src/components/ErrorMessage.tsx
EdwinGuzman: From your other PR, `tinycolor` can be used to check for WCAG color contrast for both "AA" and "AAA" standards and might be useful here if you want to check color contrast before send the info to the server.

PR #174, opened 2018-10-19
Display links to Confluence instructions

adriana-alter:
Along with https://github.com/NYPL-Simplified/circulation/pull/1075, resolves https://jira.nypl.org/browse/SIMPLY-1319

### DIFF
--- a/src/stylesheets/colors.scss
+++ b/src/stylesheets/colors.scss
@@ -6,4 +6,3 @@ $red-dark: #97272C;
 $red-error: tint($red, 80%);
 $light-gray: #D7D4D0;
 $dark-gray: #080807;
-$blue: #3A7DFD;
--- a/src/stylesheets/editable_input.scss
+++ b/src/stylesheets/editable_input.scss
@@ -8,7 +8,7 @@
 
 .collapsible {
   box-shadow: none;
-  border: 1px solid $blue;
+  border: 1px solid $linkcolor;
   border-radius: 0;
   .panel-heading {
     border: none;
@@ -17,7 +17,7 @@
     .panel-title {
       width: 100%;
       text-align: left;
-      background: $blue;
+      background: $linkcolor;
       padding: 4px 7px;
       border: none;
     }


adriana-alter:
@aslagle @EdwinGuzman Because of the compatibility issues with the new version of react-bootstrap, I've changed the Collapsible component to use an older version of Panel, and styled it to look like the new version of Panel.  (We should still deal with updating our version of DefinitelyTyped at some point, but that's a different conversation.)  Ready for review!

EdwinGuzman:
I would've liked that you kept the older style of the collapsible component since it ties it down to that specific version of react-bootstrap. But, you already did the work and it doesn't seem like a lot so it's fine. If we ever update, that code would be redundant and would have to be removed so it'd be good to note that as comments in the scss file.

src/components/Collapsible.tsx
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  toggle() {
+    let icon = this.state.icon === "chevron-down" ? "chevron-up" : "chevron-down";
+    this.setState({ open: !this.state.open, icon });
+  }
+
+  renderHeader() {
+    return (
+      <div onClick={this.toggle}>
```
EdwinGuzman: Let's not have `onClick`s on elements that may not support them like divs. It works but it'll be better if this was on the `<button>` itself. Currently, when I use VoiceOver, it let's me toggle the dopdown with both the header and the button. There might be a better way to do this. (This also produces "redundant links" since two elements that are close to each other perform the same action).

Also, there's no focus ring around either the header or the button when you tab through them and that's confusing.

adriana-alter: I decided to put the `onClick` on the header because that seemed better for accessibility (the button is kind of a small target)...no?

I'll put the focus rings in!

EdwinGuzman: It's not just about the size of the target but how the HTML elements interact with screenreaders. Technically, you can add `role="button"` to the div along with the onClick handler but then you might as well use a button for that. It's always preferable to use native elements so they can be supported on all browsers and having the onclick on the div may not be supported everywhere (IE 👀 ).

adriana-alter: Ah, that makes sense.  What do you think--should I 1) make the whole header be a button element (and CSS it so as to keep the current styling), or 2) have the `onClick` be just on the existing button?

EdwinGuzman: Let let me check the react-bootstrap doc for this version and see what they natively support and do. But making the whole header a button might be the way to go.

src/stylesheets/editable_input.scss
```diff
@@ -6,6 +6,34 @@
...
+.collapsible {
+  margin-top: 20px;
+  box-shadow: none;
+  .panel-heading {
+    border-bottom: none;
+    font-weight: bold;
+    .collapsed, .panel-title a {
+      color: inherit;
+      &:visited {
+        color: #333;
```
aslagle: It would be better to use variables for the colors, ideally variables that already exist.

aslagle:
I'm not sure about the styling of the button/title but the code looks good.

adriana-alter:
I'm not sure about the styling either.  Open to suggestions!

EdwinGuzman:
I think it looks fine, plain and simple. This looks much better to me 👍

EdwinGuzman:
I am concern about using css icons and would prefer SVG since they are more accessible but not a big deal. For the next one.

Something like the SVG icons at NYPL might be useful: https://nypl.github.io/dgx-svg-icons/#genericwedgeicon

aslagle:
@mgiraldo suggested making it look more like a link

adriana-alter:
Making it look like a link seems really confusing (since it then turns out not to be a link)...  Opinions on making the background blue, like an info button (which would make it look slightly more link-y)?

adriana-alter:
Re-styled as per input from Jack!  Blue background and border; plus/minus signs instead of chevrons; label saying "Instructions."

screen shot 2018-10-25 at 11 58 22 am

screen shot 2018-10-25 at 11 58 44 am

src/stylesheets/colors.scss
```diff
@@ -6,3 +6,4 @@ $red-dark: #97272C;
 $red-error: tint($red, 80%);
 $light-gray: #D7D4D0;
 $dark-gray: #080807;
+$blue: #3A7DFD;
```
aslagle: It would be better to use `$linkcolor` which already exists in opds-web-client and is this exact color.

The reds and greens make sense here because they are for error/success indicators rather than part of the theming. I'm not sure what the grays are used for, those should maybe be changed (not as part of this PR, just in general).

PR #177, opened 2018-10-26
Update to 0.0.82

adriana-alter:


aslagle:
You should also update the CHANGELOG here.

PR #178, opened 2018-10-26
Revert "Update to 0.0.82"

adriana-alter:
In the change log: skip straight from 0.0.80 to 0.0.82, or list changes for 0.0.82 just as "bug fix" or something?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 ## Changelog
 
+### v0.0.82
+#### Fixed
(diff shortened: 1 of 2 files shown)


aslagle:
I'd put 0.0.82 as bug fix and maybe say under 0.0.81 that it has a bug and you shouldn't use it, update to 0.0.82.

aslagle:
This looks good - you shouldn't revert the version though.

adriana-alter:
Ah.  That wasn't on purpose.  Fixed it.

PR #179, opened 2018-10-26
Required form fields

EdwinGuzman:
Fixes [SIMPLY-1318](https://jira.nypl.org/browse/SIMPLY-1318) along with [server-core #979](https://github.com/NYPL-Simplified/server_core/pull/979) and [circulation #1084](https://github.com/NYPL-Simplified/circulation/pull/1084).

Updated the prop that read a field's "optional" value to read its "required" value. The label should only appear for text inputs as it doesn't seem necessary for selects (since there's a default) or checkboxes or radio inputs.
screen shot 2018-10-26 at 5 02 38 pm

If a field is optional and not a radio, select, or checkbox, then the description will render `(Optional) ...` before the field's description. _Note:_ I did not include this in the forms for the book editor. This includes the metadata and book cover forms for individual books. It seems like that could be a bigger or different problem (or could be update in this PR if necessary).

When an admin submits and there are any errors, the focus will go to the top of the page onto the alert message block. This is a pattern used in the [Design Toolkit](https://github.com/NYPL/design-toolkit). I also copied over the error styling from the DT instead of importing the whole package (perhaps something to worry about later on).


src/components/EditableInput.tsx
```diff
@@ -1,10 +1,14 @@
...
+import { FetchErrorData } from "opds-web-client/lib/interfaces";
 
 export interface EditableInputProps extends React.HTMLProps<EditableInput> {
   elementType?: string;
   label?: string;
   description?: string;
   onChange?: (e: any) => any;
+  required?: boolean;
+  error?: FetchErrorData;
+  optionalText?: boolean;
```
aslagle: Why do you need both `required` and `optionalText`?

src/components/EditableInput.tsx
```diff
@@ -27,22 +35,30 @@ export default class EditableInput extends React.Component<EditableInputProps, E
...
 
   render() {
+    const checkboxOrRadioOrSelect = !!(
+      this.props.type === "checkbox" ||
+      this.props.type === "radio" ||
+      this.props.elementType === "select");
+    const optionalText = (this.props.optionalText && !this.props.required && !checkboxOrRadioOrSelect)
+      ? "(Optional)" : "";
+    const descriptionText = this.props.description ? this.props.description : "";
+    const description = `${optionalText} ${descriptionText}`;
```
aslagle: You could make the space part of `optionalText` so you don't get an extra space if it's empty.

src/components/EditableInput.tsx
```diff
@@ -27,22 +35,30 @@ export default class EditableInput extends React.Component<EditableInputProps, E
...
+    const errorClass = (this.props.error && this.props.error.status >= 400 &&
+      !this.state.value && this.props.required) ?
+      "field-error" : "";
```
aslagle: Nice! I was thinking the server would have to specify what field was bad, but this is a lot easier.

src/components/ProtocolFormField.tsx
```diff
@@ -166,26 +179,29 @@ export default class ProtocolFormField extends React.Component<ProtocolFormField
...
+            required={setting.required}
             name={setting.key}
             description={setting.description}
             accept="image/*"
+            error={this.props.error}
             />
         </div>
       );
     } else if (setting.type === "color-picker") {
+      let colorPickerDescription = `(Optional) ${setting.description}`;
```
aslagle: This is not really optional since there's no way to not pick something - it would submit the default. It's more like a select or checkbox in that way.

src/interfaces.ts
```diff
@@ -175,6 +175,7 @@ export interface SettingData {
   description?: string;
   default?: string | string[];
   optional?: boolean;
```
aslagle: Should `optional` be removed?

src/components/EditableInput.tsx
adriana-alter: Seconded.  And I think it might be more readable to just use `required` everywhere.

src/components/ErrorMessage.tsx
```diff
@@ -60,6 +61,18 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, voi
         </Alert>
       );
     }
+
+    return (
+      <div tabIndex={0} ref="errorMessage">
```
adriana-alter: What is this doing?

src/components/ErrorMessage.tsx
```diff
@@ -60,6 +61,18 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, voi
...
+        {alertElement}
+      </div>
+    );
+  }
+
+  componentDidMount() {
+    if (this.refs["errorMessage"]) {
+      ReactDOM.findDOMNode<HTMLDivElement>(this.refs["errorMessage"]).focus();
```
adriana-alter: What's the purpose of focusing on the error message?  The focus ring kind of makes it look like the error message component is an input field (or at least like something will happen if you click on it).  I think it would be enough to just have the form automatically scroll to the top of the page if there's an error message (the way that the Site Configuration forms now do).

src/components/ProtocolFormField.tsx
```diff
@@ -52,9 +54,11 @@ export default class ProtocolFormField extends React.Component<ProtocolFormField
             type="text"
             disabled={this.props.disabled}
             name={setting.key}
-            label={setting.label + (setting.optional ? " (optional)" : "")}
```
adriana-alter: "Optional" is capitalized in the `EditableInput` labels but lowercased here.  (Capitalized would be better, since you have "Required" capitalized, but either way let's have it be consistent.)

src/components/EditableInput.tsx
```diff
@@ -27,22 +35,30 @@ export default class EditableInput extends React.Component<EditableInputProps, E
   }
 
   render() {
+    const checkboxOrRadioOrSelect = !!(
+      this.props.type === "checkbox" ||
+      this.props.type === "radio" ||
+      this.props.elementType === "select");
+    const optionalText = (this.props.optionalText && !this.props.required && !checkboxOrRadioOrSelect)
```
adriana-alter: It might make more sense to have `checkboxOrRadioOrSelect` be part of the definition of `required` in the first place--if a field is a checkbox/radio/select (i.e. something with a default), then there should be no way that `this.props.required` would ever be true, so this check wouldn't be necessary.

src/components/ProtocolFormField.tsx
EdwinGuzman: No, this is the label which will not have `${thing} (optional)` and instead we will have `(Optional) ${thing}` in the description as Jack suggested.

src/components/EditableInput.tsx
EdwinGuzman: Right, I think I may need to rethink the required prop for these types of inputs.

src/components/EditableInput.tsx
EdwinGuzman: By default the input will display `(Optional) ...` for its description. But from the short discussion on the book cover form or the book metadata form, having `(Optional) ...` in those text input fields seemed wrong. So it's to weird, but to further give the option to display `(Optional) ` in the description.

src/components/EditableInput.tsx
EdwinGuzman: I would really like to have the server specify what fields are bad, and not just one at a time, so we can highlight the error-ed ones. Similar to what's here: https://nypl.github.io/design-toolkit/sections/forms.html#formErrors
I think that's a stretch for now so maybe later on.

src/components/EditableInput.tsx
aslagle: That makes sense. Would it be better to have the default be false, since the ones that should have it be true mostly use `ProtocolFormField` already and wouldn't require setting it in as many places?

EdwinGuzman: It's about the same number of `<EditableInput>` that would need to be updated if the default is false instead of true. I'm going to leave it like this and see if it should be updated later when I work on grouping the fields together.

PR #180, opened 2018-10-29
Checking if a setting configuration supports library staging registration.

EdwinGuzman:
Fixes [SIMPLY-1344](https://jira.nypl.org/browse/SIMPLY-1344).

The shared ODL collection registration for libraries should not have UI controls for selecting what type of staging level the library should be in. This checks to see if the library supports staging selection and renders the proper controls if it does.

### DIFF
--- a/src/components/LibraryRegistration.tsx
+++ b/src/components/LibraryRegistration.tsx
@@ -24,7 +24,9 @@ export default class LibraryRegistration extends React.Component<LibraryRegistra
   }
 
   render() {
-    if (this.props.item && this.protocolSupportsRegistration() && this.props.data && this.props.data.allLibraries && this.props.data.allLibraries.length > 0) {
+    const supportsStaging = this.protocolSupportsType("supports_staging");
+    if (this.props.item && this.protocolSupportsType("supports_registration") &&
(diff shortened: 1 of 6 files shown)


src/components/LibraryRegistration.tsx
```diff
@@ -137,11 +141,11 @@ export default class LibraryRegistration extends React.Component<LibraryRegistra
     return libraryInfo && libraryInfo[prop];
   }
 
-  protocolSupportsRegistration(): boolean {
+  protocolSupportsType(prop: string): boolean {
     if (this.state.protocol && this.props.data.protocols) {
       for (const protocol of this.props.data.protocols) {
         if (protocol.name === this.state.protocol) {
-          return !!protocol.supports_registration;
+          return !!protocol[prop];
```
aslagle: I think you're losing some Typescript checking by using the string here, but that's okay.

EdwinGuzman: Oh that's interesting. How so?

aslagle: If you check `protocol.support_registration` Typescript will verify that `supports_registration` is defined in the type. If you pass in an arbitrary string I don't think it will.

EdwinGuzman: Hmmmm good to point out.

PR #181, opened 2018-10-31
Group a11y fields

EdwinGuzman:
First pass at fixing [SIMPLY-1317](https://jira.nypl.org/browse/SIMPLY-1317).

Wrapping most input fields inside different forms with `<fieldset>` if it made sense to include a title and to group them together. Added `visuallyHidden` css rules to hide most fieldset `legend` elements because they are not needed, but are useful for screen readers.

Not much really changes except the DOM structure. I did not tackle `LibraryEditForm `, `ProtocolFormField`, or `ServiceEditForm` but will attempt for the next pass where I also hope to group together required and non-required fields together.


src/components/BookEditForm.tsx
```diff
@@ -31,174 +31,177 @@ export default class BookEditForm extends React.Component<BookEditFormProps, Boo
...
+                      />
+                  </span>
+                </WithRemoveButton>
+              )
+            }
+            <span className="contributor-form">
+              <EditableInput
+                elementType="select"
                 disabled={this.props.disabled}
-                onRemove={() => this.removeContributor(contributor)}
```
adriana-alter: Did you mean to delete this line (and if so, why)?

EdwinGuzman: No, I did not. But I don't remember removing that line and `master` doesn't seem to have this line, either. Not sure what's going on with this line. `EditableInput` doesn't have an `onRemove` prop anyway so this shouldn't be a problem.

EdwinGuzman: 😨 👻

src/components/BookCoverEditor.tsx
```diff
@@ -131,34 +133,36 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, void>
...
-                    label="License"
-                    >
-                    { Object.keys(this.props.rightsStatuses).map(uri => {
-                          let status = this.props.rightsStatuses[uri];
-                          if (status.allows_derivatives) {
-                            return (
-                              <option value={uri} key={uri}>{status.name}</option>
-                            );
+                  <fieldset>
+                    <legend><h4>Rights:</h4></legend>
```
aslagle: I like this HTML structure but I think the styling could be improved. Looks like now there's a horizontal line below the legend but nothing to separate it from the inputs above it? Admittedly the styling was bad before too, so I don't think it needs to be fixed in this branch.

src/components/BookEditForm.tsx
aslagle: I think github is just confused by so many lines with indentation changes. This line is now line 60 - it was moved, not deleted.

src/components/ChangePasswordForm.tsx
```diff
@@ -40,7 +40,6 @@ export class ChangePasswordForm extends React.Component<ChangePasswordFormProps,
   render(): JSX.Element {
     return (
       <div className="change-password-form">
-        <h2>Change Password</h2>
```
aslagle: Do we want the error and success message above the header now?

src/components/ComplaintForm.tsx
```diff
@@ -35,20 +35,22 @@ export default class ComplaintForm extends React.Component<ComplaintFormProps, a
...
-            elementType="select"
-            ref="type"
-            name="type"
-            placeholder=""
-            disabled={this.props.disabled}>
-            <option value="">complaint type</option>
-            { complaintTypes.map(type => <option key={type} value={type}>{type}</option>) }
-          </EditableInput> &nbsp;
+          <fieldset>
+            <legend><h3>Add Complaint</h3></legend>
```
aslagle: It would be nice if the "Complaints" header in Complaints.tsx was styled the same as this.

src/components/CustomListEditor.tsx
```diff
@@ -61,12 +61,15 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
-              />
+            <fieldset>
+              <legend className="visuallyHidden">List name</legend>
+              <TextWithEditMode
+                text={listTitle}
+                placeholder="list title"
+                onUpdate={this.changeTitle}
+                ref="listTitle"
+                />
+            </fieldset>
```
aslagle: This vertical alignment is a little messed up for me.
screen shot 2018-10-31 at 5 57 19 pm

aslagle: Also, should not be optional.

aslagle: Maybe that's something we missed in the last branch.

src/components/SitewideSettingEditForm.tsx
```diff
@@ -42,61 +42,64 @@ export default class SitewideSettingEditForm extends React.Component<SitewideSet
...
-              disabled={this.props.disabled}
-              readOnly={!!(this.props.item && this.props.item.key)}
-              name="key"
-              ref="key"
-              label="key"
-              value={this.props.item && this.props.item.key}
-              onChange={this.onChange}
-            >
+            <fieldset>
+              <legend className="visuallyHidden">Select key value pair for sitewide setting</legend>
```
aslagle: This description isn't quite right - you select the key, but enter the value. And if you're editing a setting, you can't change the key, only the value.

src/components/SitewideSettingEditForm.tsx
```diff
@@ -42,61 +42,64 @@ export default class SitewideSettingEditForm extends React.Component<SitewideSet
...
+              <EditableInput
+                elementType="select"
+                disabled={this.props.disabled}
+                readOnly={!!(this.props.item && this.props.item.key)}
+                name="key"
+                ref="key"
+                label="key"
```
aslagle: While we're here, this should be capitalized.

src/stylesheets/mixins.scss
```diff
@@ -9,3 +9,31 @@ $white: #FFF !default;
 @function tint($color, $percentage) {
   @return mix($white, $color, $percentage);
 }
+
+.visuallyHidden {
```
aslagle: This isn't actually a mixin. Is there another file we could put it in?

src/stylesheets/mixins.scss
```diff
@@ -9,3 +9,31 @@ $white: #FFF !default;
...
+    clip: auto;
+    height: auto;
+    margin: 0;
+    overflow: visible;
+    position: static;
+    width: auto;
+  }
+}
+
+.displayVisuallyHidden {
```
aslagle: I don't think you used this?

src/components/BookEditForm.tsx
EdwinGuzman: Logic > ghosts

src/components/ChangePasswordForm.tsx
EdwinGuzman: I wasn't sure about that. Semantically I think this is better, but visually maybe not.

src/components/BookEditForm.tsx
adriana-alter: 👻

src/components/ChangePasswordForm.tsx
adriana-alter: In all of the system config forms, the error/success message is below the header.  (Not that that's necessarily the right way to do it, but I think let's just keep everything consistent for now and address this sort of thing in a different branch.)

src/components/CustomListEditor.tsx
adriana-alter: There are a few other things about this interface that I'm concerned are potentially confusing and/or slightly visually off; how about I just make a separate ticket devoted to optimizing the list form UI (with Jack's input) which will, among other things, fix this vertical alignment?  (I do think the `optional` thing should be fixed ASAP, though; it makes sense that that's an issue from the last branch, in which case it would be a really easy fix...right...?)

src/components/SitewideSettingEditForm.tsx
adriana-alter: Seconded.  How about "Enter or edit a value for the selected sitewide setting key"?  (Or will people not then understand that they need to select a key if it's a new setting?)

src/components/BookCoverEditor.tsx
adriana-alter: Styling idea: what about grouping related fields by putting them into--drumroll, please!--collapsible panels?  No, really--check out the Bootstrap PanelGroup Accordion component: https://react-bootstrap.github.io/components/panel/#panels-accordion.  That would 1) make the styling decisions a lot easier, and 2) more importantly, reduce the amount of information being thrown at the user all at once (as per Leonard's suggestion in the Jira ticket).

src/components/CustomListEditor.tsx
aslagle: Hopefully, and the presence of optional might be causing the vertical alignment problem too.

src/components/BookCoverEditor.tsx
aslagle: That sounds good as long as there is something to indicate when the panel contains a required input, so people don't think they can skip it since it's collapsed.

src/components/CustomListEditor.tsx
EdwinGuzman: Yea, the optional text is causing the misalignment. I fixed it for a different input but not this one.

src/components/BookCoverEditor.tsx
EdwinGuzman: I can try that for the next PR. Grouping required and non-required fields together may be trickier than I expected (at least trying to do it on the server).

src/components/ComplaintForm.tsx
EdwinGuzman: That could be a different ticket where we address the styling for the legend and headings. But I'll remove the `H3` heading in the legend to make it smaller than the actual heading.

src/components/SitewideSettingEditForm.tsx
EdwinGuzman: That sounds good. It will be for screenreaders but it's understandable.

EdwinGuzman:
Updated.

adriana-alter:
Looks good to me!

PR #183, opened 2018-11-06
Group fields by type

EdwinGuzman:
This is based off and pointing to the current open PR #181 .
The second pass to fix [SIMPLY-1317](https://jira.nypl.org/browse/SIMPLY-1317).

Updated the collapsible component so that it would render a static panel component for required fields and collapsible for optional fields. Also keeping the instruction styling scss rules.

Updated some tests from `shallow` to `mount` because there is now an intermediary component that needs to be rendered before some `EditableInput` components are rendered. I did not update every single form because some, like the change password form, only have two inputs and it didn't seem necessary.

Preview:
screen shot 2018-11-06 at 5 05 55 pm

### DIFF
--- a/src/components/BookCoverEditor.tsx
+++ b/src/components/BookCoverEditor.tsx
@@ -52,9 +52,10 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, void>
   renderCoverForm() {
     return (
       <div>
+        <p>Cover must be at least 600px x 900px and in PNG, JPG, or GIF format.</p>
         <form ref="cover">
           <fieldset>
-            <legend className="visuallyHidden">Cover Metadata</legend>
+            <legend className="visuallyHidden">Cover Image</legend>
             <EditableInput
               elementType="input"
               type="text"
@@ -140,7 +141,6 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, void>
             </div>
             <div>
               <h3>Change cover:</h3>
-              <p>Cover must be at least 600px x 900px and in PNG, JPG, or GIF format.</p>
               <Collapsible
                 title="Cover Metadata"
                 openByDefault={true}
--- a/src/components/BookCoverEditor.tsx
+++ b/src/components/BookCoverEditor.tsx
@@ -46,14 +46,15 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, void>
     this.refresh = this.refresh.bind(this);
     this.preview = this.preview.bind(this);
     this.save = this.save.bind(this);
-    this.renderForm = this.renderForm.bind(this);
+    this.renderCoverForm = this.renderCoverForm.bind(this);
   }
 
-  renderForm() {
+  renderCoverForm() {
     return (
       <div>
         <form ref="cover">
           <fieldset>
+            <legend className="visuallyHidden">Cover Metadata</legend>
             <EditableInput
               elementType="input"
               type="text"
@@ -109,40 +110,6 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, void>
               />
           </div>
         }
-        { this.props.rightsStatuses &&
-          <form ref="rights">
-            <fieldset>
-              <legend><h4>Rights:</h4></legend>
-              <EditableInput
-                elementType="select"
-                disabled={this.props.isFetching}
-                name="rights_status"
-                label="License"
-                >
-                { Object.keys(this.props.rightsStatuses).map(uri => {
-                      let status = this.props.rightsStatuses[uri];
-                      if (status.allows_derivatives) {
-                        return (
-                          <option value={uri} key={uri}>{status.name}</option>
-                        );
-                      }
-                      return null;
-                    }
-                  )
-                }
-                <option value="http://librarysimplified.org/terms/rights-status/in-copyright">In Copyright</option>
-                <option value="http://librarysimplified.org/terms/rights-status/unknown">Other</option>
-              </EditableInput>
-              <EditableInput
-                elementType="textarea"
-                disabled={this.props.isFetching}
-                name="rights_explanation"
-                label="Explanation of rights"
-                optionalText={false}
-              />
-            </fieldset>
-          </form>
-        }
       </div>
     );
   }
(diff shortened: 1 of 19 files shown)


adriana-alter:
I don't think the text on the header for the collapsible panel should be blue; that makes it look like a link.  I'm not sure what color it should be instead, though--maybe just black?

EdwinGuzman:
How about leaving the panel in for the metadata in the book editor for aesthetics?

aslagle:
Maybe we should get rid of the tabs on the book details page and switch to panels?

adriana-alter:
I agree about making the required fields panel collapsible.

Let's leave the book editor alone for now; I figured that the epic that this is part of would only be about the system configuration forms anyway.  That being said, I think switching to panels on the book details page is a great idea (both because it will make the workflow clearer and for the sake of consistency now that we're using panels in so many other places).  Just not in this ticket.

I'm still not entirely sold on the text being blue.  Let's see what Jack thinks?  I'm not sure what you mean about the anchor element...

EdwinGuzman:
Alright, updated the component so that it's alway collapsible but can be open by default.

EdwinGuzman:
@jackkim9 , Adriana is not too keen on the blue for the text in the collapsible field header. What do you think?

screen shot 2018-11-07 at 1 00 21 pm

adriana-alter:
(I'm not keen on the blue because I think that makes it look like a link, and then when you click on it it turns out not to be a link, and that's potentially confusing; this isn't an aesthetic objection.)

christhepianist:
The blue does look like a link. I would try to click on it. If you want to highlight it and not link it, I’d try doing it differently.


Christine Peterson

EdwinGuzman:
Talked to Jack about this and he suggested making the header color black, keeping the plus/minor icon blue, and updating the background color for the hover state for the entire heading. From the screenshot below, the "Required Fields" header is being hovered.

screen shot 2018-11-07 at 5 04 30 pm

adriana-alter:
That looks good to me!

jackkim9:
LGTM

src/components/BookCoverEditor.tsx
```diff
@@ -74,96 +141,50 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, void>
...
-                  <h4>Preview:</h4>
-                  <img
-                    src={this.props.preview}
-                    className="book-cover preview-cover"
-                    alt="Preview of new cover"
-                    />
-                </div>
-              }
+              <Collapsible
+                title="Cover Metadata"
```
aslagle: This section isn't really metadata - it's the cover image and optional settings used to create the image. Maybe change to "Cover Image"?

src/components/Collapsible.tsx
```diff
@@ -26,18 +29,41 @@ export default class Collapsible extends React.Component<CollapsibleProps, Colla
 
   renderHeader() {
     let icon = this.state.open ? "minus" : "plus";
-    return (
-      <button bsStyle="default" onClick={this.toggle}>
-        <span>{this.props.title}</span>
+    const { type, title } = this.props;
+    let element = "div";
```
aslagle: Is using a div instead of a button for styling or semantics here?

src/components/BookCoverEditor.tsx
```diff
@@ -74,96 +141,50 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, void>
             <div>
               <h3>Change cover:</h3>
               <p>Cover must be at least 600px x 900px and in PNG, JPG, or GIF format.</p>
```
aslagle: Can this text go in the panel? Just the `<p>`, not the `<h3>`.

src/components/CustomListEditor.tsx
```diff
@@ -112,29 +113,35 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
-                this.props.entryPoints.length ? (
-                  <div className="entry-points">
-                    <span>Select the entry point to search for:</span>
-                    <div className="entry-points-selection">
-                      {this.getEntryPointsElms(this.props.entryPoints)}
-                    </div>
-                  </div>
-                ) : null
+            <Collapsible
+              title="Search"
```
aslagle: I would change the title to "Search for titles" and make the legend visually hidden on this one, so it doesn't need as much vertical space. Right now when it's expanded none of the search results fit on my screen.

It would also be great to put the collections on this page in a panel that is closed by default. Very few people (probably only me for Open Bookshelf) will need that part and it can take up a lot of space.

src/components/IndividualAdminEditForm.tsx
```diff
@@ -155,6 +156,18 @@ export default class IndividualAdminEditForm extends React.Component<IndividualA
...
+    );
+  }
+
+  render(): JSX.Element {
+    return (
+      <form ref="form" onSubmit={this.submit} className="edit-form">
+        <Collapsible
+          title="Admin Information"
+          openByDefault={true}
+          body={this.renderForm()}
```
aslagle: For this one how about putting roles in a separate panel?

src/components/Collapsible.tsx
EdwinGuzman: Right. It's a bit more semantic. From a previous comment, the `Panel` component renders its header prop as a wrapper to an anchor element. Using a button looks like:
```
<button ...>
   <a ....>Instructions</>
</button>
```
This causes an extra tab when tabbing through. A div may be better but I didn't want to also change the existing component too much. I can make it consistent and use a div instead of button.

src/stylesheets/collapsible.scss
```diff
@@ -0,0 +1,55 @@
...
+    .collapsed, .panel-title a {
+      color: white;
+      &:focus {
+        outline: none;
+        text-decoration: none;
+      }
+      .glyphicon {
+        float: right;
+        line-height: inherit;
+        color: $WHITE;
```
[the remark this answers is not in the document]
EdwinGuzman: Hmm I don't get that and Travis didn't complain, but I'll add.

src/components/BookCoverEditor.tsx
EdwinGuzman: Yes.

EdwinGuzman:
Updated based on feedback.

src/stylesheets/collapsible.scss
```diff
@@ -1,3 +1,5 @@
+$panel-heading-color: #F5F5F5;
```
adriana-alter: This is not a big deal, but: if you were to let the panel heading be #f5f5f4 instead (which is--at least to me...!--indistinguishable from #f5f5f5), you could use the already-existing $gray variable.

adriana-alter:
How about further subdividing the library form fields?  There are a LOT of optional fields--it's still pretty overwhelming.  At a minimum, let's separate the stuff involving searching/borrowing books from the list of URLs.

src/components/IndividualAdminEditForm.tsx
adriana-alter: That looks really nice!

aslagle:
I agree about subdividing the library fields but I think it should be a separate PR. The way it is now is already a huge improvement. I would say the grouping should ultimately be passed down from the server - maybe each field could also have a section.

adriana-alter:
Having a collapsible inside a collapsible (i.e. the instructions panels in analytics & admin auth) looks weird to me.  How about moving the instructions panel to the top level--i.e. for admin auth, there would be three separate panels: 1) required fields, 2) instructions, and 3) libraries?

adriana-alter:
For the things where there's only the required fields panel--CDN, search, storage, etc.--I'm not sure there's a point to having a collapsible panel.  If the goal was just to keep the styling consistent, how about putting the fields in a non-collapsing panel, without the toggle icon?

adriana-alter:
(You probably know this, but just in case it slipped your mind:) Register Libraries in Discovery still needs to go into a panel.

src/stylesheets/collapsible.scss
EdwinGuzman: ha...

EdwinGuzman:
The Register Libraries is its own separate thing and not part of the form since it's after the submit button.  So I did not include the pane there.

EdwinGuzman:
I'm going to include the instructions above the required fields because it doesn't make sense to include them after the required fields.

screen shot 2018-11-08 at 2 42 34 pm

src/components/Collapsible.tsx
EdwinGuzman: I decided to change it back to a div. That way there's less styling issues to worry about for the static panel since it won't have the on hover or cursor when hovering.

EdwinGuzman:
Non-collapsible static panel added for forms that only have one field on the page.

src/components/CustomListEditor.tsx
```diff
@@ -111,30 +93,60 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
+                    }
                   </div>
-                ) : null
+                }
+              />
+            </div>
+          }
+          <form className="form-inline" onSubmit={this.search}>
+            <Collapsible
+              title="Search for titles"
```
aslagle: I like the way this looks now but when I type into the search box and hit enter it closes the panel instead of doing the search.

adriana-alter: That seems to be something inherent to the collapsible panels; I noticed yesterday that when I type into an input field on a form with an instructions panel and hit enter, it toggles the panel instead of submitting the form.  I was already planning to take a look at that soon-ish, so I don't think it needs to be part of this PR.

adriana-alter: Update: resolved by https://github.com/NYPL-Simplified/circulation-web/pull/186.  (Apparently you need to explicitly set the type of the collapsible's button to "button"; otherwise it defaults to "submit" and makes this happen.)

PR #184, opened 2018-11-07
Register library

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1352.

(The styling change is just so that the current-stage field is the same height as the button next to it; it looks off-kilter otherwise.)

### DIFF
--- a/src/components/CDNServices.tsx
+++ b/src/components/CDNServices.tsx
@@ -24,7 +24,7 @@ function mapStateToProps(state, ownProps) {
   const data = Object.assign({}, state.editor.cdnServices && state.editor.cdnServices.data || {});
   return {
     data: data,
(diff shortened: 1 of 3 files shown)


adriana-alter:
I had completely forgotten about updating the version number and changelog.  Fixed!

adriana-alter:
Actually, while I'm at it: I think it would be nice to have the library registration form link to the edit forms for the individual libraries (for the sake of convenience if you try to register a library and turn out to be missing one of the relevant fields, like I repeatedly did this morning!).  What do you think?

aslagle:
Sounds good. You could also add a test for that last commit.

EdwinGuzman:
Updates look good.

PR #188, opened 2018-11-16
Group a11y fields

adriana-alter:
Merged master into this branch, and fixed the ensuing problems caused by the branch needing to be updated to use the refactored version of the Collapsible component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 ## Changelog
 
+### v0.0.86
+#### Fixed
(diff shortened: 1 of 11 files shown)


aslagle:
I'm confused about what this is since it looks like you already merged #181 into master?

adriana-alter:
I did merge it, but that turned out to have been a really bad idea, since the most recent changes in master are apparently incompatible with this branch.  I took it out and updated it, and now want to merge the updated version into master.

PR #190, opened 2018-11-19
Welcome page

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1283; displays a welcome message if there are no libraries yet.


src/components/__tests__/WelcomePage-test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    let header = wrapper.find(Header);
+    expect(header.length).to.equal(1);
+  });
+
+  it("shows the welcome message", () => {
+    let jumbotron = wrapper.find(Jumbotron);
+    expect(jumbotron.length).to.equal(1);
+    let title = jumbotron.find("h2");
+    expect(title.length).to.equal(1);
+    expect(title.text()).to.equal("Welcome to the admin interface!");
```
EdwinGuzman: Maybe "Circulation Admin interface"?

EdwinGuzman:
Remember to also update the version in package.json and the changelog file.

adriana-alter:
This doesn't seem to warrant its own version--I figured I would wait and get one or two more little things in there before I went changing the version number...no?

EdwinGuzman:
If that's the case then we can create a workflow where we have a feature branch where we merge multiple changes to before merging that feature branch into master. Just so we can always update the version whenever we push to master for consistency.

PR #192, opened 2018-11-21
Style sign-up form, and make it show error messages and require password

adriana-alter:
v0.0.88 hasn't been published yet, so we can add this to it rather than creating a new version...yes?

### DIFF
--- a/src/components/EditableConfigList.tsx
+++ b/src/components/EditableConfigList.tsx
@@ -26,6 +26,7 @@ export interface EditableConfigListOwnProps {
   csrfToken: string;
   editOrCreate?: string;
   identifier?: string;
+  settingUp?: boolean;
 }
 
 export interface EditableConfigListProps<T> extends EditableConfigListStateProps<T>, EditableConfigListDispatchProps<T>, EditableConfigListOwnProps {}
@@ -73,9 +74,19 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
   render(): JSX.Element {
     let EditForm = this.EditForm;
     let AdditionalContent = this.AdditionalContent || null;
+    let headers = this.getHeaders();
+
+    let className = "";
+    if (AdditionalContent) {
+      className = "has-additional-content";
+    }
+    else if (this.props.settingUp) {
+      className = "set-up";
+    }
+
     return (
-      <div className={AdditionalContent ? "has-additional-content" : ""}>
-        <h2>{this.getItemType()} configuration</h2>
+      <div className={className}>
+        <h2>{headers["h2"]}</h2>
         { this.props.responseBody && this.props.editOrCreate &&
           <Alert bsStyle="success">
             {this.successMessage()}
@@ -137,7 +148,7 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
         }
         { (this.props.editOrCreate === "create") &&
           <div>
-            <h3>Create a new {this.itemTypeName}</h3>
+            <h3>{headers["h3"]}</h3>
             <EditForm
               ref="edit-form"
               data={this.props.data}
@@ -173,6 +184,12 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
 
   label(item): string {
     return item[this.labelKey];
+  }
+
+  getHeaders() {
+    let h2 = this.props.settingUp ? "Welcome!" : `${this.getItemType()} configuration`;
+    let h3 = this.props.settingUp ? "Set up your system admin account" : `Create a new ${this.itemTypeName}`;
+    return { h2, h3 };
   }
 
   getItemType() {
@@ -226,7 +243,13 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
 
   save(data: FormData) {
     this.editItem(data).then(() => {
-      if (this.limitOne && this.props.editOrCreate === "create") {
+      // If we're setting up an admin for the first time, refresh the page
+      // to go to login.
+      if (this.props.settingUp) {
+       window.location.reload();
+       return;
+      }
+      else if (this.limitOne && this.props.editOrCreate === "create") {
         // Wait for two seconds so that the user can see the success message,
         // then go to the edit page
         setTimeout(() => {
--- a/src/components/EditableConfigList.tsx
+++ b/src/components/EditableConfigList.tsx
@@ -23,7 +23,7 @@ export interface EditableConfigListDispatchProps<T> {
 
 export interface EditableConfigListOwnProps {
   store?: Store<State>;
-  csrfToken: string;
+  csrfToken?: string;
   editOrCreate?: string;
   identifier?: string;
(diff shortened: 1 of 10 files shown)


src/components/EditableConfigList.tsx
```diff
@@ -26,6 +26,7 @@ export interface EditableConfigListOwnProps {
   csrfToken: string;
   editOrCreate?: string;
   identifier?: string;
+  settingUp?: boolean;
```
aslagle: Is it possible to move any of the changes to this class to `IndividualAdminEditForm`, since that's the only one that's used while setting up? I'm fine with defining additional methods/constants in here that can be overridden by `IndividualAdminEditForm`, but I don't think the parent class needs to know anything about setting up.

src/components/SetupPage.tsx
```diff
@@ -6,6 +6,7 @@ import { State } from "../reducers/index";
 export interface SetupPageContext {
   editorStore: Store<State>;
   csrfToken: string;
+  settingUp: boolean;
```
aslagle: This is fine but you don't actually need it - If you're in the setup page at all you already know you're setting up.

src/components/EditableConfigList.tsx
adriana-alter: I agree in theory, but if I move these elements and functions out of the parent component and into the IndividualAdminEditForm, then I'll have to also copy/paste them into each of the other forms, and that seems like a much worse idea and deeply not worth it...

src/components/SetupPage.tsx
EdwinGuzman: I think what Amy means is you can do `<IndividualAdmins ... settingUp />` (or `settingUp={true}`) since that value will always be true in the `SetupPage`.

adriana-alter: Ohhhh.  Yes.  Good catch.  Changing that now.

src/components/EditableConfigList.tsx
aslagle: I don't think so - why would the other forms need those?

adriana-alter: I'm assuming we're talking about the h2 and h3, right?  All of the forms need those.  On everything except for the setup form, the h2 always says "${thing} configuration" and the h3 always says "Create a new ${thing}."  So the h2 and h3 live in the parent element rather than being individually defined in each form.  If I move them out of the parent element in order to move them to the setup form, then they'll disappear from all of the other forms.  Or did you mean something else?

EdwinGuzman: Can you define the parent class function to say `${thing} configuration` and the subclass functions to override it and print out a different message?

src/components/EditableConfigList.tsx
```diff
@@ -73,9 +74,19 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
   render(): JSX.Element {
     let EditForm = this.EditForm;
     let AdditionalContent = this.AdditionalContent || null;
+    let headers = this.getHeaders();
+
+    let className = "";
+    if (AdditionalContent) {
+      className = "has-additional-content";
+    }
+    else if (this.props.settingUp) {
```
EdwinGuzman: Just minor stylistic comment, but this line should go on line 82.

src/components/EditableConfigList.tsx
```diff
@@ -226,7 +243,13 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
 
   save(data: FormData) {
     this.editItem(data).then(() => {
-      if (this.limitOne && this.props.editOrCreate === "create") {
+      // If we're setting up an admin for the first time, refresh the page
+      // to go to login.
+      if (this.props.settingUp) {
```
EdwinGuzman: Why move this from the `IndividualAdminEditForm` to here?

EdwinGuzman:
This looks and works well for me. The `EditableConfigList` component can be updated with Amy's comments.

src/components/EditableConfigList.tsx
aslagle: ^ yes, that's what I meant.

src/components/EditableConfigList.tsx
adriana-alter: This makes it wait to see whether there's an error message before it refreshes the page.  (It turns out that it's not that the error message wasn't being displayed before--it's that the error message was being displayed but the user didn't have a chance to see it because the form was immediately reloading).

src/components/EditableConfigList.tsx
adriana-alter: Oh.  Yes.  Probably.  I'll see what I can do!

src/components/EditableConfigList.tsx
```diff
@@ -23,9 +23,10 @@ export interface EditableConfigListDispatchProps<T> {
 
 export interface EditableConfigListOwnProps {
   store?: Store<State>;
-  csrfToken: string;
+  csrfToken?: string;
```
aslagle: Did you mean to change this?

src/components/IndividualAdminEditForm.tsx
```diff
@@ -348,12 +350,6 @@ export default class IndividualAdminEditForm extends React.Component<IndividualA
   async submit(event) {
     event.preventDefault();
     await handleSubmit(this);
```
aslagle: Now that the reload isn't in here it might be possible to get rid of this function and just call `handleSubmit` from the save button.

src/components/IndividualAdmins.tsx
```diff
@@ -23,7 +23,29 @@ export class IndividualAdmins extends EditableConfigList<IndividualAdminsData, I
...
+  }
+
+  save(data: FormData) {
+    this.editItem(data).then(() => {
+      // If we're setting up an admin for the first time, refresh the page
+      // to go to login.
+      if (this.props.settingUp) {
+       window.location.reload();
+       return;
+     }
```
aslagle: If we're not setting up I think we still want to do the thing from the parent class where it redirects to the edit page? (Overall this last commit looks great though!)

adriana-alter: The parent class only does that for things where you're only allowed to create 1; otherwise, it just clears the form and displays a link to the edit page.  (Thanks!)

aslagle: Cool, so then would we want to clear the form and display the link here?

src/components/IndividualAdminEditForm.tsx
adriana-alter: The point of having it here is so that you can submit the form by hitting enter (in addition to by clicking the save button).  It doesn't have to be async anymore, though :)

src/components/IndividualAdmins.tsx
adriana-alter: Exactly!  It still does it.  The link gets displayed by `successMessage()` in EditableConfigForm and `clearForm` gets called by IndividualAdminEditForm; none of the changes from this branch affect those.

PR #193, opened 2018-11-23
Hide error message on return to list

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1473


src/components/EditableConfigList.tsx
```diff
@@ -73,6 +73,7 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
   render(): JSX.Element {
     let EditForm = this.EditForm;
     let AdditionalContent = this.AdditionalContent || null;
+    console.log(this.props.editOrCreate);
```
EdwinGuzman: Left a console.log here.

src/components/EditableConfigList.tsx
```diff
@@ -81,7 +82,7 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
             {this.successMessage()}
           </Alert>
         }
-        { this.props.fetchError &&
+        { this.props.fetchError && this.props.editOrCreate &&
```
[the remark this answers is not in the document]
adriana-alter: That's...a really good point.  The only way I can think of to get around this is to check the error code, on the assumption that an error from fetching the list will be a 404 and an error from submitting the form will be a 400.  Does that seem reasonable, or are there too many edge cases?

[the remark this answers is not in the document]
aslagle: I think it would be better to change it so the errors are stored in separate places rather than all error types ending up in `fetchError`.

src/components/Collections.tsx
```diff
@@ -70,10 +70,12 @@ function mapStateToProps(state, ownProps) {
   if (state.editor.collectionLibraryRegistrations && state.editor.collectionLibraryRegistrations.data) {
     data.libraryRegistrations = state.editor.collectionLibraryRegistrations.data.library_registrations;
   }
+  // fetchError = an error involving loading the list of collections; formError = an error upon
+  // submission of the create/edit form (including upon submitting a change to a library's registration).
   return {
     data: data,
     responseBody: state.editor.collections && state.editor.collections.successMessage,
-    fetchError: state.editor.collections.fetchError || (state.editor.collectionLibraryRegistrations && state.editor.collectionLibraryRegistrations.fetchError) || (state.editor.registerLibraryWithCollection && state.editor.registerLibraryWithCollection.fetchError),
+    formError: state.editor.collections.formError || (state.editor.collectionLibraryRegistrations && state.editor.collectionLibraryRegistrations.fetchError) || (state.editor.registerLibraryWithCollection && state.editor.registerLibraryWithCollection.fetchError),
```
aslagle: This one doesn't look right....

adriana-alter: Yes--accidentally deleted a line.  Thank you!

src/reducers/createFetchEditReducer.ts
```diff
@@ -86,7 +86,7 @@ export default<T> (fetchPrefix: string, editPrefix?: string, extraActions?: Extr
             case `${editPrefix}_${ActionCreator.FAILURE}`:
               return Object.assign({}, state, {
                 isEditing: false,
-                fetchError: action.error
+                formError: action.error
```
aslagle: The `formError` should also be cleared in some of the other actions or it will never go away.

PR #194, opened 2018-11-23
Active header links

EdwinGuzman:
Resolves [SIMPLY-592](https://jira.nypl.org/browse/SIMPLY-592).

This is a to simple add an active class to the header nav links:
screen shot 2018-11-23 at 2 45 35 pm

Unfortunately this is not as easy as adding `activeClassName` to the `<Link>` component from react-router. It works at the top level but we have a lot of subpages that are used. For example, clicking on "Patrons" (`admin/web/patrons/library`) works fine, but going to "Reset Adobe ID" (`admin/web/patrons/dune/resetAdobeId`) doesn't work since the context seems to be lost then. Instead of using `activeClassName` I just read the current location each time and mapped it. This doesn't work for the first three nav links, however, and I think it's because Catalog, Complaints, and Hidden Books are internal routes in the `opds-web-catalog` component. So I left it to use `activeClassName` for now. Still, if you go to the catalog and then select a sublane or a book, the context is lost. Again, I think it's because it's not making a new "page request" unlike the other admin links since those pages are all considered to be one route. Help appreciated for this.


aslagle:
A designer should weigh in on this, but for the catalog, I think it's fine if the header link is not highlighted for subpages - you can click the header link again to get to the top level. For the patron manager page I'm not sure, since it uses tabs.

EdwinGuzman:
It's the same case for the List page (as the patron page) since the create and edit pages are subpages of List.

adriana-alter:
Why is the header rendered as a subcomponent of the other pages, anyway?  To my knowledge, there's no situation in which we wouldn't want the header to be rendered (assuming the user is logged in, obviously), so why not move it up to the same level as the catalog page, etc.?  I think that would not only fix this, but make more sense overall...unless there was a particular reason for doing it this way?

aslagle:
I think it's mostly because the catalog part uses opds-web-client, which normally renders a full page but lets you swap in a different header.

adriana-alter:
Ah, that makes sense.  Any way around it?

aslagle:
It seems like it would be a lot of work to change that - I don't have any ideas right now.

jackkim9:
@EdwinGuzman visually can we try something like this? https://cl.ly/322fc5445760. 

Default/On hover:
- no underline (user will know they are clickable even without them. e.g. amazon.com)

On hover:
- background-color: #3a7dfd; (I don't think this blue is $-defined anywhere. Maybe this could be $blue.)
- border-radius: 2px;
- transition: background-color 0.3s ease-in;

EdwinGuzman:
Updated. Screenshots:
screen shot 2018-11-27 at 4 46 39 pm
screen shot 2018-11-27 at 4 46 49 pm

jackkim9:
LGTM

EdwinGuzman:
@aslagle  @adriana-alter 👀

src/components/Header.tsx
```diff
@@ -55,8 +59,16 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
   }
 
   render(): JSX.Element {
-
     let isLibraryManager = this.context.library && this.context.admin.isLibraryManager(this.context.library());
+    let links = ["maincatalog", "complaints", "suppressed", "lists", "lanes", "dashboard", "patrons", "config", "account"];
+    let linkIndex = 0;
+    links.forEach((link, index) => {
```
adriana-alter: I like the way this works, but it's a little bit confusing to read; how about adding a comment and/or saving some of the stuff in that conditional as a variable (e.g. `let routerHasLink = this.context.router.getCurrentLocation().pathname.indexOf(link) !== -1)`)?

Everything else looks good to me.

EdwinGuzman: Sure, updated.

src/components/Header.tsx
```diff
@@ -30,10 +30,14 @@ export interface HeaderState {
   showAccountDropdown: boolean;
 }
 
+export interface HeaderRouter extends Router {
+  getCurrentLocation?: Function;
```
aslagle: I'm curious if it would work to use the react router type instead of defining your own interface.

src/components/Header.tsx
```diff
@@ -55,8 +59,16 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
   }
 
   render(): JSX.Element {
-
-    let isLibraryManager = this.context.library && this.context.admin.isLibraryManager(this.context.library());
+    const getCurrentLocation = this.context.router && this.context.router.getCurrentLocation();
+    let isLibraryManager =
+      this.context.library && this.context.admin.isLibraryManager(this.context.library());
+    let links = ["maincatalog", "complaints", "suppressed", "lists", "lanes", "dashboard", "patrons", "config", "account"];
```
aslagle: This seems like it will be very annoying to maintain. I would rather compare the current location against these strings in each of the individual links instead of getting the index, even if it's slightly less efficient.

aslagle: Also, does the "maincatalog" one just happen to work because the index starts at 0?

src/stylesheets/header.scss
```diff
@@ -1,62 +1,90 @@
...
+    color: $pagecolorlight;
+    font-size: 1.8em;
 
-.navbar-brand > img.nypl-logo {
-  display: inline;
-  height: 25px;
-  margin-right: 10px;
-  margin-top: -5px;
-}
+    > img.nypl-logo {
```
aslagle: I know you didn't change this but why is it here? There's no logo in this UI.

aslagle:
I have a few design comments:

1. The contrast isn't good enough. It passes AA for large text but this text is only 16pt.
2. I like how it looks for hovering but for the active link I feel like it's too prominent. In some cases like the dashboard it's the most noticeable thing on the page.
3. When you hover over a link next to the active link, shouldn't there be some space between the blue?

adriana-alter:
4. I would prefer to have the hover background-color be different from the active background-color.  (Which would also resolve #3.)

src/stylesheets/header.scss
adriana-alter: Not in this branch, but: maybe there *should* be a logo in this UI anyway.

src/components/Header.tsx
adriana-alter: `isActive` and `location` seem useful (and potentially simpler than the current implementation): https://reacttraining.com/react-router/web/api/NavLink/isactive-func

src/components/Header.tsx
adriana-alter: Again, using the React router might simplify this; looks like `match` and/or `matchPath` would do it. https://reacttraining.com/react-router/web/api/match

EdwinGuzman:
Those are good comments for @jackkim9 to work on.

src/components/Header.tsx
EdwinGuzman: Those might work. Just FYI, those are the react router v4 docs and we are on v3. V3 doesn't include `<NavLink>` which I previously did try to use. An upgrade would be great but that's not for this ticket.

EdwinGuzman: @aslagle the Typescript linter complains that `getCurrentLocation` does not exist on type Router. That's why I extended it... even though that function should be a part of the route instance?

aslagle: `Router` here is the interface from opds-web-client, not react router, so I get why that doesn't work. But I was curious if you could use the router type from react router instead of the one from opds-web-client here.

EdwinGuzman: The react-router `Router` type doesn't seem to work. Typescript can't find `isActive` or `getCurrentLocation` from the `Router` type.

@adriana-alter the `isActive` function seems to be only for v4. There' is a `isActive` defined function in the Router interface for the `opds-web-client` but it doesn't seem to be used or defined anywhere (in this repo and the child repo).

adriana-alter: Ah.  Never mind, then :)

jackkim9:
1: Good point. Let's address this in all active buttons, as this applies to all of them. (Hopefully in a separate bug ticket.)
2/4: Agreed.
3: This doesn't seem entirely necessary as the border is rounded and the user can probably tell two buttons apart that are side by side.

EdwinGuzman:
Screenshot of Jack's lately commit:

screen shot 2018-12-03 at 2 59 11 pm

src/stylesheets/colors.scss
```diff
@@ -10,3 +10,4 @@ $light-gray: #D7D4D0;
 $medium-gray: #DDD;
 $dark-gray: #080807;
 $white: #FFF;
+$blue: #3A7DFD;
```
jackkim9: @EdwinGuzman Since we added this in, can we match this as "$nypl-blue" in our .org site? (#1B7FA7). We can then use this to update active buttons.

adriana-alter:
The contrast between the dark gray and the medium gray doesn't look quite high enough to me...

jackkim9:
Contrast should be good. It's light-gray on dark-gray. (Previously dark-gray on light-gray by mistake) 
screen shot 2018-12-03 at 3 08 10 pm

screen shot 2018-12-03 at 3 10 32 pm

src/stylesheets/colors.scss
EdwinGuzman: Since this is not just used at NYPL, it'd be better to not have `nypl` in variable names.

jackkim9: Correct. That's what I meant. 
$blue: #1B7FA7;

EdwinGuzman:
Updated the logic for getting current active location and here's an updated screenshot from Jack's latests commit:

screen shot 2018-12-03 at 3 14 39 pm

src/components/Header.tsx
```diff
@@ -88,51 +96,74 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
...
+                  activeClassName="active-link"
+                >
                   Hidden Books
                 </CatalogLink>
               </li>
               <li className="header-link">
-                <Link to={"/admin/web/lists/" + this.context.library()}>Lists</Link>
+                <Link
+                  to={"/admin/web/lists/" + this.context.library()}
+                  className={currentPathname.indexOf("lists") !== -1 ? "active-link" : ""}
```
aslagle: I'd rather have these check for longer paths like "/admin/web/lists", just in case we end up with another URL with "lists" in it.

EdwinGuzman: Updated.

PR #195, opened 2018-11-23
Displaying the name of each Patron Authentication protocol configured.

EdwinGuzman:
Resolves [SIMPLY-580](https://jira.nypl.org/browse/SIMPLY-580).
Simply displaying the name of each Patron Authentication protocol before what protocol it is.

screen shot 2018-11-23 at 3 46 33 pm


adriana-alter:
Very small stylistic suggestion: put a colon instead of a dash after the name.  Otherwise, looks good!

PR #197, opened 2018-11-27
Lane empty list

EdwinGuzman:
Adding this PR as part of v0.0.90.
Resolves [SIMPLY-1481](https://jira.nypl.org/browse/SIMPLY-1481).

Just needed to update the mapping of the props and Redux store values in the Lane component to 1) correctly show form error messages and 2) correctly redirect after successfully creating a new Lane. Also did some minor code refactoring to Lanes.tsx.

### DIFF
--- a/src/components/Lanes.tsx
+++ b/src/components/Lanes.tsx
@@ -22,7 +22,7 @@ export interface LanesStateProps {
   lanes: LaneData[];
   customLists: CustomListData[];
   responseBody?: string;
-  fetchError?: FetchErrorData;
+  formError?: FetchErrorData;
   isFetching: boolean;
 }
 
@@ -62,6 +62,7 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
     this.hideLane = this.hideLane.bind(this);
     this.showLane = this.showLane.bind(this);
     this.resetLanes = this.resetLanes.bind(this);
+    this.toggle = this.toggle.bind(this);
 
     const expanded: { [key: string]: boolean } = {};
     for (const topLevelLane of this.props.lanes || []) {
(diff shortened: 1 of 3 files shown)


src/components/Lanes.tsx
```diff
@@ -150,16 +151,19 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
     return (
       <ul>
         { lanes.map(lane =>
-            <li className={(String(lane.id) === this.props.identifier) ? "active" : ""}>
+            <li
+              key={lane.display_name.replace(" ", "").trim()}
```
aslagle: Are you not supposed to have spaces in keys? I didn't know that.

src/components/Lanes.tsx
```diff
@@ -329,9 +327,9 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
 function mapStateToProps(state, ownProps) {
   return {
     lanes: state.editor.lanes && state.editor.lanes.data && state.editor.lanes.data.lanes,
-    responseBody: state.editor.lanes && state.editor.lanes.responseBody,
+    responseBody: state.editor.lanes && state.editor.lanes.successMessage,
     customLists: state.editor.customLists && state.editor.customLists.data && state.editor.customLists.data.custom_lists,
-    fetchError: state.editor.lanes.fetchError || state.editor.laneVisibility.fetchError,
```
aslagle: Don't you still need `fetchError`, for if there's an error from the request to get the list of lanes?

src/components/Lanes.tsx
aslagle: As long as I don't create two lanes named "Lane A" and "LaneA" :)

PR #198, opened 2018-11-30
Added a system configuration tab for external catalogs.

aslagle:
I'm still working on tests for the corresponding circulation branch, but this branch adds a new service type to system configuration. For now, the only protocol is "MARC Export", which will set up the circ manager to generate MARC files that can be imported into an external catalog.

### DIFF
--- a/src/components/ConfigTabContainer.tsx
+++ b/src/components/ConfigTabContainer.tsx
@@ -10,6 +10,7 @@ import AnalyticsServices from "./AnalyticsServices";
 import CDNServices from "./CDNServices";
 import SearchServices from "./SearchServices";
 import StorageServices from "./StorageServices";
+import CatalogServices from "./CatalogServices";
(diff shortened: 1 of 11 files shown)


src/components/ConfigTabContainer.tsx
```diff
@@ -165,6 +174,8 @@ export default class ConfigTabContainer extends TabContainer<ConfigTabContainerP
       return "Sitewide Settings";
     } else if (name === "cdn") {
       return "CDN";
+    } else if (name === "catalogServices") {
```
adriana-alter: This might not be the branch to do this in, but: now that we have so many tabs, how about storing and accessing the display names via an object with name/tabDisplayName as key/value pairs (i.e. having this function be something along the lines of `return displayNameConstants[name]`), rather than having everything as a separate if/else statement?

adriana-alter:
Looks good to me!

PR #199, opened 2018-12-04
Subdivide optional library fields into smaller categories

adriana-alter:
I'm not sure whether these are the best categories to divide the fields into; let me know if you think there's something else that would make more sense!

screen shot 2018-12-04 at 12 33 20 pm


src/components/LibraryEditForm.tsx
```diff
@@ -26,13 +26,57 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
     this.submit = this.submit.bind(this);
   }
 
+  renderForms(nonRequiredFields) {
+    const OPTIONAL_FIELDS = {
+      "Loans, Holds, & Searches": ["allow_holds", "enabled_entry_points", "featured_lane_size", "minimum_featured_quality", "facets_enabled_available", "facets_enabled_order", "facets_enabled_collection", "facets_default_available", "facets_default_order", "facets_default_collection", "loan_limit", "hold_limit", "max_outstanding_fines"],
+      "Patron Support": ["help-web", "help-uri", "copyright_designated_agent_email_address", "configuration_contact_email_address", "register"],
+      "Interface Customization": ["color_scheme", "web-background-color", "web-foreground-color", "web-header-links", "web-header-labels", "logo"],
+      "Collection Languages": ["large_collections", "small_collections", "tiny_collections"],
+      "Additional Information": ["library_description", "focus_area", "service_area", "terms-of-service", "privacy-policy", "copyright", "about", "license"]
```
aslagle: I think this should be defined on the server and sent down as part of or alongside the fields that are already coming from there. It should be possible to add a field on the server without needing to modify anything here.

src/components/LibraryEditForm.tsx
```diff
@@ -26,13 +26,57 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
     this.submit = this.submit.bind(this);
   }
 
+  renderForms(nonRequiredFields) {
+    const OPTIONAL_FIELDS = {
+      "Loans, Holds, & Searches": ["allow_holds", "enabled_entry_points", "featured_lane_size", "minimum_featured_quality", "facets_enabled_available", "facets_enabled_order", "facets_enabled_collection", "facets_default_available", "facets_default_order", "facets_default_collection", "loan_limit", "hold_limit", "max_outstanding_fines"],
```
aslagle: I would separate allow holds, loan limit, hold limit, and fines into their own category.
"searches" is probably not a good name for the other one - it applies to many lanes, not just search.

src/components/LibraryEditForm.tsx
```diff
@@ -26,13 +26,57 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
...
+      "Patron Support": ["help-web", "help-uri", "copyright_designated_agent_email_address", "configuration_contact_email_address", "register"],
+      "Interface Customization": ["color_scheme", "web-background-color", "web-foreground-color", "web-header-links", "web-header-labels", "logo"],
```
aslagle: I'd suggest "_Client_ Interface Customization"

src/components/LibraryEditForm.tsx
aslagle: I would make links a separate category.

EdwinGuzman: I agree about defining those in the server. Would make maintaining this in the front-end easier.

src/components/LibraryEditForm.tsx
```diff
@@ -26,13 +26,57 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
...
+      "Collection Languages": ["large_collections", "small_collections", "tiny_collections"],
+      "Additional Information": ["library_description", "focus_area", "service_area", "terms-of-service", "privacy-policy", "copyright", "about", "license"]
+    };
+    let forms = [];
+    let categories = Object.keys(OPTIONAL_FIELDS);
+    categories.map((category) => {
```
EdwinGuzman: `.map` returns a new array so in this case `.forEach` would be a better function to use.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.0.91
```
EdwinGuzman: This should now be 0.0.93.

src/components/__tests__/LibraryEditForm-test.tsx
```diff
@@ -22,8 +23,8 @@ describe("LibraryEditForm", () => {
     }
   };
   let settingFields = [
-    { key: "privacy-policy", label: "Privacy Policy" },
-    { key: "copyright", label: "Copyright" }
+    { key: "privacy-policy", label: "Privacy Policy", category: "Links" },
+    { key: "copyright", label: "Copyright", category: "Links" }
```
aslagle: It would be good to have more than one category in the test.

EdwinGuzman:
Didn't get another chance to look at it before merging but looks goos.

PR #202, opened 2018-12-10
Replicate type=number functionality

adriana-alter:
A solution to the `type=number` issue: I wrote a method that prevents the user from entering invalid characters, but submits the input to the server for validation no matter what.

### DIFF
--- a/src/components/EditableInput.tsx
+++ b/src/components/EditableInput.tsx
@@ -9,6 +9,7 @@ export interface EditableInputProps extends React.HTMLProps<EditableInput> {
   required?: boolean;
   error?: FetchErrorData;
   optionalText?: boolean;
+  validation?: string;
 }
 
 export interface EditableInputState {
@@ -73,6 +74,7 @@ export default class EditableInput extends React.Component<EditableInputProps, E
       ref: "element",
       value: this.state.value,
       checked: this.state.checked,
+      onKeyPress: this.props.validation && this.props.validation === "number" && this.validateNumber,
       onChange: this.handleChange,
       style: this.props.style,
       placeholder: this.props.placeholder,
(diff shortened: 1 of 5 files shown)


src/components/ProtocolFormField.tsx
```diff
@@ -78,7 +78,7 @@ export default class ProtocolFormField extends React.Component<ProtocolFormField
       return (
         <EditableInput
           elementType="input"
-          type="number"
```
aslagle: With the added validation, is it possible to keep `type="number"` too so that we get the nicer keyboard on mobile? It seems like this would solve the problems with it.

src/components/EditableInput.tsx
```diff
@@ -117,6 +119,13 @@ export default class EditableInput extends React.Component<EditableInputProps, E
     }
   }
 
+  validateNumber(e) {
+    const validChars = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", ".", "Enter"];
```
aslagle: Do you also need to allow tab for getting out of the field, or does that already work differently?

src/components/ProtocolFormField.tsx
adriana-alter: As far as I can tell, it's either/or, unfortunately; `type="number"` will block this from running.  Are admins really filling out these forms on mobile devices anyway, though?

src/components/EditableInput.tsx
adriana-alter: I just tried it out--it already works differently (but good idea!!!)

src/components/ProtocolFormField.tsx
aslagle: No, but I just thought it would be nice to support it if we could. If it doesn't work it's fine like this.

adriana-alter: It would be nice.  But no--I don't think there's a way to make it work.  (Maybe in the next version of HTML?  It seems like a lot of people are unhappy with how `type="number"` is currently working...)

src/components/EditableInput.tsx
EdwinGuzman: I wonder why that works differently but it is important to get out of the input for accessibility.

PR #204, opened 2018-12-10
Classification status

EdwinGuzman:
Fixes [SIMPLY-1521](https://jira.nypl.org/browse/SIMPLY-1521). It use to be unclear if a book did _not_ have an audience or fiction classification. The UI would display the first values in each input field but they weren't actually true unless if an admin saved the form.

Now, if an admin tries to save without selecting any values for a book that has no audience or fiction classifications, then the admin will complain. There's will also be a "None" radio button option which will go away once a value is selected.

screen shot 2018-12-10 at 5 27 35 pm

Also wrapped in the inputs in a fieldset and into non-collapsible Collapsible components.
screen shot 2018-12-10 at 5 18 29 pm

This is a bit hard to test without any real book that has no audience or fiction values but I think the tests are thorough.


aslagle:
I'm not sure if genres should be grouped with fiction status - they're related but they are separate things.

src/components/ClassificationsForm.tsx
```diff
@@ -33,102 +36,143 @@ export default class ClassificationsForm extends React.Component<Classifications
...
-                value={this.props.book.targetAgeRange[1]}
-                />
+                name="audience"
+                label="Audience"
+                ref="audience"
+                value={this.props.book.audience || "None"}
+                onChange={this.handleAudienceChange}
+                clientError={this.state.error && this.state.error["audience"]}
+              >
+                <option value="None">---------</option>
```
aslagle: Looks like it would be easy to only show this option if the audience is currently not set. Since we aren't letting you save with this value I think we should not let you change an existing audience to it.

adriana-alter: Just what I was about to say!  Also, I think it would make more sense to have the not-set value say "None" (rather than "----"), since the not-set radio button says "None."

adriana-alter:
I agree that genres shouldn't be grouped with fiction status.  Genres are optional...right?  How about putting them in their own collapsible (which--if I'm right about them being optional!--can start out collapsed)?

adriana-alter:
Probably not for this branch, or even this ticket, but: what is the target audience for the "related classifications" information at the bottom of the page?  I imagine this information won't mean very much to most admins; even if they know what LCSH is, are they going to know what we mean by "weight"?

I suggest that we provide some explanation for what we're talking about; how about--drumroll, please!--an informational collapsible like we have for the Google integrations?  And/or display the related classifications list next to, rather than below, the genres form; if you can actually see the classifications list updating as you add genres (rather than having to scroll all the way down and notice--or not--that there are suddenly new list items), that might help make things clearer.

src/components/ClassificationsForm.tsx
```diff
@@ -256,14 +300,32 @@ export default class ClassificationsForm extends React.Component<Classifications
...
+      fiction,
+    } = this.state;
     let data = new (window as any).FormData();
-    data.append("audience", this.state.audience);
+    let error = {};
+    if (!audience || audience === "None") {
+      error["audience"] = "No Audience classification selected.";
+    }
+    if (fiction === undefined) {
+      error["fiction"] = "No Fiction classification selected.";
```
adriana-alter: How about defining these errors on the server instead (the same way we check the required fields for other forms)?

EdwinGuzman:
Sure re: Related Classifications, but different issue.

I grouped genre together because selecting fiction/nonfiction changes the options in the "Add Genre" area. If it was collapsed you wouldn't know that they changed and would be confusing that you selected fiction and see some genres, but then select nonfiction and see other genres.

I could separate it but I do think it's bothersome to not see page changes happen when you select different values somewhere else on the page. Thoughts?

src/components/ClassificationsForm.tsx
EdwinGuzman: I was thinking about that but just made these client side only errors so you can't even submit if something on the page is wrong. Different pattern to try out?

adriana-alter:
Ah, that makes sense.  How about separating them but keeping both of them un-collapsed?

src/components/ClassificationsForm.tsx
adriana-alter: Hmm.  It kind of bothers me on principle to not have all of the form validations use the same pattern...but on the other hand, the way you're doing this makes a lot of sense, and it definitely doesn't seem worth changing all of the other form validations to match this one (unless there were other major advantages to changing them).  Is there any particular reason we originally went with the current pattern of validating everything server-side?  If not, then I think you just leave this as it is, and we table this conversation for the next time we're implementing a new form validation...

src/components/ClassificationsForm.tsx
EdwinGuzman: Updated.

EdwinGuzman:
Updated to separate the genres into a separate field.

src/components/ClassificationsForm.tsx
```diff
@@ -33,102 +36,139 @@ export default class ClassificationsForm extends React.Component<Classifications
...
-          <option value="Adult">Adult</option>
-          <option value="Adults Only">Adults Only</option>
-        </EditableInput>
-
-        { this.shouldShowTargetAge() &&
-          <div className="form-group target-age">
-            <label>Target Age Range</label>
-            <div className="form-inline">
+      <fieldset className="classifications-form">
+        <legend className="visuallyHidden">Classifications and Genres</legend>
```
aslagle: I would consider genres to be a type of classification.

adriana-alter:
At such time as we reconsider the related classifications list (agreed, not this branch!), "NONE" doesn't seem like something that should be displayed...
screen shot 2018-12-11 at 12 29 24 pm

src/components/ClassificationsForm.tsx
EdwinGuzman: I'll update this to be just `Classifications` but the titles for each collapsible as categories and genres sounds good to me.

aslagle: Not sure why genres wouldn't be categories as well though

aslagle:
The NONE is important because it indicates that library staff explicitly set this book to not have a genre, overruling any genre classifications we got from other sources.

src/components/ClassificationsForm.tsx
EdwinGuzman: How about everything under one collapsible field and label it as "categories"?

adriana-alter:
Oh, ok; never mind!

src/components/ClassificationsForm.tsx
aslagle: might as well stick with classifications if it's one collapsible.

adriana-alter: They kind of would be, admittedly.  But I don't have a better idea for what to title the first collapsible...

adriana-alter: Three collapsibles: "audience," "fiction status," "genres"?  Or is that overkill?

EdwinGuzman: I think it may be overkill. This would seem better to just leave as one labeled classifications and later on get back to if we need to.

src/components/ClassificationsForm.tsx
EdwinGuzman: I was further thinking about this and trying to see if I could make it return server validation errors instead. Here we are specifically sending "fiction" or "nonfiction" for the fiction form value. This may involve some changes for this component to use server validation but I do think it'll be worth it to be consistent. @aslagle will be updating this.

aslagle: I'm not sure what the server does not but I think it would be reasonable to return an error if it's anything other than "fiction" or "nonfiction". That doesn't necessarily mean you need to remove the client-side code though.

EdwinGuzman: Right, if I don't remove the code I added the form won't submit and instead of getting server side errors we'd just see the client-side errors. I'm okay with this and it's how I built this, but considering the server side option for consistency and for the validation pattern that Adriana pointed out.

EdwinGuzman:
Update: I'm leaving the form as one collapsible field and the client-side validations in place. I think it'd be better to revisit this later on if we want to switch to server-side validation and different groupings.

src/components/ClassificationsForm.tsx
```diff
@@ -256,14 +290,32 @@ export default class ClassificationsForm extends React.Component<Classifications
...
-    data.append("audience", this.state.audience);
+    let error = {};
+    if (!audience || audience === "None") {
+      error["audience"] = "No Audience classification selected.";
+    }
+    if (fiction === undefined) {
+      error["fiction"] = "No Fiction classification selected.";
+    }
+    if (error["audience"] || error["fiction"]) {
+      window.scrollTo(0, 0);
```
aslagle: This scrolling isn't working for me. 

I also wonder if it might be good to highlight the none even before you save, so it's obvious it needs fixed when you look at it.

EdwinGuzman: Instead of scrolling to the top, the focus will go to the error message area. This seems to always work. I realized this is better for accessibility anyways. I think the opds web client adds an "overflow: hidden" to the body when this specific page gets mounted and I think that messes up scrolling to the top programatically.

Also highlighted the None values initially when the book doesn't have those values set.

src/components/ClassificationsForm.tsx
```diff
@@ -18,9 +20,10 @@ export default class ClassificationsForm extends React.Component<Classifications
   constructor(props) {
     super(props);
     this.state = {
-      audience: props.book ? props.book.audience : null,
-      fiction: props.book ? props.book.fiction : null,
-      genres: []
+      audience: props.book && props.book.audience ? props.book.audience : "None",
```
aslagle: This isn't working because if the audience in the props is undefined componentWillMount changes it back from "None" to undefined.

EdwinGuzman: I see. But in the submit check it's checking to see if audience is undefined or "None" to throw the error. I will update the `componentWillMount` function to not change it back to undefined.

aslagle: Yep, that fixed it.

PR #205, opened 2018-12-12
Updating readme docs.

EdwinGuzman:
Updated some docs to get a new version on npm to try out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 #### Updated
 - Documentation for the repo.
 
-### v0.0.94
+### v0.0.94 - Do not use
 #### Deprecated
 - There's been an issue found in this version of the npm package and we recommend not to use this version.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Documentation for the repo.
 
 ### v0.0.94
+#### Deprecated
+- There's been an issue found in this version of the npm package and we recommend not to use this version.
+
 #### Updated
 - Implemented a custom number validator for better error handling.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 #### Updated
 - Documentation for the repo.
 
-### v0.0.94 - Do not use
-#### Deprecated
+### v0.0.94
+#### Note: this version contains an npm bug. Please upgrade to v0.0.95.
 - There's been an issue found in this version of the npm package and we recommend not to use this version.
 
 #### Updated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.0.95
+#### Updated
+- Documentation for the repo.
+
 ### v0.0.94
 #### Updated
 - Implemented a custom number validator for better error handling.
--- a/README.md
+++ b/README.md
@@ -1,13 +1,37 @@
 # circulation-web
-Web front-end for the Circulation Manager administrative interface
+Web front-end for the Circulation Manager administrative interface.
+
+[![npm version](https://badge.fury.io/js/simplified-circulation-web.svg)](https://badge.fury.io/js/simplified-circulation-web)
+
+[![Build Status](https://travis-ci.org/NYPL-Simplified/circulation-web.svg?branch=master)](https://travis-ci.org/NYPL-Simplified/circulation-web)
 
 ## Setup
 
 This package is meant to be used with the Library Simplified [Circulation Manager](https://github.com/NYPL-Simplified/circulation).
 
-To use the published version with your circulation manager, run `npm install` from `api/admin`.
+Suggested local setup:
+1. `/[path to project folder]/circulation`
+2. `/[path to project folder]/circulation-web`
 
-If you're working on the admin interface and want to test local changes, you can link your local clone of this repository to your local circulation manager. Run `npm link` in this repository, then run `npm link simplified-circulation-web` from `api/admin` in the circulation manager.
+#### Use npm version
+
+To use the published version with your circulation manager, run `npm install` from `api/admin` in the `circulation` local installed repository (1. above).
+
+#### Use local development version
+
+If you're working on the admin interface and want to test local changes, you can link your local clone of this repository to your local circulation manager. Run `npm link` in this repository (2. above), then run `npm link simplified-circulation-web` from `api/admin` in the circulation manager (1. above).
+
+These steps will allow you to work on the front-end admin interface and see updates while developing.
+
+In order to make changes:
+1. link the local clone repository to local circulation manager,
+2. run the `circulation` manager using `python app.py`,
+3. run the `circulation-web` interface using `npm run dev`
+4. visit `localhost:6500/admin/`
+
+## Web Catalog
+
+The Circulation Manager admin interface relies on the [OPDS Web Catalog](https://github.com/NYPL-Simplified/opds-web-client) as its base React component and application. For more information, please check out the repository.
 
 ## Publishing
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1,20 +1,51 @@
...
 
 ## Publishing
 
 This package is [published to npm](https://www.npmjs.com/package/simplified-circulation-web).
 
 To publish a new version, you need to create an npm account and be a collaborator on the package. Then you can run `npm publish` from your local copy of the repository.
 
+## Tests
+
+Like the codebase, all the unit tests are written in Typescript. Tests are written for all React components as well as redux and utility functions, and all can be found in their respective `__tests__` folders.
```
adriana-alter: Is it worth mentioning that they're in enzyme, or is that not relevant to the people who would be reading this?

EdwinGuzman: It is useful. But I wanted to stay away from enzyme, redux, react, and other tools we use for this pass. That's a bit more involved and could use better and more in-depth docs.

README.md
```diff
@@ -1,20 +1,51 @@
...
+[![npm version](https://badge.fury.io/js/simplified-circulation-web.svg)](https://badge.fury.io/js/simplified-circulation-web)
+
+[![Build Status](https://travis-ci.org/NYPL-Simplified/circulation-web.svg?branch=master)](https://travis-ci.org/NYPL-Simplified/circulation-web)
 
 ## Setup
 
 This package is meant to be used with the Library Simplified [Circulation Manager](https://github.com/NYPL-Simplified/circulation).
 
-To use the published version with your circulation manager, run `npm install` from `api/admin`.
+Suggested local setup:
```
adriana-alter: This might just be because I haven't had coffee yet, but I'm finding it confusing to list the steps here and then explain later on which ones to use; you're listing them here as 1 and 2, but then for npm you only use 1, and for local you use them in the opposite order.  It would be clearer to just move these lines into the individual npm and local paragraphs below.  (Yes, that would involve copy/pasting step 1 into two different places, but it's worth it!)

README.md
```diff
@@ -1,20 +1,51 @@
 # circulation-web
-Web front-end for the Circulation Manager administrative interface
+Web front-end for the Circulation Manager administrative interface.
+
+[![npm version](https://badge.fury.io/js/simplified-circulation-web.svg)](https://badge.fury.io/js/simplified-circulation-web)
+
+[![Build Status](https://travis-ci.org/NYPL-Simplified/circulation-web.svg?branch=master)](https://travis-ci.org/NYPL-Simplified/circulation-web)
```
adriana-alter: While we're listing links: how about linking to the circ manager (production) site?

README.md
```diff
@@ -1,20 +1,51 @@
...
 
 To publish a new version, you need to create an npm account and be a collaborator on the package. Then you can run `npm publish` from your local copy of the repository.
 
+## Tests
+
+Like the codebase, all the unit tests are written in Typescript. Tests are written for all React components as well as redux and utility functions, and all can be found in their respective `__tests__` folders.
+
+To run the tests, perform `npm test`.
+
+We use Travis CI for continuous integration and require pull requests tests for a new feature or bug fix to pass on Travis CI.
```
adriana-alter: Missing word--"require pull requests AND tests"?

README.md
EdwinGuzman: That's just for NYPL admins. There is no one open location for anyone to see since many circulation manager sites can be set up.

adriana-alter: Any point in having a demo account, or putting screenshots of the interface in these docs...or is that not worth it?

README.md
EdwinGuzman: Sure.

README.md
aslagle: Maybe link to confluence since that has lots of screenshots and how to guides already.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.0.95
+#### Updated
+- Documentation for the repo.
+
 ### v0.0.94
```
aslagle: Should probably add that 0.0.94 should not be used.

EdwinGuzman:
Updated.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 - Documentation for the repo.
 
 ### v0.0.94
+#### Deprecated
+- There's been an issue found in this version of the npm package and we recommend not to use this version.
```
adriana-alter: Can we word this more strongly?  0.0.94 appears to be BROKEN; this isn't just a matter of recommending that people not use it.  Definitely do not use it.

README.md
```diff
@@ -45,7 +52,7 @@ Like the codebase, all the unit tests are written in Typescript. Tests are writt
 
 To run the tests, perform `npm test`.
 
-We use Travis CI for continuous integration and require pull requests tests for a new feature or bug fix to pass on Travis CI.
+We use Travis CI for continuous integration and require pull requests and tests for a new feature or bug fix to pass on Travis CI.
```
adriana-alter: Again, maybe it's because no coffee, but I'm finding the structure of this sentence very confusing.  It reads like, we require pull requests in order for the Travis tests to pass...but you meant, before we deploy a new feature or bug fix, we require both that 1) there are pull requests and 2) the Travis tests pass...yes?

If I'm right about that, then how about:  "We use Travis CI for continuous integration.  Before your new feature or bug fix is deployed, it must pass the Travis CI tests.  In addition, you must submit a pull request."

CHANGELOG.md
EdwinGuzman: "appears to be broken" makes it seems like we haven't found out why... It's also deprecated and should be enough?

README.md
```diff
@@ -5,33 +5,40 @@ Web front-end for the Circulation Manager administrative interface.
...
-
-To use the published version with your circulation manager, run `npm install` from `api/admin` in the `circulation` local installed repository (1. above).
+To use the published version with your circulation manager, run `npm install` from `api/admin` in the `circulation` local installed repository.
 
 #### Use local development version
 
-If you're working on the admin interface and want to test local changes, you can link your local clone of this repository to your local circulation manager. Run `npm link` in this repository (2. above), then run `npm link simplified-circulation-web` from `api/admin` in the circulation manager (1. above).
+Suggested local setup:
+1. `/[path to project folder]/circulation`
+2. `/[path to project folder]/circulation-web`
```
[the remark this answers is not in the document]
EdwinGuzman: Hmm these don't need to be numbered, that's true. But that's just the folder structure and doesn't matter for the later steps

README.md
EdwinGuzman: You can't have a new feature without a pull request, from any project. So it seemed implied to me you have to submit a pull request.

CHANGELOG.md
adriana-alter: How about doing what we did for 0.0.81 (for the sake of consistency, if nothing else)?  We said `Note: this version contains a bug.  Please upgrade directly to v0.0.82.`  We could substitute "appears to be broken" for "contains a bug."

README.md
adriana-alter: Fair.  But then take the part about pull requests out of the sentence; it's just making it confusing, so if it doesn't have to be explicitly stated anyway...

README.md
```diff
@@ -52,7 +52,7 @@ Like the codebase, all the unit tests are written in Typescript. Tests are writt
 
 To run the tests, perform `npm test`.
 
-We use Travis CI for continuous integration and require pull requests and tests for a new feature or bug fix to pass on Travis CI.
+We use Travis CI for continuous integration. Any pull requests submitted must have tests and those tests must pass on Travis CI.
 
 ## License
```
adriana-alter: I like it :)

PR #206, opened 2018-12-12
Allow dollar sign in number field

adriana-alter:
While QA-ing the number field validations, I discovered that--as evidenced by the libraries which already exist on QA--admins usually want to start the maximum fines field with a dollar sign.  I changed the client-side validation so that you can type in a dollar sign if it's the first character in that field.  I also changed the server-side validation so that you don't get an error message when you submit the resulting string.

In the event that we ever add another case in which people should be allowed to type a non-numeric character into a number field, I'll abstract this out and implement it more elegantly; for now, since it's the only one, hard-coding it seemed good enough.

### DIFF
--- a/src/components/EditableInput.tsx
+++ b/src/components/EditableInput.tsx
@@ -121,7 +121,11 @@ export default class EditableInput extends React.Component<EditableInputProps, E
 
   validateNumber(e) {
     const validChars = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", ".", "Enter"];
-    if (validChars.indexOf(e.key) < 0) {
+
+    // An admin might want to type a dollar sign at the beginning of the "maximum amount in fines" field.
+    let allowDollar = ((e.currentTarget.name === "max_outstanding_fines") && (e.target.selectionStart === 0));
+
+    if ((validChars.indexOf(e.key) < 0) && !allowDollar) {
       e.preventDefault();
     }
   }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.0.92",
+  "version": "0.0.94",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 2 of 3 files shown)


EdwinGuzman:
I don't think this makes sense to do and there are no other examples where number fields or ecommerce fields allow a user to enter a dollar sign. I would prefer if there is description that the amount, if added, will be in dollars. Or perhaps prefix the input field with a dollar sign outside of the input field. It doesn't seem right to me to allow the user to enter this and then strip it away in the server.

adriana-alter:
As it stands, if admins try to update an existing library, and they happen to already have a dollar sign in there, they're going to get an error message and have to change it...  (Maybe we just go ahead and make them change it.  But it seemed worth avoiding that.) . I like the idea of prefixing the field with a dollar sign, though!

src/components/EditableInput.tsx
```diff
@@ -121,7 +121,11 @@ export default class EditableInput extends React.Component<EditableInputProps, E
 
   validateNumber(e) {
     const validChars = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", ".", "Enter"];
-    if (validChars.indexOf(e.key) < 0) {
+
+    // An admin might want to type a dollar sign at the beginning of the "maximum amount in fines" field.
+    let allowDollar = ((e.currentTarget.name === "max_outstanding_fines") && (e.target.selectionStart === 0));
```
aslagle: I would rather allow dollar signs in all the number fields than have this hard-coded. The server could do more validation if necessary.

adriana-alter: Ok.  How about: I get rid of this code--i.e. we don't allow dollar signs anywhere--but we 1) use my server code changes (so that admins who already have dollar signs won't get an error message when they update their libraries), and 2) add copy on the front end, as per Edwin's suggestion, so that users don't find it surprising or confusing that they can't enter dollar signs?

aslagle: Sure.

PR #208, opened 2019-01-16
Made it possible to change the order of lanes with the same parent lane.

aslagle:
This adds drag-and-drop functionality to the list of lanes, and buttons for saving or resetting the order changes. You can only change the order within the parent, you can't move a lane to a different parent.

Unfortunately the drag and drop library broke many of enzyme's selectors, so the way the tests have to look up elements is ugly. I'm also getting some typings errors in the travis build that I'm still looking into.

### DIFF
--- a/src/components/Lanes.tsx
+++ b/src/components/Lanes.tsx
@@ -12,6 +12,7 @@ import LaneEditor from "./LaneEditor";
 import LoadingIndicator from "opds-web-client/lib/components/LoadingIndicator";
 import ErrorMessage from "./ErrorMessage";
 import EditableInput from "./EditableInput";
+import SaveButton from "./SaveButton";
 import AddIcon from "./icons/AddIcon";
 import GrabIcon from "./icons/GrabIcon";
 import PencilIcon from "./icons/PencilIcon";
@@ -138,12 +139,12 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
           { this.state.orderChanged &&
             <div className="order-change-info">
               <h2>Change Lane Order</h2>
-              <button
-                className="save-order btn btn-default"
-                onClick={this.saveOrder}
+              <SaveButton
+                submit={this.saveOrder}
+                disabled={this.props.isFetching}
+                text="Save Order Changes"
                 >
-                Save Order Changes
-              </button>
+              </SaveButton>
               <a
                 href="#"
                 className="cancel-order-changes"
--- a/src/components/Lanes.tsx
+++ b/src/components/Lanes.tsx
@@ -248,16 +248,9 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
           <div className={"lane-info" + (provided ? " draggable" : "")}>
             <span>
               { snapshot && <GrabIcon /> }
-              { this.isExpanded(lane) &&
-                <div className="collapse-button" onClick={() => { this.toggle(lane); }}>
-                  <PyramidIcon />
-                </div>
-              }
-              { !this.isExpanded(lane) &&
-                <div className="expand-button" onClick={() => { this.toggle(lane); }}>
-                  <PyramidIcon />
-                </div>
-              }
+              <div className={this.isExpanded(lane) ? "collapse-button" :  "expand-button"} onClick={() => { this.toggle(lane); }}>
+                <PyramidIcon />
+              </div>
               { lane.display_name + " (" + lane.count + ")" }
             </span>
             { lane.visible && !this.state.orderChanged &&
--- a/src/components/Lanes.tsx
+++ b/src/components/Lanes.tsx
@@ -477,7 +477,7 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
       lanesToUpdate = lanes;
     }
 
-    lanesToUpdate = lanesToUpdate.slice(0, oldIndex).concat(lanesToUpdate.slice(oldIndex + 1, lanes.length));
+    lanesToUpdate = lanesToUpdate.slice(0, oldIndex).concat(lanesToUpdate.slice(oldIndex + 1, lanesToUpdate.length));
     lanesToUpdate.splice(newIndex, 0, draggedLane);
 
     if (parent) {
--- a/src/components/Lanes.tsx
+++ b/src/components/Lanes.tsx
@@ -202,92 +202,106 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
     const draggingLane = this.findLaneForIdentifier(this.state.lanes, this.state.draggableId);
     const disabled = this.isDropDisabled(draggingLane, parent);
 
-    return (
-      <Droppable
-        droppableId={parent ? String(parent.id) : "top"}
-        isDropDisabled={disabled}
-        >
-        {(provided, snapshot) => (
-          <ul
-            ref={provided.innerRef}
-            className={(snapshot.isDraggingOver && !disabled) ? "droppable dragging-over" : "droppable"}
-            >
-            { lanes.map(lane =>
-                <Draggable draggableId={String(lane.id)} key={lane.id}>
-                {(provided, snapshot) => (
-                    <li>
-                      <div
-                        className={(snapshot.isDragging ? "dragging " : " ") + ((String(lane.id) === this.props.identifier) ? "active" : "")}
-                        ref={provided.innerRef}
-                        style={provided.draggableStyle}
-                        {...provided.dragHandleProps}
-                        >
-                        <div className="lane-info">
-                          <span>
-                            <GrabIcon />
-                            { this.isExpanded(lane) &&
-                              <div className="collapse-button" onClick={() => { this.toggle(lane); }}>
-                                <PyramidIcon />
-                              </div>
-                            }
-                            { !this.isExpanded(lane) &&
-                              <div className="expand-button" onClick={() => { this.toggle(lane); }}>
-                                <PyramidIcon />
-                              </div>
-                            }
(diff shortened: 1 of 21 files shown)


src/components/Lanes.tsx
```diff
@@ -64,17 +75,34 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
...
       expanded[topLevelLane.id] = true;
     }
 
     this.state = {
-      expanded: expanded
+      expanded: expanded,
+      draggableId: null,
+      draggingFrom: null,
+      lanes: this.copyLanes(this.props.lanes || []),
+      orderChanged: false
```
aslagle: I added a separate boolean to keep track of whether the order changed rather than repeatedly trying to compare two sets of lanes.

src/components/Lanes.tsx
```diff
@@ -150,72 +199,95 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
...
-            </li>
-          )
-        }
-      </ul>
+            { lanes.map(lane =>
+                <Draggable draggableId={String(lane.id)} key={lane.id}>
+                {(provided, snapshot) => (
+                    <li>
+                      <div
+                        className={(snapshot.isDragging ? "dragging " : " ") + ((String(lane.id) === this.props.identifier) ? "active" : "")}
```
aslagle: I moved the "active" class from the `li` just to facilitate testing, because I was having a hard time getting the `li` now that it's in the `Draggable`. It should look the same.

adriana-alter:
It's doing a weird thing whereby, when I try to drag a lane, it often gets stuck, and the lane ends up superimposed over the rest of the list (the rest of the list can be scrolled up and down behind the stuck one).  Clicking "cancel" fixes it (but still...).

screen shot 2019-01-16 at 2 58 23 pm

adriana-alter:
screen shot 2019-01-16 at 3 12 08 pm

adriana-alter:
The equal signs in front of the collapse buttons shouldn't be there...right?  (I'm trying to figure out what's generating them...)

aslagle:
That was meant to be a drag icon so you would know you could drag it. I don't have a designer :(

aslagle:
This latest commit will hopefully fix the lane getting stuck.

EdwinGuzman:
The last commit fixed the issue. It was also occurring when you tried dragging an outer lane into an inner lane.

Can you make it so that you cannot drag a lane when it's the only sublane, such as `sublane` and `test3 sublane` in the image below:
screen shot 2019-01-16 at 3 21 18 pm

aslagle:
Oh, that's a good idea.

adriana-alter:
Yes, fixed!

Re: the drag icon, I think three bars would be better than two; two just says "equal sign" to me.  How about the MenuIcon?  https://nypl.github.io/dgx-svg-icons/#!/MenuIcon . Also, I'm seeing some websites suggesting using vertical bars to indicate that something is draggable; if you rotate the MenuIcon 90 degrees...

aslagle:
Two bars is what we use on the list page so I made this match.

EdwinGuzman:
The cancel button for the lane order changes has a misaligned cancel icon:

screen shot 2019-01-16 at 3 17 11 pm

adriana-alter:
Also, I think the cancel button should be bigger/more noticeable (and more button-y looking!).

EdwinGuzman:
* Not sure if related to this PR but I cannot delete lanes. I get a 500 error.
* If you re-order the lanes but decide to create a sublane and save the new sublane, the new order will be lost.
* If you re-order the lanes but decide to hide a lane, or make a lane visible, the new order will be lost.

There's a lot going on in this `Lanes` page. I would not be opposed to creating a new page where all you do is reorder the lanes. This seems like a big enough feature that admins can use (and a big user-facing feature) that merits its own page _just_ for reordering.

aslagle:
The cancel button was also meant to match other pages with save/cancel - not the misalignment of it though.

adriana-alter:
I don't know whether it's a result of this PR, but: the number next to the lane display name is supposed to indicate how many sub-lanes the lane has...yes?  If so, problem:
screen shot 2019-01-16 at 3 35 29 pm

aslagle:
No, that's supposed to be how many books are in it.

adriana-alter:
+1 to Edwin's suggestion about breaking the lane re-ordering into its own page.

aslagle:
I think it will be a lot easier to fix this page than make a new one.

adriana-alter:
Attempting to save order changes is consistently giving me a 404.  `{status: 404, response: "Request failed", url: "/lib/admin/lanes/change_order"}`

aslagle:
@adriana-alter you need the circulation branch too.

adriana-alter:
How about adding a success message (for consistency with how saving other stuff works), or not worth it?

src/components/Lanes.tsx
```diff
@@ -1,4 +1,8 @@
 import * as React from "react";
+const dnd = require("react-beautiful-dnd");
```
EdwinGuzman: Just curios, any reason why you imported it this way instead of `import { DragDropContext, Droppable, Draggable } from "react-beautiful-dnd";`?

aslagle: I was copying from the list page and I don't remember why I did it there, but probably this version of the library doesn't support the other way.

EdwinGuzman: It runs fine using this version, too. Whichever is fine but just for consistency.

aslagle: you tried the import and it worked?

EdwinGuzman: Yes.

aslagle: cool, I would rather switch then.

adriana-alter:
You still have the old version of the react-beautiful-dnd import in LaneCustomListsEditor and LaneCustomListsEditor-test.

src/components/Lanes.tsx
```diff
@@ -116,15 +116,15 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
             <h2>Lane Manager</h2>
             <div>
               <Link
-                className="create-lane btn btn-default"
-                to={"/admin/web/lanes/" + this.props.library + "/create"}
+                className={"create-lane btn btn-default" + (this.state.orderChanged ? " disabled" : "")}
+                to={this.state.orderChanged ? null : ("/admin/web/lanes/" + this.props.library + "/create")}
```
adriana-alter: Why bother making the Link's `to` attribute conditional?  If `this.state.orderChanged` is true, then the link is disabled anyway; clicking on it won't do anything no matter what `to` is.

src/components/Lanes.tsx
```diff
@@ -130,29 +130,33 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
...
               { this.state.lanes && this.state.lanes.length > 0 && this.renderLanes(this.state.lanes, null) }
             </DragDropContext>
           </div>
 
-          { this.props.editOrCreate === "create" &&
+          { this.state.orderChanged &&
+            <div className="order-change-info">
+              <h2>Change Lane Order</h2>
+              <button
+                className="save-order btn btn-default"
```
adriana-alter: This is not necessarily worth it, but if you wanted to make this (very slightly!) shorter/more readable, the reusable SaveButton component would work here.

aslagle: It needs a form and there isn't one in this case...

aslagle: Although it doesn't actually seem to use the form.

adriana-alter: I was just thinking that; investigating now to see what's up with the form!

adriana-alter: It definitely doesn't need the form (I tried it out with a component that already uses the SaveButton).

adriana-alter:
Note: I'm getting ready to push a branch that updates to 0.0.98, so (depending on timing) this one is probably going to be 0.0.99.

src/components/Lanes.tsx
EdwinGuzman: Yeah, it works with just the `disabled` class.

aslagle: I think the `disabled` class relies on CSS pointer events which are not supported in safari? But I did not actually test in safari so it may have some other fallback that works. I just thought it was safer to also remove the href.

EdwinGuzman: Yeah, I'd keep the conditional `to` here just in case. Or do more cross browser testing.

src/components/Lanes.tsx
```diff
@@ -150,72 +201,112 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
...
+                <div className="collapse-button" onClick={() => { this.toggle(lane); }}>
+                  <PyramidIcon />
                 </div>
               }
-              { this.isExpanded(lane) && lane.sublanes && lane.sublanes.length > 0 && this.renderLanes(lane.sublanes, lane) }
-            </li>
-          )
-        }
-      </ul>
+              { !this.isExpanded(lane) &&
```
EdwinGuzman: I know this isn't new but the `PyramidIcon` components could be refactored to a conditional where only the class name is different since it whether or now `this.isExpanded(lane)` is true or false, this always renders.

EdwinGuzman:
I agree that `Lanes.tsx` is very long. We use the same library for the custom list entries editor. It might be possible to refactor them to also have droppable and draggable components, but that might not be worth it (for this PR or in general).

adriana-alter:
Assuming we're not going to mess with refactoring `Lanes.tsx` in this PR, everything looks good to me!

PR #210, opened 2019-02-05
Diagnostics

adriana-alter:
screen shot 2019-02-05 at 4 46 21 pm

- Tests are still very much in progress.  Might as well just ignore them for now.
- Formatting/styling is still in progress.
- I am planning to add a lot of comments.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.1.00
```
EdwinGuzman: This should be `v0.1.0`.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+#### Added
+- Created a Diagnostics page so that admins can view the output of scripts.
```
EdwinGuzman: Can you mention that they are log scripts that run behind the scenes by the server?

src/components/DiagnosticsPage.tsx
```diff
@@ -0,0 +1,55 @@
...
+  goToTab(tab: string) {
+    this.setState({ tab });
+  }
+
+  render(): JSX.Element {
+    return(
+      <div className="diagnostics-page">
+        <Header />
+        <div className="body">
+          <h1>Diagnostics</h1>
```
EdwinGuzman: `h1` is typically reserved for the home page. To be consistent with the other pages this should be an `h2`.

src/components/DiagnosticsServiceTabs.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  handleSelect(event) {
+    let tab = event.currentTarget.dataset.tabkey;
+    this.props.goToTab(tab);
+  }
+
+  tabs() {
+    let tabs = {};
+    let serviceNames = Object.keys(this.props.content);
+    serviceNames.map((serviceName) => {
```
EdwinGuzman: Use `forEach`.

src/components/DiagnosticsServiceTabs.tsx
```diff
@@ -0,0 +1,55 @@
...
+    });
+    return tabs;
+  }
+
+  tabDisplayName(name) {
+    // Get a flattened array of all the timestamps associated with this service.
+    let timestampArray = [].concat(...Object.values(this.props.content[name]));
+    let hasException = timestampArray.some((ts) => ts.exception);
+    // If any of the timestamps has an exception, display a warning.  Otherwise, display the number of timestamps.
+    let badge = hasException ? <span className="badge danger">!</span> : <span className="badge">{timestampArray.length}</span>;
```
EdwinGuzman: These badges are not self-explanatory and they are confusing if you just see a number without an explanation. I had to read the code to understand that they were the number of timestamps. But is it the number of timestamps total or the number or collections with timestamps?

src/components/DiagnosticsServiceTabs.tsx
```diff
@@ -0,0 +1,55 @@
...
+        body={this.renderTimestamps(collections[collectionName])}
+      />
+    );
+  }
+
+  renderTimestamps(timestamps: Array<TimestampData>): JSX.Element {
+    let tsList = timestamps.map(timestamp =>
+      <li><Timestamp timestamp={timestamp} /></li>
+    );
+    return <ul>{tsList}</ul>;
```
EdwinGuzman: What if there are no timestamps for whatever reason? You would want to display some kind of message.

src/components/DiagnosticsTabContainer.tsx
```diff
@@ -0,0 +1,98 @@
...
+    if (this.DISPLAY_NAMES[name]) {
+      return this.DISPLAY_NAMES[name];
+    } else {
+      return super.tabDisplayName(name);
+    }
+  }
+
+  tabs() {
+    let tabs = {};
+    let serviceTypes = ["coverage_provider", "monitor", "script", "other"];
```
EdwinGuzman: I think this makes this display:
screen shot 2019-02-07 at 2 29 20 pm

src/components/TabContainer.tsx
```diff
@@ -4,8 +4,8 @@ import { Navigate, PathFor } from "../interfaces";
 import { State } from "../reducers/index";
 
 export interface TabContainerProps extends React.Props<TabContainerProps> {
-  store: Store<State>;
-  csrfToken: string;
+  store?: Store<State>;
+  csrfToken?: string;
```
EdwinGuzman: Why are these optional now?

src/interfaces.ts
```diff
@@ -145,6 +145,31 @@ export interface StatsData {
   [key: string]: LibraryStatsData;
 }
 
+export interface DiagnosticsData {
+  monitor?: Array<DiagnosticsServiceData>;
```
EdwinGuzman: Any reason why you are using `Array<DiagnosticsServiceData>` instead of `DiagnosticsServiceData[]`? I'd use the latter to be consistent with the convention we've been using.

src/reducers/diagnostics.ts
```diff
@@ -0,0 +1,6 @@
+import ActionCreator from "../actions";
+import { DiagnosticsData } from "../interfaces";
+
+import createFetchEditReducer from "./createFetchEditReducer";
+
+export default createFetchEditReducer<DiagnosticsData>(ActionCreator.DIAGNOSTICS);
```
EdwinGuzman: You may need another reducer since all you are doing is fetching data and not editing anything.

src/stylesheets/diagnostics.scss
```diff
@@ -0,0 +1,109 @@
...
+        .config {
+          width: 100%;
+          .tab-container {
+            .nav-tabs {
+              border-bottom: 1px solid $medium-gray;
+              border-right: 1px solid $medium-gray;
+              border-top: none;
+              li {
+                border-color: $medium-gray;
+                a {
```
EdwinGuzman: This is nested about 8 class names deep! This is not very efficient for the compiler since it'll compile it to `.diagnostic-page .body .tab-container .tab-content ......`. I'd suggest to break it apart to make it more readable and not go more than 4-to-5 class deep. You could also break it up per component (DiagnosticsServiceTabs, DiagnosticsServiceTypes, etc).

EdwinGuzman: For other NYPL apps this is one of the rules we've tried to follow: https://smacss.com/book/applicability

src/components/DiagnosticsServiceTabs.tsx
adriana-alter: Given how the server code works, there's no possible way that a service type could have services and collections but no timestamps.

src/components/DiagnosticsServiceTabs.tsx
adriana-alter: It's the number of timestamps total.  What do you think would make that clearer?  Or do you think it's not worth displaying the number at all?

src/components/TabContainer.tsx
adriana-alter: Not every implementation of the `TabContainer` necessarily needs the store and the token--for example, `DiagnosticsServiceTabs` doesn't communicate with the server, so it has no use for these.  I figured it made more sense to make them optional than to pass around a bunch of irrelevant props which will never get used.

src/reducers/diagnostics.ts
adriana-alter: Are there downsides to leaving it like this?  This does work...

src/components/DiagnosticsServiceTabs.tsx
EdwinGuzman: No, keep them there, it looks nice. But maybe have text after it saying "total timestamps"?

src/reducers/diagnostics.ts
EdwinGuzman: Not really, but it's adding extra functionality that is not being used and could be simpler.

src/components/DiagnosticsServiceTabs.tsx
adriana-alter: If there's text after it, it will no longer look nice!  What about a tooltip?  Or are there accessibility issues with that?

EdwinGuzman: Could you make the left side tabs wider and the right side content narrower? Don't think tooltips would be all that accessible but I'd be okay with that.

adriana-alter:
The Bootstrap tooltip seemed to be needlessly complicated to implement, so I made a simpler one from scratch instead.

screen shot 2019-02-08 at 10 58 15 am

src/stylesheets/diagnostics.scss
```diff
@@ -106,3 +106,36 @@ pre {
...
+    width: 300px;
+    padding: 8px 0;
+    text-align: center;
+    vertical-align: middle;
+    border-radius: 4px;
+    color: white;
+    background: $dark-gray;
+    opacity: .75;
+  }
+  .tool-tip::after {
```
EdwinGuzman: I think the `::before` and `::after` are pseudo classes that make elements inaccessible. But I don't mind as long as we can work on this later. Just don't forget.

adriana-alter: My understanding is that the ::after content will be inaccessible, but that's just the arrow, so it doesn't matter; the accessibility of the actual text inside the element won't be affected.  (But yes, we can work on it later!)

src/stylesheets/diagnostics.scss
```diff
@@ -1,109 +1,108 @@
...
-  height: 100%;
-  .body {
-    min-height: 80%;
-    h1  {
-      margin: 20px 15px;
+  h2  {
+    margin: 20px 15px;
+  }
+  // Service Type tabs
+  .body > .tab-container > .nav-tabs {
```
EdwinGuzman: Doing this and what was there before are the same. Perhaps you can do something like 

```
.diagnostic-page {
  h2 {}
  .tab-container .nav-tabs { ... }
  .config .nav-tabs { ... }
  .tab-content {
    .collapsible {
      .timestamps {
         padding: 0;
      }
}

.timestamps {
 .collapsible { }
}
```

That way you're defining basic rules for the inner classes but then extracting them out and making them their own separate thread. You can't do that with `.nav-tabs` because you have that class under `tab-container` and `config` but it'd be good to make it more readable.

src/stylesheets/diagnostics.scss
```diff
@@ -1,109 +1,108 @@
...
+    background: $red-tint;
   }
-  .exception {
-    border-color: $red-tint;
-    background: $red-error;
-    box-shadow: inset 0 .5px .5px $red-tint;
+  border-color: $red-tint;
+}
+
+.diagnostics-page .badge {
```
EdwinGuzman: This `.badge` can be moved into the larger `.diagnostic-page` rule.

src/stylesheets/diagnostics.scss
```diff
@@ -1,109 +1,108 @@
...
+  .panel-title {
+    background: $green-tint;
   }
+  border-color: $green-tint;
+}
 
-  pre {
-    border: none;
-    background: inherit;
+.timestamps .collapsible.panel.panel-danger {
```
EdwinGuzman: You can combine this with the above
```
.timestamps {
  .collapsible {
    &.panel {
       &.panel-success {}
       &.pane-danger {}
     }
  }
}
```

src/stylesheets/diagnostics.scss
EdwinGuzman: That's true. But then jumping to that text inside or reading it outloud by screenreaders will be difficult.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.0.99",
+  "version": "0.1.00",
```
EdwinGuzman: This should be `0.1.0`.

src/components/__tests__/DiagnosticsPage-test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  it("renders a tab container", () => {
+    let container = wrapper.find(DiagnosticsTabContainer);
+    expect(container.length).to.equal(1);
+  });
+
+  it("switches tabs", () => {
+    expect(wrapper.state()["tab"]).to.equal("coverage_provider");
+    expect(wrapper.find(DiagnosticsTabContainer).prop("tab")).to.equal("coverage_provider");
+
+    wrapper.instance().goToTab("monitor");
```
EdwinGuzman: Nice.  For a previous PR I mentioned how if you tested a component's method in isolation, you would want to pass parameters instead of using `this.props....` because it would be easier to test. This is what I meant. I don't think we ever really test individual methods but we could start for more code coverage (not needed for this PR).

src/components/__tests__/DiagnosticsServiceTabs-test.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+      let timestamp1 = timestamps.at(0);
+      expect(timestamp1.prop("timestamp")).to.equal(ts1);
+
+      let timestamp2 = timestamps.at(1);
+      expect(timestamp2.prop("timestamp")).to.equal(ts2);
+    });
+  });
+
+  it("switches tabs", () => {
```
EdwinGuzman: Maybe this could go under a `"behavior"` description but not a big deal!

src/components/__tests__/ToolTip-test.tsx
```diff
@@ -0,0 +1,64 @@
...
+  });
+
+  it("initially hides the tooltip", () => {
+    expect(wrapper.state()["show"]).to.be.false;
+    expect(wrapper.find(".tool-tip").hasClass("hide")).to.be.true;
+  });
+
+  it("shows the tooltip on mouseEnter", () => {
+    let spyShow = spy(wrapper.instance(), "show");
+    wrapper.setProps({ show: spyShow });
```
EdwinGuzman: `show` is suppose to be a boolean but here you are setting it to a spy function.  Odd that it still works.

src/components/__tests__/ToolTip-test.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    spyShow.restore();
+  });
+
+  it("hides the tooltip on mouseLeave", () => {
+    wrapper.setState({ show: true });
+    expect(wrapper.find(".tool-tip").hasClass("hide")).to.be.false;
+
+    let spyHide = spy(wrapper.instance(), "hide");
+    wrapper.setProps({ hide: spyHide });
```
EdwinGuzman: Same as the above. Also, `<ToolTip>` does not have `show` or `hide` props but only a `show` state variable. So there's a bit of confusion here.

src/stylesheets/diagnostics.scss
```diff
@@ -0,0 +1,98 @@
...
+  .services .tab-content {
+    padding: 0;
+    min-width: 70%;
+    height: 100%;
+    // Individual collection
+    .collapsible {
+      margin-bottom: .5em;
+      &:not(.panel-success){
+        &:not(.panel-danger) {
+          border-color: $medium-gray;
```
EdwinGuzman: Hmm is this meant to be `.collapsible:not(.panel-success):not(.panel-danger)` or `.collapsible:not(.panel-success), .collapsible:not(.panel-danger)`?

adriana-alter: I tried all the variations; this is the one that works as intended.

src/components/__tests__/ToolTip-test.tsx
adriana-alter: It has functions called `show` and `hide`, not props.  I agree that it's confusing for both the state variable and the function to be called `show`, though; will change!

src/components/__tests__/ToolTip-test.tsx
adriana-alter: There's also a function named `show` (see comment below)

src/components/__tests__/ToolTip-test.tsx
EdwinGuzman: The state and function being the same name is not confusing. It's that you're setting a prop called `hide` here but you don't call `this.props.hide` in the component. I know there's a `hide` function, but you shouldn't have to set that with `setProps`.

adriana-alter: I assume it's a weird Enzyme/Sinon thing; the spy function only works if I set it as a prop.

adriana-alter: Oh, and there's a `.hide` className.  This is officially too many things with the same name; I'm confusing myself now!  Changing the names.

PR #211, opened 2019-02-11
Timestamp shouldn't collapse, & styling improvements

adriana-alter:


src/components/DiagnosticsServiceTabs.tsx
```diff
@@ -36,7 +36,7 @@ export default class DiagnosticsServiceTabs extends TabContainer<DiagnosticsServ
         <span className="badge danger">!</span> :
         <ToolTip trigger={<span className="badge">{timestampArray.length}</span>} text="Total number of timestamps for this service"/>;
 
-    return <span>{super.tabDisplayName(name)}{badge}</span>;
+    return <section><span className="service-name">{super.tabDisplayName(name)}</span>{badge}</section>;
```
EdwinGuzman: This made me check and see that the tooltip container is a section, too. That's not needed since that's not actually a semantic section, it's helper text.

adriana-alter: What should it be, in that case--another span, or...?

src/stylesheets/collapsible.scss
```diff
@@ -31,15 +31,17 @@
     }
   }
   &.panel-success {
-    border-color: $green-tint;
+    border-color: transparentize($green-bright, .4);
```
EdwinGuzman: This color is nicer.

EdwinGuzman:
Looks good. Update package.json.

src/components/DiagnosticsServiceTabs.tsx
EdwinGuzman: A div is fine but not a big deal.

PR #212, opened 2019-02-12
Show metadata service name instead of protocol.

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1717


EdwinGuzman:
This is fine but I suggest updating this so it's consistent to what we have now on the Patron Authentication screen- `[title]: [protocol]`.
screen shot 2019-02-12 at 4 14 19 pm

PR #213, opened 2019-02-21
Elastic search self tests

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1715; goes with these [core](https://github.com/NYPL-Simplified/server_core/pull/1030) and [circulation](https://github.com/NYPL-Simplified/circulation/pull/1184) branches.

### DIFF
--- a/src/components/SelfTestResult.tsx
+++ b/src/components/SelfTestResult.tsx
@@ -44,22 +44,24 @@ export default class SelfTestResult extends React.Component<SelfTestResultProps,
   }
 
   renderResult(result, colorResultClass) {
-    let list = Array.isArray(result);
+    let isList = Array.isArray(result);
 
-    let content = list ?
+    let content = isList ?
       <ol>{result.map((item, idx) => <li key={idx}>{item}</li>)}</ol> :
       result;
 
     let body = <pre className="result-description">{content}</pre>;
 
-    // If the result is a list, or is more than one line long (in which case it's probably a stringified JSON object),
-    // display it in a collapsible panel.
-    let collapsible = (list || result.split("\n").length > 1);
-    return (collapsible && this.renderCollapsible(result, body, colorResultClass, list)) || body;
+    // If the result is a list, or is more than one line long, display it in a collapsible panel. (So far, the most likely
+    // way to have a multi-line result is if it's from a search service test that returns a prettified JSON object,
+    // but future tests might return normal strings which happen to be very long.)
+
+    let collapsible = (isList || result.split("\n").length > 1);
+    return (collapsible && this.renderCollapsible(result, body, colorResultClass, isList)) || body;
   }
 
-  renderCollapsible(result, body, colorResultClass, list) {
-    let title = list ? `Results (${result.length})` : "Results";
+  renderCollapsible(result, body, colorResultClass, isList) {
+    let title = isList ? `Results (${result.length})` : "Results";
 
     return <Collapsible
       title={title}
--- a/src/components/SelfTestResult.tsx
+++ b/src/components/SelfTestResult.tsx
@@ -1,0 +1,70 @@
+import * as React from "react";
+import Collapsible from "./Collapsible";
+import { SelfTestsResult } from "../interfaces";
+
+export interface SelfTestResultProps {
+  result: SelfTestsResult;
+  isFetching: boolean;
+}
+
+export default class SelfTestResult extends React.Component<SelfTestResultProps, void> {
+  constructor(props) {
+    super(props);
+    this.renderResult = this.renderResult.bind(this);
+    this.renderCollapsible = this.renderCollapsible.bind(this);
+  }
+
+  render() {
+    const colorResultClass = this.props.result.success ? "success" : "failure";
+    return (
+      <li className={this.props.isFetching ? "loading-self-test" : colorResultClass} key={this.props.result.name}>
+        <h4>{this.props.result.name}</h4>
+        {
+          this.props.result.success && this.props.result.result && !this.props.result.result.length ?
+          <span className="warning">The test ran successfully, but no results were found.</span>
+          : null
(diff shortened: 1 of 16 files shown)


EdwinGuzman:
* When you don't have the elasticsearch server running you get this error:
screen shot 2019-02-21 at 5 27 19 pm
I'm not too concern with this and that error is kind of useful.
* When you don't have any tests running you get this message and I do think it should be fixed:
screen shot 2019-02-21 at 5 27 57 pm

src/reducers/searchServices.ts
```diff
@@ -2,4 +2,33 @@ import { SearchServicesData } from "../interfaces";
 import ActionCreator from "../actions";
 import createFetchEditReducer from "./createFetchEditReducer";
 
-export default createFetchEditReducer<SearchServicesData>(ActionCreator.SEARCH_SERVICES, ActionCreator.EDIT_SEARCH_SERVICE);
+// This is only to update a specific search service's self test results property.
+const loadCB = (state, action) => {
```
EdwinGuzman: This is very similar to what the `collections` reducer uses. Can you refactor it so that they both share one function but that the data variable property name changes?

src/components/SelfTests.tsx
```diff
@@ -143,6 +146,16 @@ export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
     );
   }
 
+  renderResults(result) {
+    if (Array.isArray(result)) {
+      let resultList = result.map((item, idx) => <li key={idx}>{item}</li>);
+      return <ol className="result-description">{resultList}</ol>;
+    }
+    else {
```
EdwinGuzman: Just have the else on the previous line.

src/components/SelfTests.tsx
```diff
@@ -188,7 +201,7 @@ function mapDispatchToProps(dispatch, ownProps) {
   let fetcher = new DataFetcher();
   let actions = new ActionCreator(fetcher, ownProps.csrfToken);
   const integrationId = ownProps.item.id;
-  const url = `/admin/collection_self_tests/${integrationId}`;
+  const url = `/admin/${ownProps.type.replace(" ", "_")}_self_tests/${integrationId}`;
```
EdwinGuzman: 👍

src/components/SelfTestResult.tsx
```diff
@@ -0,0 +1,70 @@
...
+          <span className="warning">The test ran successfully, but no results were found.</span>
+          : null
+        }
+        {
+          this.props.result.result && this.props.result.result.length ?
+            this.renderResult(this.props.result.result, colorResultClass)
+            : null
+        }
+        <p className="success-description">
+          success: {`${this.props.result.success}`}
```
aslagle: Is `props.result.success` a boolean?

src/components/SelfTestResult.tsx
```diff
@@ -0,0 +1,70 @@
...
+              exception: {this.props.result.exception.message}
+            </p>
+          )
+        }
+      </li>
+    );
+  }
+
+  renderResult(result, colorResultClass) {
+    let list = Array.isArray(result);
```
aslagle: I'd recommend naming things like this `isList` or `isCollapsible`

src/components/SelfTestResult.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    let content = list ?
+      <ol>{result.map((item, idx) => <li key={idx}>{item}</li>)}</ol> :
+      result;
+
+    let body = <pre className="result-description">{content}</pre>;
+
+    // If the result is a list, or is more than one line long (in which case it's probably a stringified JSON object),
```
aslagle: If a stringified JSON object is more than one line long, it's probably being unnecessarily prettified.

Either way, it's not a very reliable way to check if something's JSON. I would try to parse it and see if it works.

src/components/SelfTestResult.tsx
adriana-alter: Yes

src/components/SelfTestResult.tsx
adriana-alter: The point isn't to check whether it's JSON (but I realize the comment is very misleading; I'll change that!)--the point is just to check whether it's more than one line.  At the moment, the most likely way that the result would be a multi-line string is that it would be a prettified JSON object...but, in theory, the result could just be a normal string that happens to be more than one line long, in which case we would still want it to be displayed in a collapsible.  

(Re: unnecessary prettification--I think that's a word...?!-- I initially tried just displaying un-prettified JSON here, and it was way too hard to read; I'd rather err on the side of having the results taking up too much space than of having them be illegible.)

aslagle: That's true, I don't think you'd want to display unprettified JSON. But is there nothing better you can pull out of the JSON to display? If it's a problem detail you would know in advance what fields it has.

adriana-alter: Unfortunately not; the whole point of some of the new search integration tests is specifically that they display the entire JSON object--e.g. the full search document (as pictured below), or the raw search results:
screen shot 2019-02-28 at 9 53 54 am

aslagle: Ah, maybe I should have looked at the back-end part.

With the query stuff at the top like that it isn't immediately obvious whether it's working. Could you pull out the part with actual results?

adriana-alter: There's another test that shows the actual results--the point of this one is just to show what the query is, not whether it worked.  If this helps: at the moment, the search self-tests are 1) a list of titles and authors, 2) the search document, 3) the raw results (i.e. the full version of what #1 was pulled from), 4) the total number of search results for this search term, 5) the total number of documents in this search index, and 6) the total number of documents for each collection.

PR #214, opened 2019-03-01
Patron auth self tests

adriana-alter:
Goes with https://github.com/NYPL-Simplified/circulation/pull/1197
Resolves https://jira.nypl.org/browse/SIMPLY-1730

### DIFF
--- a/src/components/SelfTests.tsx
+++ b/src/components/SelfTests.tsx
@@ -86,20 +86,28 @@ export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
       `Tests last ran on ${startDate} ${startTime} and lasted ${duration}s.` :
       "No self test results found.";
 
-    if (this.props.type === "patron authentication service" && !this.props.item.libraries.length) {
-      disableToggleView = true;
-      testDescription = "You must associate this service with at least one library before you can run self tests for it.";
-    }
-
     const failedSelfTest = selfTestException ? selfTestException : "";
+    const disableToggle = integration.self_test_results && integration.self_test_results.disabled;
 
-    const button =  (<button
-                      onClick={(e) => this.runSelfTests(e)}
-                      className="btn btn-default runSelfTests"
-                      disabled={this.props.isFetching}
-                    >
-                      Run tests
-                    </button>);
+    const toggleButton = (
+      <button
+        onClick={this.toggleView}
+        disabled={disableToggle}
+        className="btn btn-default"
+      >
+        {resultsLabel} Results
+      </button>
+    );
+
+    const runButton = (
+      <button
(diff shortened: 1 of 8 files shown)


src/components/SelfTests.tsx
```diff
@@ -81,10 +82,15 @@ export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
...
     const isFetching = !!(this.props.isFetching && this.state.runTests);
 
-    const testDescription = integration.self_test_results && integration.self_test_results.start ?
+    let testDescription = integration.self_test_results && integration.self_test_results.start ?
       `Tests last ran on ${startDate} ${startTime} and lasted ${duration}s.` :
       "No self test results found.";
 
+    if (this.props.type === "patron authentication service" && !this.props.item.libraries.length) {
+      disableToggleView = true;
+      testDescription = "You must associate this service with at least one library before you can run self tests for it.";
```
aslagle: This seems to me like a problem detail that should come from the server.

adriana-alter: That was my initial plan, but it turned out to be a lot more complicated and didn't seem worth it to me...and I couldn't think of any specific disadvantages to doing it this way...  What do you think?  (Also, I wouldn't want it to display like an error message; no need to alarm admins by making it look like there's a problem when there isn't!)

aslagle: It kind of is a problem though - that would probably mean the admin forgot to add a library.

Also, it seems like this would depend on the specific service type. Some services might require library-specific info to run all their tests but others might not.

adriana-alter: Come to think of it: should libraries be required when an admin sets up a patron auth service?

This message only gets displayed if it's a patron auth service.  Or do you mean there are types of patron auth service that might not need libraries to run tests...?

aslagle: Yeah, I mean types of patron auth service. Some of them might be able to run tests without using any library settings.

I don't think we should require a library when saving a patron auth service since I think there are scenarios where you might want to temporarily have one without a library, but in most cases it's probably a mistake.

aslagle: I think Clever and FirstBook probably could for example, I don't know if we have self tests for them written yet though.

adriana-alter: Ok--I'll re-investigate doing this server-side.  Which types of patron auth service would be able to run tests without library settings (so that I can make sure to leave them out of this check)?

adriana-alter: We have one for FirstBook, but I think that particular test does need a library...  I'll check everything for now, and then reassess once we have more self-tests!

PR #215, opened 2019-03-07
Terms of service

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-467 (we decided to--at least for now--just do the terms-of-service checkbox, not the email addresses).

(The refactoring is just because the component was really long and I couldn't see what I was doing; doesn't affect functionality.)

screen shot 2019-03-07 at 11 14 20 am

### DIFF
--- a/src/components/LibraryRegistration.tsx
+++ b/src/components/LibraryRegistration.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import { ServiceEditFormProps } from "./ServiceEditForm";
 import EditableInput from "./EditableInput";
 import LibraryRegistrationForm from "./LibraryRegistrationForm";
-import { ServicesWithRegistrationsData, LibraryDataWithStatus } from "../interfaces";
+import { ServicesWithRegistrationsData, LibraryDataWithStatus, LibraryData } from "../interfaces";
 
 export interface LibraryRegistrationState {
   registration_stage?: { [key: string]: string } | null;
@@ -15,6 +15,12 @@ export interface LibraryRegistrationProps extends ServiceEditFormProps<ServicesW
 }
 
 export default class LibraryRegistration extends React.Component<LibraryRegistrationProps, LibraryRegistrationState> {
+  TEXT = {
+      "success": [ "Registered", "Update registration" ],
+      "warning": [ "Not registered", "Register" ],
+      "failure": [ "Registration failed", "Retry registration" ]
+  };
+
   constructor(props) {
     super(props);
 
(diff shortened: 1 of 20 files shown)


src/components/LibraryRegistration.tsx
```diff
@@ -14,111 +15,138 @@ export interface LibraryRegistrationProps extends ServiceEditFormProps<ServicesW
 }
 
 export default class LibraryRegistration extends React.Component<LibraryRegistrationProps, LibraryRegistrationState> {
+  TEXT = {
```
EdwinGuzman: The variable name could be more descriptive.

EdwinGuzman: Also, maybe instead of an array an object would be better to be descriptive about what each string is and will be used for. It just makes looking at it and know what it will be used for easier instead of looking for it in the code (ie the second value is used for the button label).

src/components/LibraryRegistration.tsx
```diff
@@ -14,111 +15,138 @@ export interface LibraryRegistrationProps extends ServiceEditFormProps<ServicesW
...
+        </div>
+      );
+    }
+    return null;
+  }
+
+  libraryRegistrationItem(library: LibraryData): JSX.Element {
+    let statusString = this.getStatus(library);
+    return (
+      <div className="service-with-registrations-library" key={library.short_name}>
```
EdwinGuzman: If you're going to refactor this then you should make this a ul/li for the map function.

src/components/LibraryRegistrationForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  toggleChecked(): void {
+    this.setState({ checked: !this.state.checked });
+  }
+
+  componentWillReceiveProps(nextProps): void {
+    this.setState({ checked: nextProps.checked });
+  }
+
+  render(): JSX.Element {
```
EdwinGuzman: I just realized that I didn't mention this before for other components but I am _use_ to having the `render` method be the last method in a React class, but once I started working on this repo I noticed that the `render` function was always the first one so I tried keeping it that way. I noticed other newer components don't follow this convention and it's not a big deal but we should stick with one. And since it was `render` goes first (after any lifecycle methods) then we should try to stick with it.

src/components/__tests__/LibraryRegistrationForm-test.tsx
```diff
@@ -0,0 +1,117 @@
...
+    expect(button.prop("disabled")).to.be.true;
+  });
+
+  it("disables the button if the disabled prop is set to true, regardless of the checkbox", () => {
+    expect(wrapper.state()["checked"]).to.be.true;
+    wrapper.setProps({ disabled: true });
+    let button = wrapper.find("button");
+    expect(button.prop("disabled")).to.be.true;
+
+    wrapper.setProps({ library: libraryNoLink });
```
EdwinGuzman: So a library without a terms of service link should not be able to register?

EdwinGuzman:
This seems fine but a few questions and stylistic changes.

I think it's interesting that now I have a harder time seeing what is going on than before. There are a lot more functions to look through since things are broken down to even smaller pieces.

src/components/__tests__/LibraryRegistrationForm-test.tsx
adriana-alter: Yes it should; the checkbox won't be rendered for it, is all.  This test is checking something else: if you set the `disabled` prop to true, that should override everything else--i.e. the button should be disabled no matter what, irrespective of anything having to do with the terms of service link.

adriana-alter:
It is very interesting that you're having a harder time seeing what's going on...  I couldn't see at all what was going on before.  Any suggested middle ground?

src/components/__tests__/LibraryRegistrationForm-test.tsx
EdwinGuzman: Right. I saw that in the admin and wanted to make sure it was right.

EdwinGuzman:
I was more pointing out that I was having a harder time at first glance. If you read it closely it's fine, but it's that clear on first look. No middle ground, this is fine.

Also, I didn't mean for you to change EVERY component and its render function placement _right now_, but thanks for doing so.

PR #216, opened 2019-03-11
Minor bug fix - Library registration form

adriana-alter:
Very minor bug (probably introduced by me in the last round of refactoring): in cases where the library registration failed, the class name needs to be set to "bg-danger" (which will turn the background red) not to "bg-failure" (which won't do anything).

I'm not sure it's worth publishing a new version just for this; what do you think?  Any other small things we can put into 0.1.5?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.1.5
+#### Fixed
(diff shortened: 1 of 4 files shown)


EdwinGuzman:
For a stylistic bug, I think it's okay to wait before merging.

PR #217, opened 2019-03-15
Geographic areas

adriana-alter:
Goes with the server-side `geographic-areas` branch.

Library edit form now:
1) contains a "Geographic Areas" section
2) lets you put in multiple values for each area
3) displays a tooltip with instructions about how to format the input 
4) if you've entered a zip code, displays a tooltip telling you where it corresponds to, so you can make sure you entered it correctly.  (That's kind of a placeholder; if/when we add a map component, I figure that's where it will go.)

Lots of refactoring in `ProtocolFormField`, including making the thing for entering multiple values its own component (`InputList`); except for the tooltip stuff, this was just so I could see what I was doing, and doesn't affect functionality.

Screen Shot 2019-03-15 at 12 36 07 PM
Screen Shot 2019-03-15 at 12 35 37 PM
Screen Shot 2019-03-15 at 12 35 26 PM

### DIFF
--- a/src/components/EditableInput.tsx
+++ b/src/components/EditableInput.tsx
@@ -11,6 +11,7 @@ export interface EditableInputProps extends React.HTMLProps<EditableInput> {
   optionalText?: boolean;
   validation?: string;
   clientError?: boolean;
+  extraContent?: string | JSX.Element;
 }
 
 export interface EditableInputState {
@@ -60,7 +61,10 @@ export default class EditableInput extends React.Component<EditableInputProps, E
             { this.props.type === "radio" && <span>{this.props.label}</span> }
           </label>
         }
-        {!this.props.label && this.renderElement()}
+        <section className={this.props.extraContent ? "with-add-on" : ""}>
+          { this.props.extraContent }
+          {!this.props.label && this.renderElement()}
+        </section>
         {description.trim() && this.renderDescription(description)}
       </div>
     );
--- a/src/components/EditableInput.tsx
+++ b/src/components/EditableInput.tsx
@@ -61,10 +61,10 @@ export default class EditableInput extends React.Component<EditableInputProps, E
             { this.props.type === "radio" && <span>{this.props.label}</span> }
           </label>
         }
-        <section className={this.props.extraContent ? "with-add-on" : ""}>
+        <div className={this.props.extraContent ? "with-add-on" : ""}>
           { this.props.extraContent }
           {!this.props.label && this.renderElement()}
-        </section>
+        </div>
         {description.trim() && this.renderDescription(description)}
       </div>
     );
--- a/src/components/InputList.tsx
+++ b/src/components/InputList.tsx
@@ -9,7 +9,6 @@ export interface InputListProps {
   labelAndDescription: (SettingData) => JSX.Element[];
   setting: SettingData;
   disabled: boolean;
-  // renderToolTip: (item: {} | string, format: string) => JSX.Element;
   value: Array<string | {}>;
 }
 
--- a/src/components/InputList.tsx
+++ b/src/components/InputList.tsx
@@ -7,12 +7,12 @@ export interface InputListProps {
   labelAndDescription: (SettingData) => JSX.Element[];
   setting: SettingData;
   disabled: boolean;
-  toolTip: (item: any, format: string) => JSX.Element;
-  value: any;
+  renderToolTip: (item: {} | string, format: string) => JSX.Element;
+  value: Array<string | {}>;
 }
 
 export interface InputListState {
-  listItems: string[];
+  listItems: Array<string | {}>;
 }
 
 export default class InputList extends React.Component<InputListProps, InputListState> {
@@ -43,13 +43,27 @@ export default class InputList extends React.Component<InputListProps, InputList
               disabled={this.props.disabled}
               onRemove={() => this.removeListItem(listItem)}
             >
-              {this.props.createEditableInput(setting, {type: "text", description: null, disabled: this.props.disabled, value: value, name: setting.key, label: null, extraContent: this.props.toolTip(listItem, setting.format)})}
+              {this.props.createEditableInput(setting, {
+                type: "text",
+                description: null,
+                disabled: this.props.disabled,
+                value: value,
+                name: setting.key,
+                label: null,
+                extraContent: this.props.renderToolTip(listItem, setting.format)
+              })}
             </WithRemoveButton>
           );
         })}
         <div>
           <span className="add-list-item">
-            { this.props.createEditableInput(setting, {value: null, disabled: this.props.disabled, ref: "addListItem", label: null, description: null})}
+            { this.props.createEditableInput(setting, {
+              value: null,
+              disabled: this.props.disabled,
+              ref: "addListItem",
+              label: null,
+              description: null
+            })}
           </span>
           <button
             type="button"
@@ -62,7 +76,7 @@ export default class InputList extends React.Component<InputListProps, InputList
     );
   }
 
-  removeListItem(listItem: string) {
+  removeListItem(listItem: string | {}) {
     this.setState({ listItems: this.state.listItems.filter(stateListItem => stateListItem !== listItem) });
   }
 
--- a/src/components/InputList.tsx
+++ b/src/components/InputList.tsx
@@ -1,13 +1,15 @@
 import * as React from "react";
 import WithRemoveButton from "./WithRemoveButton";
 import { SettingData } from "../interfaces";
+import ToolTip from "./ToolTip";
+import { LocatorIcon } from "@nypl/dgx-svg-icons";
 
 export interface InputListProps {
   createEditableInput: (setting: SettingData, customProps?: any, children?: JSX.Element[]) => JSX.Element;
   labelAndDescription: (SettingData) => JSX.Element[];
   setting: SettingData;
   disabled: boolean;
-  renderToolTip: (item: {} | string, format: string) => JSX.Element;
+  // renderToolTip: (item: {} | string, format: string) => JSX.Element;
   value: Array<string | {}>;
 }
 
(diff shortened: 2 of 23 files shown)


aslagle:
How do you get the info for the tooltip? Does that only work after you save?

adriana-alter:
The server is looking it up via [an external integration](https://uszipcode.readthedocs.io/index.html).  As it stands, it only works if you save, but we could certainly implement something else (like a "check this zipcode" button) to hit that server code instead, if you think that would be useful...

EdwinGuzman:
Related JIRA ticket https://github.com/NYPL-Simplified/circulation-web/pull/217 (Description is mostly here, not in JIRA).

EdwinGuzman:
Related JIRA ticket https://jira.nypl.org/browse/SIMPLY-1753 (description is here, not in the ticket).

EdwinGuzman:
A couple of things:

* If I try to hover over the instruction's tooltip it gets cut off on the right side even on the widest viewport for a laptop size:
Screen Shot 2019-03-15 at 3 22 58 PM
Screen Shot 2019-03-15 at 3 25 49 PM
* These instructions are very helpful and I don't like that I have to hover over the tooltip to see them. For me, the tooltip is useful for smaller pieces of important information but not as much as what they currently have.
* You can't currently tab on the instruction tooltip to see the instructions.
* You can add an empty area input. When you submit it doesn't actually go through so it's not an actual problem, but for the UI it is allowing you to add an empty value.
* The `X` to remove an input is not vertically aligned.

Screen Shot 2019-03-15 at 3 24 38 PM 
* The tooltip giving you an area for the input after you save is very nice. The CSS makes it seems like it is clickable when you hover over it but it is not (only thing I would change for this).

src/components/EditableInput.tsx
```diff
@@ -60,7 +61,10 @@ export default class EditableInput extends React.Component<EditableInputProps, E
             { this.props.type === "radio" && <span>{this.props.label}</span> }
           </label>
         }
-        {!this.props.label && this.renderElement()}
+        <section className={this.props.extraContent ? "with-add-on" : ""}>
```
EdwinGuzman: I'm not sure I would use a `section` here instead of a `div`. `section`s should be for bigger sectional pieces of information but here it's used just for the input field (for this PR the area input field).

src/components/InputList.tsx
```diff
@@ -0,0 +1,82 @@
...
+        { this.props.labelAndDescription(setting) }
+        { this.state.listItems && this.state.listItems.map((listItem) => {
+          let value = typeof(listItem) === "string" ? listItem : Object.keys(listItem)[0];
+          return (
+            <WithRemoveButton
+              key={value}
+              disabled={this.props.disabled}
+              onRemove={() => this.removeListItem(listItem)}
+            >
+              {this.props.createEditableInput(setting, {type: "text", description: null, disabled: this.props.disabled, value: value, name: setting.key, label: null, extraContent: this.props.toolTip(listItem, setting.format)})}
```
EdwinGuzman: Break these long lines down to multiple lines to make them easier to read. The current linter doesn't have any rules for line length but I like the max 80 or 100 character line rule.

src/components/InputList.tsx
```diff
@@ -0,0 +1,82 @@
...
+import WithRemoveButton from "./WithRemoveButton";
+import { SettingData } from "../interfaces";
+
+export interface InputListProps {
+  createEditableInput: (setting: SettingData, customProps?: any, children?: JSX.Element[]) => JSX.Element;
+  labelAndDescription: (SettingData) => JSX.Element[];
+  setting: SettingData;
+  disabled: boolean;
+  toolTip: (item: any, format: string) => JSX.Element;
+  value: any;
```
EdwinGuzman: Is there a better and more explicit value you can use rather than `any`? I haven't looked closely at what `value` is suppose to be but I'm not sure if we want to use `any`.

adriana-alter:
@EdwinGuzman How would you prefer that the instructions be displayed--just as part of the description, or in a collapsible, or in a tooltip-like thing triggered by clicking rather than hovering, or or or...?

I completely agree that 1) the tooltip needs to be accessible and 2) the UI shouldn't let you add empty values.  But both of those issues predate and are conceptually separate from this branch; I think they should have their own tickets, not be part of this.

src/components/ProtocolFormField.tsx
```diff
@@ -2,217 +2,176 @@ import * as React from "react";
...
+import InputList from "./InputList";
+import { LocatorIcon } from "@nypl/dgx-svg-icons";
 import { SettingData } from "../interfaces";
 import { FetchErrorData } from "opds-web-client/lib/interfaces";
 
 export interface ProtocolFormFieldProps {
   setting: SettingData;
   disabled: boolean;
-  value?: string | string[];
+  value?: string | any[];
```
EdwinGuzman: From below, `value` can be either a single string or an array of strings? Why not always make it an array and check if there's only one value to extract rather than have the server return two different types of structures?

EdwinGuzman:
For the instructions, being part of the description makes the most sense to me or a section directly under it. I think you've done that in collapsibles before.

Sounds good for separate issues for the other two concerns. I raised the first one about the accessible tooltip because now there's important information in there and it would make sense to make that information accessible.

src/components/ProtocolFormField.tsx
aslagle: This seems okay to me since it's used for different fields, but I don't know why it would be `any`. It seems just as weird to make the server use an array for something that will only have one value. We could potentially add a type and do something fancy with typescript so there's different interfaces for the array and the single value.

src/components/ProtocolFormField.tsx
```diff
@@ -2,217 +2,178 @@ import * as React from "react";
...
-              >Add</button>
-          </div>
-        </div>
-      );
-    } else if (setting.type === "image") {
-      return (
-        <div>
-          <label>{setting.label}</label>
-          { this.props.value &&
-            <img src={String(this.props.value)} />
```
aslagle: Did this code that renders the image end up somewhere?

src/components/InputList.tsx
```diff
@@ -0,0 +1,82 @@
+import * as React from "react";
+import WithRemoveButton from "./WithRemoveButton";
+import { SettingData } from "../interfaces";
+
+export interface InputListProps {
+  createEditableInput: (setting: SettingData, customProps?: any, children?: JSX.Element[]) => JSX.Element;
+  labelAndDescription: (SettingData) => JSX.Element[];
+  setting: SettingData;
+  disabled: boolean;
+  toolTip: (item: any, format: string) => JSX.Element;
```
aslagle: Why is the tooltip in here?

src/components/InputList.tsx
```diff
@@ -0,0 +1,82 @@
...
+    this.setState({ listItems: (newProps.value || []) });
+  }
+
+  render(): JSX.Element {
+    const setting = this.props.setting;
+    return (
+      <div>
+        { this.props.labelAndDescription(setting) }
+        { this.state.listItems && this.state.listItems.map((listItem) => {
+          let value = typeof(listItem) === "string" ? listItem : Object.keys(listItem)[0];
```
aslagle: I don't think you need this check since you've defined the type of `listItems` to be a string array.

src/components/ProtocolFormField.tsx
```diff
@@ -2,217 +2,178 @@ import * as React from "react";
...
-            <span className="description" dangerouslySetInnerHTML={{__html: setting.description }} />
-          }
-          <ColorPicker
-            value={String(this.props.value || setting.default)}
-            setting={setting}
-            ref="element"
-            />
-        </div>
+        <span className="input-group-addon">
+          <ToolTip trigger={icons[format]} direction="point-right" text={Object.values(item)[0]}/>
```
aslagle: I don't understand what the item is here, but it would probably be better to look up a specific key rather than taking the first value if there are multiple values.

src/components/ProtocolFormField.tsx
adriana-alter: Yes--it's on line 48.

src/components/InputList.tsx
adriana-alter: It's not the actual tooltip; it's just a function called `toolTip` which is passed down as a prop from `ProtocolFormField` and renders the tooltip.  That seemed to make more sense than importing the actual component again here.  (I will change the name to something less confusing, though!) . Or do you think I should just import it again?

src/components/InputList.tsx
adriana-alter: I defined the type wrong; there could be objects as well as strings in there.

src/components/ProtocolFormField.tsx
adriana-alter: There's only ever one value.

src/components/InputList.tsx
adriana-alter: Actually, now that `ProtocolFormField` is no longer using the tooltip, it does make more sense to just import it directly into `InputList`...

src/components/ProtocolFormField.tsx
aslagle: Ah I see it know. Did the label move after the image?

src/components/ProtocolFormField.tsx
aslagle: I haven't looked at the server code creating this but maybe this object isn't the best representation if there's only one thing.

adriana-alter: That is...very possible.  Right now, I'm not seeing a better way to do it, but I'm about to start working on a project involving moving/refactoring a bunch of validation code (including the thing that's generating this object), so I will figure it out then!

src/components/ProtocolFormField.tsx
adriana-alter: The label for the image is being generated during the `createEditableInput` method.  (Or is that not what you meant?)

aslagle: Yeah, I think the label used to be above the image and the editable input with the label is after the image now. Maybe it still looks fine, just wasn’t sure if it was intentional.

adriana-alter: Oh, I see what you mean.  No, it's still rendering in the same order!  But, during the course of checking that, I have just realized that the "No file chosen" message is now being styled as a label; will fix.

PR #218, opened 2019-03-19
Show full names of languages in library edit form

adriana-alter:
Admins who are configuring a library can now choose languages from a dropdown (like in the book edit form) rather than having to enter ISO-639-2 codes.

Screen Shot 2019-03-25 at 3 57 55 PM

### DIFF
--- a/src/components/Autocomplete.tsx
+++ b/src/components/Autocomplete.tsx
@@ -5,7 +5,7 @@ export interface AutocompleteProps {
   autocompleteValues: string[];
   disabled: boolean;
   name: string;
-  label: string;
+  label?: string;
   value?: string;
 }
 
@@ -38,5 +38,9 @@ export default class Autocomplete extends React.Component<AutocompleteProps, voi
 
   getValue() {
     return (this.refs["input"] as any).getValue();
+  }
+
+  clear() {
+    (this.refs["input"] as any).setState({ value: "" });
   }
 }
--- a/src/components/Autocomplete.tsx
+++ b/src/components/Autocomplete.tsx
@@ -5,7 +5,7 @@ export interface AutocompleteProps {
   autocompleteValues: string[];
   disabled: boolean;
   name: string;
-  label?: string;
+  label: string;
   value?: string;
 }
 
--- a/src/components/Autocomplete.tsx
+++ b/src/components/Autocomplete.tsx
@@ -7,6 +7,7 @@ export interface AutocompleteProps {
   name: string;
   label: string;
   value?: string;
+  onChange?: () => void;
 }
 
 export default class Autocomplete extends React.Component<AutocompleteProps, void> {
@@ -23,7 +24,8 @@ export default class Autocomplete extends React.Component<AutocompleteProps, voi
           value={this.props.value}
           ref="input"
           optionalText={false}
-          />
+          onChange={this.props.onChange}
+        />
         <datalist
           id={this.props.name + "-autocomplete-list"}
           >
--- a/src/components/BookDetailsEditor.tsx
+++ b/src/components/BookDetailsEditor.tsx
@@ -7,7 +7,7 @@ import editorAdapter from "../editorAdapter";
 import ButtonForm from "./ButtonForm";
 import BookEditForm from "./BookEditForm";
 import ErrorMessage from "./ErrorMessage";
-import { BookData, RolesData, MediaData } from "../interfaces";
+import { BookData, RolesData, MediaData, LanguagesData } from "../interfaces";
 import { FetchErrorData } from "opds-web-client/lib/interfaces";
 import { State } from "../reducers/index";
 
@@ -15,6 +15,7 @@ export interface BookDetailsEditorStateProps {
   bookData?: BookData;
   roles?: RolesData;
   media?: MediaData;
+  languages?: LanguagesData;
   bookAdminUrl?: string;
   fetchError?: FetchErrorData;
   editError?: FetchErrorData;
@@ -25,6 +26,7 @@ export interface BookDetailsEditorDispatchProps {
   fetchBook: (url: string) => void;
   fetchRoles: () => void;
   fetchMedia: () => void;
+  fetchLanguages: () => void;
   editBook: (url: string, data: FormData | null) => Promise<any>;
 }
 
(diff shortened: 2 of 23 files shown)


adriana-alter:
Goes with https://github.com/NYPL-Simplified/circulation/pull/1211

adriana-alter:
Okay--it uses the autocomplete functionality now!

src/components/BookDetailsEditor.tsx
```diff
@@ -100,9 +98,9 @@ export class BookDetailsEditor extends React.Component<BookDetailsEditorProps, v
             { this.props.bookData.editLink &&
               <BookEditForm
                 {...this.props.bookData}
+                store={this.props.store}
```
aslagle: I'm wondering if there was any particular reason why you changed this? The pattern we've generally followed has been to connect top-level components and pass everything down as props, but as I think I've mentioned before this was based on redux recommendations that are outdated now. I think it would be great to rethink this whole pattern, but I'm not sure how I feel about diverging from it in one case unless we're planning to start switching more generally. I'm also not what's recommended now - like maybe the store should be in context rather than props if it's going to be used more widely.

src/components/InputList.tsx
```diff
@@ -16,7 +20,20 @@ export interface InputListState {
   listItems: Array<string | {}>;
 }
 
+export interface InputListContext {
+  editorStore: Store<State>;
+  csrfToken: string;
```
aslagle: I don't think the CSRF token is needed for this?

aslagle:
This pretty much looks good but I did have one substantial question I'd like to discuss.

src/components/BookDetailsEditor.tsx
adriana-alter: Do you mean, why did I connect the LanguageField?  Now that it's being used in two places (BookEditForm and InputList), I thought it made more sense to have it be the thing that calls `fetchLanguages`, rather than having each of the parent components, which are already responsible for a lot of stuff, get the list of languages and pass it down.  Also, I don't know how likely it is that we will ever want to use the LanguageField in a third place, but just in case we do, it seems like a better idea to make it as self-sufficient (for lack of a better word) and reusable as possible.

I guess the LanguageField could get the store via context rather than props (I think??) but that seems to me like much more of a divergence from our existing patterns.  (Also, according to a tutorial by Dan Abramov, it's better not to use context unless you have to, insofar as, by introducing global variables, you're basically working against React's data flow.)  

Or am I not understanding your question?

aslagle: Yeah, I haven't had time to look at recent React/Redux tutorials and recommended best practices in detail, so I was wondering if you had. I think it would be great to start using a pattern more like this and gradually refactor the old code as well, but I don't think we should use it in this one place and stop there. 

I'm not sure I'd agree about context though I haven't looked at that tutorial. It could be problematic for things that might change and require components to rerender, but that wouldn't happen with the store. It would be another divergence from our existing patterns, but if that is a better way to use redux we might want to start switching that too?

src/components/InputList.tsx
aslagle: You removed it from another line but not here?

src/components/BookDetailsEditor.tsx
adriana-alter: I agree that it would be really nice to eventually update the old code to use this pattern...but, in that case, I think it's worth just going forward and merging this as it stands; it seems counterproductive to change it to use the older pattern if, down the road, we're just going to have to restore it to the current implementation as part of a bigger refactoring project anyway.

I don't think the context thing is a better way to use Redux--just a different way.

aslagle: I was suggesting that we decide now what pattern we want to use going forward, and make this the first example of it. Going forward, any new code can use the new pattern, and we can refactor the rest when we have time or are working on old parts of the code anyway.

adriana-alter: Ah, that makes sense!  Hmm.  I'm pretty sure this pattern is the best for this component, at least...and it's similar to the pattern in the registry interface, for whatever that's worth...

src/components/BookEditForm.tsx
```diff
@@ -1,16 +1,19 @@
 import * as React from "react";
 import EditableInput from "./EditableInput";
-import { BookData, ContributorData, RolesData, MediaData, LanguagesData } from "../interfaces";
+import { BookData, ContributorData, RolesData, MediaData } from "../interfaces";
 import WithRemoveButton from "./WithRemoveButton";
 import Autocomplete from "./Autocomplete";
```
EdwinGuzman: You're not using this reference anymore.

src/components/Autocomplete.tsx
```diff
@@ -5,7 +5,7 @@ export interface AutocompleteProps {
   autocompleteValues: string[];
   disabled: boolean;
   name: string;
-  label: string;
+  label?: string;
```
EdwinGuzman: Why is this now optional? It doesn't seem like  the label prop has become optional in any of its use.

src/components/BookDetailsEditor.tsx
EdwinGuzman: I'm not big on updating to the latest code recommended best practice when the tools aren't updated to the latest. It may work but they're not always intended to work. I'm not sure what Dan Abramov's tutorial has set for the versions of the tools but it could be farther than what we are using.

Aside from that, by passing down the `store` prop down so it can go to the `LanguageField` component now makes that LanguageField component coupled to the app since it has to connect to the store to get its values. It's nice to have a language input field component that is also autocomplete but it's not a dummy component that can just render it's props values, now it has to connect to the store and I'm not sure I like that. I prefer the higher order component to do more and pass down its values. Again, this is without knowing the latest recommendation for this pattern.

EdwinGuzman:
Unless I'm doing something wrong, I do not get the same screen with the search icon in the library's form for the language input (after I save).

It's also confusing how it says that "each value must be ... a language code" but the autocomplete returns full language names (but if I enter `fre` and save it converts to `French`).

Screen Shot 2019-03-22 at 3 13 21 PM

adriana-alter:
Ah; I should have updated the screenshot.  Now that the tooltip is gone, the search icon is no longer a thing.

EdwinGuzman:
If that's the case then I think you also want to update `InputList` again to remove that icon if it's not going to be used. Updated screenshots whenever you make any changes are always helpful.

adriana-alter:
By the way, the description doesn't say "must be a code" anymore; you need the `language-values` server branch.

EdwinGuzman:
Cool, I wasn't aware and would be nice to include next time.

src/components/BookDetailsEditor.tsx
adriana-alter: The tutorial is for the versions we're using.  I do think his point about not gratuitously dragging context into everything makes a lot of sense.  That's a separate issue, though; re: whether to pass the store down as a prop or just pass the list of languages down, I remain undecided.  I do see what you mean about not wanting the LanguageField component coupled to the app; it seems unlikely that we would have occasion to reuse it in another app, but still, on principle...  But on the other hand, having both BookEditForm and InputList make calls to `fetchLanguages` for the sole purpose of passing that list straight down to LanguageField seems...inelegant.  I will try it, though, and we'll see how it strikes us!

adriana-alter:
Okay, I have un-connected (disconnected?) the LanguageField, and moved the `fetchLanguages` calls to BookDetailsEditor (where it was before) and to Libraries (following the pattern that Collections and DiscoveryServices use: extending GenericEditableConfigList and fetching extra data in `componentDidMount`).  Thoughts?

Also, would it be worth using the InputList version of the LanguageField in the book edit form too--i.e. would we want to be able to tag books as having multiple languages?  (I'm thinking of things like translated poetry books, which often have both the original and the translation...)

aslagle:
I don't think multiple languages for a book is supported on the server, so that would have to be done first.

EdwinGuzman:
This is fine and I'm okay with this update compare to the component connecting to the Store directly before. I am concern about the `additionalData` prop that you're passing down just because of the name. If you had more data later on you want to pass down there might be name confusion. But for now this is okay and probably won't be a problem later.

PR #219, opened 2019-03-19
Don't let users submit blank fields in InputList

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1797

### DIFF
--- a/src/components/InputList.tsx
+++ b/src/components/InputList.tsx
@@ -14,14 +14,17 @@ export interface InputListProps {
 
 export interface InputListState {
   listItems: Array<string | {}>;
+  newItem?: string;
 }
 
 export default class InputList extends React.Component<InputListProps, InputListState> {
   constructor(props) {
     super(props);
     this.state = {
-      listItems: (props.value as string[] || [])
+      listItems: (props.value as string[] || []),
+      newItem: ""
     };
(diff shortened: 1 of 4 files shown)


adriana-alter:
That was my first idea, but when I tried it out I thought it made the form look broken in a way that might throw users.  Might still be worth it though.  What do you think?

aslagle:
Hard to say without trying it, but having nothing happen when you click the button could also make the form seem broken.

adriana-alter:
That's a very good point.  I'll try disabling it again!

src/components/InputList.tsx
```diff
@@ -69,7 +73,7 @@ export default class InputList extends React.Component<InputListProps, InputList
           <button
             type="button"
             className="btn btn-default add-list-item"
-            disabled={this.props.disabled}
+            disabled={this.props.disabled || !this.state.newItem.length}
```
EdwinGuzman: This seems fine. When I add one value and then add it, the next input field will have a button that _looks_ enabled but when the input field is empty and you try to click the add button, it doesn't work (which is good).

adriana-alter: It definitely should not look enabled; good catch.  Will fix.

PR #221, opened 2019-03-29
Import panel

adriana-alter:
Goes with the ReusableComponents `refactor-panel` branch


EdwinGuzman:
This seems fine but I'll re-check once the npm package is updated.

EdwinGuzman:
This is weird but why are you force pushing commits? I'll re-check this PR once the reusable component is published to test it directly from npm.

adriana-alter:
For reasons I cannot begin to guess at, git kept insisting that there were unintegrated remote changes even though there definitely weren't, telling me I needed to run `git pull` and then telling me that everything was already up to date when I did it, etc.; force pushed to override that.

EdwinGuzman:
That's odd. PR looks good but there's this color contrast issue:
Screen Shot 2019-04-04 at 9 49 20 AM

I don't think I saw this styling in the reusable components repo. Was it there?

adriana-alter:
Weird; styling re: the blue panel on hover is in the reusable components repo, but it is not supposed to do THAT.  Possibly a bootstrap thing overriding something?  In any event, it's not coming from this repo; needs to be changed in the reusable components one.  I will make a separate ticket for it.  Good to merge this in the meantime...?

PR #223, opened 2019-04-04
Fix overflowing exceptions

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1856; stops THIS from happening: 
Screen Shot 2019-04-04 at 12 14 46 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.1.9",
+  "version": "0.2.0",
   "description": "Web Front-end for Library Simplified Circulation Manager Admin Interface",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.2.0",
+  "version": "0.1.10",
   "description": "Web Front-end for Library Simplified Circulation Manager Admin Interface",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.1.9",
+  "version": "0.2.0",
```
EdwinGuzman: Why is this a minor update instead of a patch update?

adriana-alter: I'm not clear on the difference; what would doing it as a patch update entail?

EdwinGuzman: `major.minor.patch` so updates like this one, a bug fix, would be a patch update. If we introduce a new feature or configuration update, it'd be a minor update. And if we introduce a big feature or breaking changes to a previous stable version, then it'd be major. This didn't have a good pattern before since we went up to `0.0.99` I think, but now we can be better about the semantic versioning.

adriana-alter: So would this be `0.1.91`...?

EdwinGuzman: `0.1.10`

EdwinGuzman:
Looks good. I have some upcoming updates to circ-web and this isn't a bug fix that needs to go up right now. So let's wait to deploy just once.

EdwinGuzman:
Accidentally merged my PR before this one, so fixed conflicts.

PR #224, opened 2019-04-08
Delete collections

EdwinGuzman:
This helps fix [SIMPLY-1846](https://jira.nypl.org/browse/SIMPLY-1846).

This overrides the window prompt when deleting a collection to let the user know that the deletion will occur in the background. It also displays a simple UI for deleted collections that removes the edit or delete buttons.

This goes along with the `delete-collections` branch in circulation and PR https://github.com/NYPL-Simplified/circulation/pull/1220

Examples:
Screen Shot 2019-04-04 at 12 07 51 PM

Screen Shot 2019-04-08 at 12 45 45 PM


src/components/Collections.tsx
```diff
@@ -60,6 +61,31 @@ export class Collections extends GenericEditableConfigList<CollectionsData, Coll
...
+
+  renderLi(item, index): JSX.Element {
+    if (item.marked_for_deletion) {
+      return (
+        <li className="deleted-collection" key={index}>
+          <TrashIcon />
+          <h4>{this.label(item)}</h4>
+          <p>This collection has been marked for deletion and cannot be edited. Deletion
+            of this collection will run in the background and will be removed from this
+            list once completed.
```
adriana-alter: This seems a little unclear (sounds like deletion will be removed from the list); how about something like "Deletion of this collection will run in the background.  The collection will be removed from this list once deletion is complete"?

src/components/Collections.tsx
```diff
@@ -60,6 +61,31 @@ export class Collections extends GenericEditableConfigList<CollectionsData, Coll
...
+        </li>
+      );
+    };
+    return super.renderLi(item, index);
+  }
+
+  async delete(item: CollectionData): Promise<void> {
+    const deleteInfo = "Marking this collection for deletion will make it " +
+      "uneditable until it is fully deleted in the background.";
+    if (window.confirm(`Set "${this.label(item)}" for deletion? ${deleteInfo}`)) {
```
adriana-alter: Maybe add something about "this cannot be undone"?  I'm worried that people will interpret "marking" for deletion as being a state they can toggle, like flagging.

src/components/EditableConfigList.tsx
```diff
@@ -181,7 +160,47 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
+  }
+
+  renderLi(item, index): JSX.Element {
+    let AdditionalContent = this.AdditionalContent || null;
+    return (
+      <li key={index}>
+        <a
+          className="btn btn-default edit-item"
+          href={this.urlBase + "edit/" + item[this.identifierKey]}
+          disabled={item.marked_for_deletion}
```
adriana-alter: Why disable the link instead of just hiding it (which I thought was happening anyway)?

src/components/Collections.tsx
EdwinGuzman: Sure.

src/components/EditableConfigList.tsx
EdwinGuzman: Forgot to remove that.

src/components/Collections.tsx
```diff
@@ -60,6 +61,31 @@ export class Collections extends GenericEditableConfigList<CollectionsData, Coll
...
+          </p>
+        </li>
+      );
+    };
+    return super.renderLi(item, index);
+  }
+
+  async delete(item: CollectionData): Promise<void> {
+    const deleteInfo = "Marking this collection for deletion will make it " +
+      "uneditable until it is fully deleted in the background.";
```
adriana-alter: Why is this broken up with the plus sign; did you mean to interpolate something...?

EdwinGuzman: There's no linting rule for this (in this project) but I like keeping line length under 100, preferably under 80, characters per line. It get's long and unreadable.

adriana-alter: How about the string on its own line (i.e. new line after `const deleteInfo`, in that case?  This just looks to me like something's been left out of the code, which doesn't help with readability...

src/components/EditableConfigList.tsx
```diff
@@ -73,12 +73,24 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
+    }
   }
 
   render(): JSX.Element {
+    const headers = this.getHeaders();
+    const listAllData = !this.props.isFetching && !this.props.editOrCreate &&
+      this.props.data && this.props.data[this.listDataKey];
+    const createANewItem = listAllData &&
+      (!this.limitOne || this.props.data[this.listDataKey].length === 0) &&
+      this.canCreate();
```
adriana-alter: I really like having these as variables, but can the names be something that indicate that they're booleans (e.g. `canCreateANewItem`)?

src/components/Collections.tsx
aslagle: I think you could just say "This collection is currently being deleted". Not sure if the more technical detail is needed, depending on who is using this.

src/components/Collections.tsx
aslagle: Again I think this could be less technical. I think it's important to communicate that the collection deletion won't happen immediately, and that it cannot be undone once it's initiated. Maybe also that books from the collection will immediately stop showing up in the catalog and patrons will lose existing loans/holds, if that's true.

aslagle: Also, with ES6 you can use backticks to write a multi-line string.

src/components/Collections.tsx
adriana-alter: I think it's still helpful to tell/reassure people that the collection will disappear once it's deleted.  I agree that telling them that deletion is running in the background is probably unnecessary, though.

src/components/EditableConfigList.tsx
aslagle: Why not use the `canCreate` method for the second one?

EdwinGuzman: What do you mean for the second one? I didn't update any logic, just wanted to make it into variables.

aslagle: I mean why not put this logic inside the `canCreate` method?

src/components/Collections.tsx
EdwinGuzman: The string on its own line is still too long (as it initially was).
The back ticks work but it's not multi-line text. So if it's
```
const deleteInfo = `Marking this collection for deletion will make it 
  uneditable until it is fully deleted in the background.`;
```
will make it render it two lines. I prefer to just use quotes separated by the plus sign.

Will update the text though.

EdwinGuzman: It also picks up the tabs as actual tabs:
Screen Shot 2019-04-08 at 5 32 24 PM

src/components/EditableConfigList.tsx
EdwinGuzman: The method gets overridden in one function with a different logic statement so we may want to keep this logic statement along with any other that get added.

adriana-alter: How about changing the names to something more accurate/descriptive, then?  Or adding comments?  I'm finding it very confusing to have a `canCreate()` method that always just returns true.  (I realize that existed before this branch did; this might not be the time to change it!)

aslagle: I also think it's confusing to have both `canCreate` and `canCreateANewItem`.

EdwinGuzman: Yes, I think at this point I'll undo the change and leave the logic in the html as it was before.

PR #225, opened 2019-04-16
Import button

adriana-alter:
The original purpose of this PR was just to bring in the reusable Button component, but when I pulled in the new webpack/typescript stuff, everything suddenly broke such that this wouldn't build; the most recent commit is the result of me spending all day yesterday trying to figure out how to fix it.  The current implementation is definitely not the "right" one--the tests don't run and there are a lot of error messages in the console--but at least the app builds now, and this produces dramatically fewer error messages than any of the (many!) other things I tried.  The majority of the error messages are related to the components that we're importing from opds-web-client, and I am 99% sure that they will resolve themselves once we've updated that repo.  As for the others: suggestions are welcome.  It's probably easiest to just look at the changes for the most recent commit.

 Example of "this will go away when we update opds-web-client" error message: 
Screen Shot 2019-04-16 at 10 48 27 AM

Example of "I don't know what to do with this" error message:
Screen Shot 2019-04-16 at 10 47 41 AM


EdwinGuzman:
This doesn't sound right to me. You should not have gotten these many errors when importing the package.

A lot of updates are not necessary. You are bringing in `@types` but are still using `typings`. It's one or the other. This also haven't been updated to use webpack 4 so I don't think things should be updated either. I will take a closer look at this but may need to close this.

I don't think this is a case of `"this will go away when we update opds-web-client"` because the opds-web-client has nothing to do with this update since all you brought in was the `reuseable-components` package... so this is confusing to me.

package.json
```diff
@@ -22,10 +22,12 @@
   },
   "dependencies": {
     "@nypl/dgx-svg-icons": "0.3.4",
+    "@types/react": "^15.4.0",
+    "@types/react-dom": "^16.8.4",
     "bootstrap": "^3.3.6",
     "font-awesome": "^4.6.3",
     "isomorphic-fetch": "^2.2.1",
-    "library-simplified-reusable-components": "^1.1.2",
+    "library-simplified-reusable-components": "^1.2.1",
```
EdwinGuzman: I'm confused, where did this version come from?

src/__tests__/actions-test.ts
```diff
@@ -52,7 +52,7 @@ describe("actions", () => {
       const fetchMock = stub().returns(new Promise<any>((resolve, reject) => {
         resolve({ status: 200, text: responseText });
       }));
-      fetch = fetchMock;
+      let fetch = fetchMock;
```
EdwinGuzman: I get that these need to be updated to clean up an error message but I don't think you need to do this now... I only encountered this when updating packages. I'm not sure if you need to do this _now_.

src/components/Autocomplete.tsx
```diff
@@ -10,7 +10,7 @@ export interface AutocompleteProps {
   onChange?: () => void;
 }
 
-export default class Autocomplete extends React.Component<AutocompleteProps, void> {
+export default class Autocomplete extends React.Component<AutocompleteProps, {}> {
```
EdwinGuzman: These changes are also odd because this is what the updated typescript made me change, but typescript wasn't updated in this PR.

src/components/CustomListEditor.tsx
```diff
@@ -200,10 +197,11 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
     }
-    let entriesChanged = false;
-    if (this.props.list && this.props.list.books.length !== this.state.entries.length) {
-      entriesChanged = true;
-    } else {
+    let entriesChanged =
+      this.props.list && this.props.list.books &&
+      (this.props.list.books.length !== this.state.entries.length);
+
+    if (!entriesChanged) {
```
EdwinGuzman: Just double checking, did you check that this update still works as intended?

adriana-alter:
The Uglify changes are to get rid of:
Screen Shot 2019-04-18 at 10 22 01 AM

src/components/CustomListEntriesEditor.tsx
```diff
@@ -221,14 +218,12 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
                               {this.getMediumSVG(this.getMedium(book))}
-                              {this.getCatalogLink(book)}
                               <div className="links">
-                                <a
-                                  href="#"
-                                  onClick={() => { this.delete(book.id); }}
-                                  >Remove from list
-                                    <TrashIcon />
-                                </a>
+                                {this.getCatalogLink(book)}
```
EdwinGuzman: Are you sure you want to move this here? The updated style looks fine but it is not consistent with the search results:
Screen Shot 2019-04-18 at 10 25 05 AM

src/components/CustomLists.tsx
```diff
@@ -103,119 +104,54 @@ export class CustomLists extends React.Component<CustomListsProps, CustomListsSt
...
-            />
-          }
+          { this.makeSidebar() }
+          { this.props.editOrCreate && this.makeEditor(entryCount, listCollections) }
         </div>
       </div>
     );
   }
 
+  makeSidebar(): JSX.Element {
```
EdwinGuzman: `renderSidebar` instead?

src/__tests__/actions-test.ts
adriana-alter: Absolutely right; just reverted.

src/components/CustomListsSidebar.tsx
```diff
@@ -0,0 +1,91 @@
+import * as React from "react";
+import PencilIcon from "./icons/PencilIcon";
+import TrashIcon from "./icons/TrashIcon";
+import { Button } from "library-simplified-reusable-components";
+import { CustomListData, LibraryData } from "../interfaces";
+import EditableInput from "./EditableInput";
+import { Link } from "react-router";
+// import { Link } from "react-router-dom";
```
EdwinGuzman: Remove if not needed.

src/components/CustomListsSidebar.tsx
```diff
@@ -0,0 +1,91 @@
...
+      </div>
+    );
+  }
+
+  renderSortButtons(): JSX.Element {
+    return(
+      <fieldset>
+        <legend className="visuallyHidden">Select list sort type</legend>
+        {
+          [["A-Z", "asc"], ["Z-A", "desc"]].map((order) => {
```
EdwinGuzman: Make this a variable instead.

adriana-alter: Make what a variable?  The array that I'm calling `map` on, or the result of `map`, or...?

src/components/Lanes.tsx
```diff
@@ -117,18 +117,16 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
             <h2>Lane Manager</h2>
             <div>
               <Link
-                className={"create-lane btn btn-default" + (this.state.orderChanged ? " disabled" : "")}
```
EdwinGuzman: This is needed to make it look disabled.

src/components/Lanes.tsx
```diff
@@ -117,18 +117,16 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
...
+                className="btn create-lane"
                 to={this.state.orderChanged ? null : ("/admin/web/lanes/" + this.props.library + "/create")}
-                >
-                  Create Top-Level Lane
-                  <AddIcon />
+              >
+                <span>Create Top-Level Lane <AddIcon /></span>
```
EdwinGuzman: Why did you add extra spans?

src/components/CustomListsSidebar.tsx
EdwinGuzman: Yes, the array.

src/components/Lanes.tsx
adriana-alter: I was originally trying to use the reusable Button here, and its `content` prop takes one component (not a string AND an icon).  But the styling is easier this way anyway (I made the links look like buttons, which I think makes more sense for something that says "create", given that all our other "create" things are buttons).

EdwinGuzman:
The latest update for `reusable-button` PR https://github.com/NYPL-Simplified/reusable-components/pull/9 makes this work now. Tests are running but not all are passing.

This could use more review before making the PR but some are minor details that are easy to miss.

When creating a lane, the save button's text is not readable:
Screen Shot 2019-04-18 at 10 35 22 AM

This looks fine but the transition speed for the text and icon are not the same (not big at all):
Screen Shot 2019-04-18 at 10 16 58 AM

But, to go with the issue above, the hover state for the delete buttons are not consistent:
Screen Shot 2019-04-18 at 10 16 51 AM

The trash can icon on hover isn't the right color:
Screen Shot 2019-04-18 at 10 16 48 AM

And here the text isn't right on hover:
Screen Shot 2019-04-18 at 10 16 39 AM

And then there's the inconsistency issue on the list creator with the search results and items in the current list.

adriana-alter:
Thank you!  Working on fixing the tests right now, but will fix those styling issues afterwards.

EdwinGuzman:
Once everything is ready, add Amy as a reviewer.

adriana-alter:
The transition speed issue is coming from `reusable-components`; will fix there.

EdwinGuzman:
I'm not sure where you are regarding styling updates, but there are a few issues. There is that issue with the scss linter (don't use `!important`). The buttons on the search results list are still off:
Screen Shot 2019-04-19 at 3 48 25 PM

Also some styling issues on the lane creator:
1) Screen Shot 2019-04-19 at 3 57 07 PM
2) Screen Shot 2019-04-19 at 3 57 04 PM



I also think the Lane creator is broken but it could not be because of this branch...

src/components/ContextProvider.tsx
```diff
@@ -18,7 +18,7 @@ export interface ContextProviderProps extends React.Props<any> {
 
 /** Provides a redux store, configuration options, and a function to create URLs
     as context to admin interface pages. */
-export default class ContextProvider extends React.Component<ContextProviderProps, void> {
```
EdwinGuzman: Since there was never really an issue with typescript or webpack when importing the reusable-components package, can you revert these changes? They will be necessary later, but until then `void` was the proper way to set this.

adriana-alter:
I think some of the styling issues (including the one that I resorted to trying to fix via `!important`) are coming from the `opds-web-catalog` stylesheet and should probably be fixed there instead.  I'll try again from here, though!

In what way do you think the Lane creator is broken?  (I can't imagine it's from this branch; dropping in buttons shouldn't have affected the functionality...) . I'm finding the Lane code very hard to work with; I'd like to make a separate branch to refactor it (e.g. break out the sidebar into a separate component like I did for CustomLists).  But if something is broken, I can fix while I'm at it.

EdwinGuzman:
For the reverting to `void` part, I meant in all the places you updated it to `{}`...
Yea, refactoring the Lane component is fine. It'd be better if you refactor big components in separate branches because it can get confusing having to review new change and big refactors like in this list one.

For lanes: editing one lane seems to update the rest of the lanes to the same data. I'm trying to see if I can replicate it on the master branch.

adriana-alter:
Will revert to `void` in the other places!

I just played around with the Lanes page on QA; I don't think it's working right, but it works the same as the version from this branch.  I can't replicate the specific issue that you mentioned on this branch, though; how did you make it happen?

src/components/CustomListEditor.tsx
EdwinGuzman: Hm?

src/components/__tests__/CirculationEvents-test.tsx
```diff
@@ -159,11 +160,11 @@ describe("CirculationEvents", () => {
     });
 
     it("shows download form when download button is clicked", () => {
-      let button = wrapper.find("button")
-        .filterWhere(button => button.text() === "Download CSV");
+      let button = wrapper.find(Button);
       expect(button.length).to.equal(1);
+      expect(button.prop("content")).to.equal("Download CSV");
       expect(wrapper.state("showDownloadForm")).to.equal(false);
-      button.simulate("click");
```
EdwinGuzman: I would like these tests to still have the previous workflow instead of calling the component's method. They still produce the same results but you're interacting with the component rather than calling a method.

EdwinGuzman:
The search results is still not updated. It would look nice if it had the same layout as the list entries.

Screen Shot 2019-04-24 at 12 29 00 PM

EdwinGuzman:
Although I like refactoring code, I think refactoring tasks should not be done with other tasks in the same PR.

In the list page, the behavior changed when creating a new list.
_Previously_
1) You are editing a list
2) Click on "Create New List"
3) A new list editor comes up where the title input field is ready for a new title, there is a "Save list title" link, and the "Save this list" button is disabled.

_Now_
1) You are editing a list
2) Click on "Create New List"
3) A new list editor comes up but the previous list's title appears as the new list's title. The "Save this list" button is not disabled and the "Cancel Changes" button appears. (Clicking on "Save this list" does nothing except give a flash, but users shouldn't be able to click this button until they make an actual change).

src/testHelper.ts
```diff
@@ -1,5 +1,7 @@
 import { jsdom } from "jsdom";
-
+import {} from "chai";
```
EdwinGuzman: Why are these here?...

EdwinGuzman: They are not needed.

EdwinGuzman:
One more thing, the error that showed up regarding UglifyJS still shows up. Adding it did not remove the error when building the source. The documentation also says that it's for webpack 4 so maybe that's why it's not doing anything.

src/components/__tests__/CirculationEvents-test.tsx
adriana-alter: If I shallow-render the wrapper, then `simulate("click")` doesn't work anymore, now that the button is its own separate component.  If I mount the wrapper instead, it loops forever (I think because it's trying to call CirculationEvents' onMount stuff); trying to find a workaround for that now...

EdwinGuzman: Right, I think `simulate("click")` only works on `mount`. Hmm maybe you can stub those calls that are made on mount?

adriana-alter: Tried that.  The problem is that the thing that needs to be stubbed is `fetchAndQueue`, which is a component of the method, not a prop...so I would need to stub it after the wrapper is already mounted (because I would need to stub it via `wrapper.instance()`, so `wrapper` needs to already exist)...but mounting the wrapper is the thing that's causing the problem in the first place...

EdwinGuzman: Alright, I'd worry more about the bug on the List page.

adriana-alter: Got the test to work!  (List bug up next...)

src/components/BookEditForm.tsx
```diff
@@ -33,12 +34,10 @@ export default class BookEditForm extends React.Component<BookEditFormProps, Boo
...
       <form ref="form" onSubmit={this.save.bind(this)} className="edit-form">
         {this.renderForm()}
-        <button
-          className="btn btn-default"
+        <Button
           disabled={this.props.disabled}
-          type="submit">
-          Submit
-        </button>
+          callback={this.save.bind(this)}
```
aslagle: I don't think you need this on both the form and the button. Does the form's still work?

src/components/BookEditForm.tsx
```diff
@@ -118,12 +117,13 @@ export default class BookEditForm extends React.Component<BookEditFormProps, Boo
               ref="addContributorName"
               optionalText={false}
               />
-            <button
+            <Button
               type="button"
-              className="btn btn-default add-contributor"
+              className="add-contributor"
```
aslagle: Looking at `Button` seems like this no longer has the `btn-default` class? I don't remember what that class does so maybe it's fine, but did you mean to get rid of it?

adriana-alter: Yes--the `<Button />` component already has the `btn btn-default` class built in.

aslagle: But it only uses `btn-default` if you don't pass in another class name, and this does.

src/components/CustomListEditor.tsx
```diff
@@ -61,8 +61,8 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     return (
       <div className="custom-list-editor">
         <div className="custom-list-editor-header">
-          <div>
-            <fieldset>
+          <div className="edit-custom-list-title">
```
aslagle: The button looks nice but I definitely would prefer to keep it on the same line as the list name here, to leave more vertical space for the list items and search results. On my laptop I can only see one and a half list items. (Not sure if the class name is what changed that.)

src/components/CustomListEditor.tsx
```diff
@@ -139,11 +137,10 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
                     />&nbsp;
-                  <button
-                    className="btn btn-default"
-                    type="submit">Search
-                      <SearchIcon />
-                  </button>
+                  <Button
+                    content={<span>Search<SearchIcon /></span>}
+                    callback={this.search}
+                  />
```
aslagle: This also needs to be on the same line as the search box for vertical space reasons.

src/components/CustomListEditor.tsx
aslagle: This looks okay to me.

src/components/CustomListEntriesEditor.tsx
aslagle: I agree these should look the same, and everything also seems bigger than the old version...

src/components/CustomListEditor.tsx
```diff
@@ -75,22 +75,20 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
-                href="#"
-                className="cancel-changes"
-                onClick={this.reset}
-                >Cancel changes
-                  <XCloseIcon />
-              </a>
+              <Button
+                className="inverted"
+                callback={this.reset}
+                content={<span>Cancel Changes <XCloseIcon /></span>}
```
aslagle: Maybe it's just me but the x icon looks silly on a button.

src/components/CustomListsSidebar.tsx
```diff
@@ -0,0 +1,97 @@
...
+            />);
+          })
+        }
+      </fieldset>
+    );
+  }
+
+  renderListInfo(list: CustomListData): JSX.Element {
+    return (
+      <li key={list.id} className={this.props.identifier === list.id ? "active" : "" }>
```
aslagle: The `active` class doesn't seem to be working.

src/components/CustomListsSidebar.tsx
```diff
@@ -0,0 +1,97 @@
...
+        <div className="custom-list-info">
+          <p>{ list.name }</p>
+          <p>Books in list: { list.entry_count }</p>
+          <p>ID-{ list.id }</p>
+        </div>
+        <div className = "custom-list-buttons">
+          {
+            this.props.identifier === list.id ?
+              <Button disabled={true} content="Editing" /> :
```
aslagle: This also does not seem to work.

src/components/LaneCustomListsEditor.tsx
```diff
@@ -68,6 +68,7 @@ export default class LaneCustomListsEditor extends React.Component<LaneCustomLis
                             </div>
                             <div className="links">
                               <a
+                                className="btn inverted"
```
aslagle: When you tab to one of these buttons the text disappears.

src/components/Lanes.tsx
```diff
@@ -117,18 +117,16 @@ export class Lanes extends React.Component<LanesProps, LanesState> {
...
               </Link>
               <Link
-                className={"reset-lanes btn" + (this.state.orderChanged ? " disabled" : "")}
+                className={"btn reset-lanes inverted " + (this.state.orderChanged ? "disabled" : "")}
```
aslagle: This button also has disappearing text issues - maybe all inverted buttons?

adriana-alter:
@EdwinGuzman When I'm editing a list and click "create new list," the previous list's title does appear as the new title (will fix!), but I'm not getting the other issues; "save this list" is definitely disabled, and there is no "cancel changes" button.  Any additional steps to reproduce??

EdwinGuzman:
No, I could reproduced with the steps I gave above and nothing else.

src/components/Lanes.tsx
adriana-alter: Yes, the problem is the inverted buttons (good catch!!), but it's coming from the reusable-components repo; will fix there.

src/components/BookEditForm.tsx
adriana-alter: Ah.  Good point.  It doesn't matter in this case, though (I just tried it out both ways, and can't see any difference).

EdwinGuzman: It doesn't matter for this case because the overlook style is different, but that's a good case for better keeping track of classes as props for other components.

src/stylesheets/custom_lists.scss
```diff
@@ -58,7 +61,8 @@
       border: 1px solid $pagecolorlight;
       border-bottom-color: $pagetextcolor;
       &.active {
-        background-color: $pagecolorlight;
+        background-color: darken($pagecolorlight, 5);
```
EdwinGuzman: The second parameter is a percentage and being explicit is better.

src/components/__tests__/CustomListsSidebar-test.tsx
```diff
@@ -0,0 +1,111 @@
...
+    let secondListEdit = secondListButtons.find(Link).at(0);
+    let secondListDelete = secondListButtons.find(Button).at(0);
+    expect(secondListEdit.text()).to.contain("Edit");
+    expect(secondListEdit.prop("to")).to.equal("/admin/web/lists/library_name/edit/2");
+    expect(secondListDelete.text()).to.contain("Delete");
+    secondListDelete.simulate("click");
+    expect(deleteCustomList.callCount).to.equal(2);
+  });
+
+  it("disables the edit button if the list is already being edited", () => {
```
EdwinGuzman: This is fine, but another test where you click the button to go from `edit` to `editing` would be nice.

EdwinGuzman:
I am still seeing focus/active button issues but the new package isn't imported yet so I will test once that is brought in.

The comment about uglify wasn't implemented.

These are very minor but pointing them out:
This wasn't that much better before but it didn't overflow.
Screen Shot 2019-04-26 at 1 17 57 PM

Some separation between the buttons is helpful.
Screen Shot 2019-04-26 at 1 18 00 PM

adriana-alter:
Can't test clicking the button to go from "edit" to "editing"; calling `.simulate("click")` on a Link component throws an error: `Invariant Violation: <Link>s rendered outside of a router context cannot navigate.` . (Unless you shallow render the wrapper rather than mounting it...but in that case, `.simulate("click")` doesn't do anything anyway.) . The version of this test currently on master also sets the props rather than clicking.

I can't reproduce the styling issue where the two lane buttons are squished together, even if I make the browser window really narrow; interesting.  Here's how it looks on my computer:   
Screen Shot 2019-04-26 at 2 48 20 PM

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 ## Changelog
 
+### v0.1.13
+#### Updated
+- Implemented the new Panel customization option from v1.3.2 of `reusable-components`.
```
EdwinGuzman: Since the open PR in the reusable components for the panel won't be merged into 1.3.2, can you publish it to make this easier to test?

adriana-alter: Published!

EdwinGuzman:
There's a weird flash on the buttons in the Lane page when you move some lanes around and then cancel those changes. Not sure where it's coming from or why it only affects some buttons but not all.

adriana-alter:
Not coming from the buttons; will investigate when I do the Lanes refactoring (which I will start as soon as it's no longer blocked by this).

CHANGELOG.md
EdwinGuzman: Great, bring it into this PR...

EdwinGuzman: Also, the description for this changelog statement isn't right.

adriana-alter: What's the problem with the changelog description...?

EdwinGuzman: Never mind, I thought the "customization" was the "display" that wasn't merged but it's the "on enter" callback.

EdwinGuzman:
This seems better now. On the list page, you can no longer scroll through the search results or the items in the existing list. So you cannot get to the "load more" button or see past four items if your window is short.

@aslagle let us know if I missed anything.

adriana-alter:
The scrolling issue is not coming from this branch; I've just discovered that it's also broken on QA.

aslagle:
The scrolling on the list page seems to work for me...

aslagle:
I think the view details/add to list buttons are still too big.

Here's my screen:
Screen Shot 2019-04-29 at 8 42 48 PM

adriana-alter:
@aslagle Re: the buttons being too big--I think the best fix for this is to define a new class (like, `.small-button`) in the button stylesheet in the reusable-components repo, not to override the reusable-components styling here.  (I'm already planning to do a branch in reusable-components to add a different button style; can easily put this in there too.)

EdwinGuzman:
@aslagle that's weird because the scrolling doesn't work for me locally or on our qa/prod server. Browser issue?

EdwinGuzman:
@aslagle I believe the sizing will be addressed separately but I'm going to approve for now. I'm okay with overriding the reusable-component button styling here since that was one use case for the components: pull the pre-defined components in and customize if needed.

aslagle:
Did you figure out the scrolling issue?

adriana-alter:
@aslagle We did not figure out the scrolling issue, but given that it's happening on QA, it's definitely not coming from this branch; will make a ticket and address separately.

adriana-alter:
@aslagle Okay to merge?  Really don't want to carry this into the next sprint :)

PR #226, opened 2019-04-16
Resolve opds errors

adriana-alter:
Quick fix for (some of the) error messages that are coming from `opds-web-client`.  (This will presumably no longer be necessary once `opds-web-client` has been updated, but in the meantime...)


EdwinGuzman:
The update for this has no dependency on the `opds-web-client` being updated and the errors are not coming from it (I didn't introduce any of the changes I'm making on it yet). Just want to make that clear.

I don't think this PR is needed now that the other PR is working and tests are running.

EdwinGuzman:
I'm closing this as it's not needed.

PR #228, opened 2019-04-22
Use new Panel onEnter prop

adriana-alter:
Implementation of https://github.com/NYPL-Simplified/reusable-components/pull/10, which resolves https://jira.nypl.org/browse/SIMPLY-1882

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.1.13
+#### Updated
+- Implemented the new Panel customization option from v1.3.2 of `reusable-components`.
+
 ### v0.1.12
 #### Updated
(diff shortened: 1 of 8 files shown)


EdwinGuzman:
I noticed that the panel now lost the smooth transition when opening and closing and now just opens and closes. Is this intended?

adriana-alter:
The smooth transition was coming from react-bootstrap.  I think it's fine for the panel to just open and close.  (We can certainly consider replicating the react-bootstrap functionality by writing our own animation code--probably pretty easy via keyframes--but that's a conversation for another day!)

EdwinGuzman:
Looks good but you need to actually import the new `1.3.2` version of the reusable-components package.

adriana-alter:
Will do once `1.3.2` is published!

PR #229, opened 2019-04-29
Updated packages, Webpack 4, Typescript 2.7.2, opds-web-client 0.2.0

EdwinGuzman:
Fixes [SIMPLY-1872](https://jira.nypl.org/browse/SIMPLY-1872).

This uses the `opds-web-client` version 0.2.0 which has the webpack, typescript, and other package updates. The syntax changes and unit test updates are very similar to the `opds-web-client` updates. Something that is different is the webpack config. It didn't make sense to have a dev and prod file but for sake of convention, I split it up into dev and prod. I added the uglify plugin not just to remove errors (which Adrianna tried to remove before), but the production build works unlike the plugin which comes build in (I believe it's the terser plugin).

I am seeing a difference for the buttons when I hover over them but it could use the yet to be published 1.3.2 version of the reusable-components. I will wait for that.


adriana-alter:
Running `npm install` gives me this.  Hmm.
Screen Shot 2019-04-29 at 3 32 33 PM

EdwinGuzman:
I have a feeling that's because of the linking from the `reusable-components`. Tests on Travis pass so I'm kind of inclined to think it's a local issue.

adriana-alter:
It wasn't a problem when I ran `npm install` on the updated registry_admin, though, and that's also linked to reusable-components...

EdwinGuzman:
Delete your `node_modules` folder and maybe `package-lock.json` and trying installing again.

adriana-alter:
Have tried that three times.  (Also tried unlinking reusable-components.)

EdwinGuzman:
Did you remove the `typings` folder, too?

adriana-alter:
Yes.

adriana-alter:
Now I am also getting this:
Screen Shot 2019-04-29 at 5 05 02 PM

EdwinGuzman:
Try latest commit. Within the time I made the PR and we re-checked, it seems `@types/react` had a minor version update. The `^` tells it to install any new minor patches so instead of `16.4.7` it installed `16.5.8` or something and that brought in a new package which caused everything to explode. I removed the carets from all the other packages to be safe. Blah.

adriana-alter:
Yes, this fixed it!

src/__tests__/actions-test.ts
```diff
@@ -292,19 +273,18 @@ describe("actions", () => {
...
-        resolve({ status: 200 });
-      }));
-      fetch = fetchMock;
+
+      fetchMock.mock(editBookUrl, "done");
+      let fetchArgs = fetchMock.calls();
 
       await actions.editBook(editBookUrl, formData)(dispatch);
-      expect(dispatch.callCount).to.equal(2);
+      expect(dispatch.callCount).to.equal(3);
```
aslagle: Why did this change? Is it similar to the one in the registry?

src/components/ButtonForm.tsx
```diff
@@ -8,7 +8,7 @@ export interface ButtonFormProps extends React.HTMLProps<ButtonForm> {
 /** Submit button. */
 export default class ButtonForm extends React.Component<ButtonFormProps, any> {
   render(): JSX.Element {
-    let disabledProps = this.props.disabled ? { disabled: true } : {};
+    let disabledProps: any = this.props.disabled ? { disabled: true } : {};
```
aslagle: :(

src/components/ManagePatrons.tsx
```diff
@@ -18,16 +19,16 @@ export interface ManagePatronsContext {
...
   static contextTypes: React.ValidationMap<ManagePatronsContext> = {
-    editorStore: React.PropTypes.object.isRequired,
-    csrfToken: React.PropTypes.string.isRequired,
+    editorStore: PropTypes.object.isRequired,
+    csrfToken: PropTypes.string.isRequired,
   };
 
-  static childContextTypes: React.ValidationMap<void> = {
-    library: React.PropTypes.func
+  static childContextTypes: React.ValidationMap<any> = {
```
aslagle: Does it work to do `React.ValidationMap<{library: () => string}>`?

src/components/ButtonForm.tsx
adriana-alter: The ButtonForm has been replaced altogether in the new `import-button` branch anyway.

package.json
```diff
@@ -10,59 +10,67 @@
...
+    "react-dom": "15.4.0",
+    "react-redux": "5.0.6",
+    "react-router": "3.2.0",
+    "recharts": "1.0.1",
+    "redux": "3.7.2",
+    "redux-thunk": "2.1.0",
+    "request": "2.85.0"
   },
   "devDependencies": {
-    "chai": "^3.5.0",
```
adriana-alter: Pretty sure you also now need to explicitly include [react-addons-test-utils](https://www.npmjs.com/package/react-addons-test-utils) as a devDependency.

src/__tests__/actions-test.ts
EdwinGuzman: Yes, it's similar. The `load` dispatch function gets called in the `postForm` call.

src/components/ManagePatrons.tsx
EdwinGuzman: I'll try that.

package.json
EdwinGuzman: It's there now and was there before.

src/components/InputList.tsx
```diff
@@ -109,7 +109,7 @@ export default class InputList extends React.Component<InputListProps, InputList
     if (typeof(item) === "object") {
       return (
         <span className="input-group-addon">
-          <ToolTip trigger={icons[format]} direction="point-right" text={Object.values(item)[0]}/>
+          <ToolTip trigger={icons[format]} direction="point-right" text={Object.values(item)[0].toString()}/>
```
adriana-alter: Why this change?  In what context would this not already be a string?

src/components/ManagePatrons.tsx
EdwinGuzman: Yeah, it seems to work.

adriana-alter:
I have not played around with this yet (and it's not a big deal), but it looks to me like, for the fetchMock tests, there are cases in which using [fetchMock's built-in methods](http://www.wheresrhys.co.uk/fetch-mock/#api-inspectionlastCall) would be cleaner than declaring and indexing into `fetchArgs`.

EdwinGuzman:
I updated a lot to fetch mock's API but I rather keep a bigger refactor to a different branch/PR.

src/components/InputList.tsx
EdwinGuzman: You get this error when running tests and the build:
```
TS2322: Type '{ trigger: any; direction: string; text: {}; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<ToolTip> & Readonly<{ children?: ReactNode; }> & R...'.
  Type '{ trigger: any; direction: string; text: {}; }' is not assignable to type 'Readonly<ToolTipProps>'.
    Types of property 'text' are incompatible.
      Type '{}' is not assignable to type 'string'.
```

It didn't make sense to me either why/when that wouldn't be a string but I guess it could be empty. So explicitly made it a string since it expects a string type.

adriana-alter: WEIRD.  Okay then!

PR #230, opened 2019-05-03
Reintroducing nightwatch and fixing tests

EdwinGuzman:
This fixes [SIMPLY-1927](https://jira.nypl.org/browse/SIMPLY-1927). I'm breaking this up into two PRs. This one is just to be able to run the tests and not error out (a few tests fail but at least the script doesn't completely break).

Many assertions are commented out but that's because a lot of the DOM has changed and the selectors are not relevant anymore. I renamed `titleContains.js` to `titleContains-old.js` because it seems this is now a native function that does not need to be added but I will double check. The second PR will involve updating any more functions/patterns and fixing and cleaning the tests.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.2.1
```
aslagle: I don't think this needs a version bump since the changes only affect tests and the published package should be the same.

aslagle:
It runs for me 👍

CHANGELOG.md
EdwinGuzman: That's true but I would like to keep track of all the changes that go into master. Perhaps I can just add a note that this version will not be published to npm?

aslagle: Maybe we can just combine with another change. You can merge this and I'll add on my next PR?

EdwinGuzman: Sounds good!

PR #232, opened 2019-05-07
Nightwatch test fixes

EdwinGuzman:
Fixes [SIMPLY-1928](https://jira.nypl.org/browse/SIMPLY-1928).

Introduced [Page Objects](http://nightwatchjs.org/guide/#page-objects) (happy to be using these again but in a different context), refactored a couple of tests accordingly, and added a new test for a failure log in.
There's one test that is commented out in the `tests/browser/redirect.js ` file which fails for me but could be fixed when flask is updated (but still doesn't work for me). If that can be uncommented and tested, that'd be great.

I tried to convert most of the Nightwatch tests to es6 syntax but they use a lot of chaining and `this` in their API and those, according to some docs, can't be easily updated. So, you'll see a lot of `function(...)` calls instead of `=>`.

Other minor fixes, while testing the book page I noticed the complaints tab's button was off so I fixed that, as well as the global search button which was not aligned correctly.
Screen Shot 2019-05-07 at 11 39 32 AM

Screen Shot 2019-05-07 at 12 39 06 PM

(Will eventually need a ticket to tangle all the CSS).


aslagle:
The test "navigate to the first lane and back" fails for me.

aslagle:
But the redirect test passes when I uncomment it.

src/components/ComplaintForm.tsx
```diff
@@ -51,7 +52,7 @@ export default class ComplaintForm extends React.Component<ComplaintFormProps, a
               { complaintTypes.map(type => <option key={type} value={type}>{type}</option>) }
             </EditableInput> &nbsp;
           </fieldset>
-          <input className="btn btn-default" type="submit" value="Submit" disabled={this.props.disabled} />
+          <Button callback={() => {}} disabled={this.props.disabled} />
```
aslagle: What's the empty callback for?

EdwinGuzman: The `Button` component requires a `callback` prop for the `onClick` call. It's default type is `submit` so having an empty function doesn't affect it, the form still submits.

tests/browser/commands/signIn.js
```diff
@@ -1,11 +1,18 @@
...
-    .waitForElementVisible("nav", 10000);
+/**
+ * @method signIn
+ * @param {string} username The login form username.
+ * @param {string} password The login form password.
+ * @api command
+ * @description This custom command uses the `login` page object to
+ *  fill out the form.
+ */
+exports.command = function(username, password) {
```
aslagle: Something like this seems like it should work fine if you change it to an arrow function.

EdwinGuzman:
The "navigate to the first lane and back" works for me. What error do you get?

aslagle:
```Timed out while waiting for element <.lane-books li:first-child a> to be present for 5000 milliseconds. - expected "visible" but got: "not found"```

File: tests/browser/commands/signIn.js (hunk 2 above)
Comment (aslagle) [reply tests/browser/commands/signIn.js]: Possibly you need `->` instead of `=>` to prevent binding `this` if nightwatch needs to set it.

Comment (EdwinGuzman) [reply tests/browser/commands/signIn.js]: I'm not familiar with `->`... and it doesn't seem to work.

Comment (aslagle) [reply tests/browser/commands/signIn.js]: Oops, that seems to be something I was remembering from coffeescript that isn't part of ES6. It's weird that ES6 doesn't have something like that...

**Comment by aslagle** [comment]
It's timing out inside goHome, at `.waitForElementVisible(bookSelector, 5000)`. I think mine just takes too long to load.

**Comment by EdwinGuzman** [comment]
Sounds like it, try increasing the timeout. I upgraded flask but the test still fails. Do I need to update anything else?

**Comment by aslagle** [comment]
What version of flask do you have now? I'm on 1.0.2.

**Comment by EdwinGuzman** [comment]
Same version.

**Comment by aslagle** [comment]
You shouldn't need anything else, but you could try upgrading everything just in case.

**Comment by aslagle** [comment]
I increased the timeout and now "navigate to the first lane and back" passes, but I'm getting this silly error:
```
– navigate to two different book classifications tabs (27.637s)
   Failed [notEqual]: ('Digital Media and Technology in Afterschool Programs, Libraries, and Museums' != 'Digital Media and Technology in Afterschool Programs, Libraries, and Museums') - expected "Digital Media and Technology in Afterschool Programs, Libraries, and Museums" but got: "Digital Media and Technology in Afterschool Programs, Libraries, and Museums"
```

**Comment by EdwinGuzman** [comment]
I tried uninstalling and installing all the requirements but it still doesn't work. I trust your end though.
I think that silly error is fine. We can take a better and closer look at these tests and the output later. Those strings are the same, I'm guessing some encoding issue?

**Comment by aslagle** [comment]
The error is because the strings are supposed to be different - that confused me at first too.

**Comment by EdwinGuzman** [comment]
I see... that's odd:

Screen Shot 2019-05-08 at 3 53 32 PM

I don't imagine these tests would rely heavily on the catalog that is being tested on, but do you think that could be it?

**Comment by aslagle** [comment]
Yeah, that does seem to be it. Looks like it's assuming that the second book in the lane will not be the first book when you click into the full lane, but in my case it is. It would be better if it clicked two different books from the page with the swim lanes, that way they'd be more likely to be different.

**Comment by EdwinGuzman** [comment]
Makes sense. I updated it so that the two books are selected from the first swim lane instead of first going into the lane and selecting a book.

**Comment by aslagle** [comment]
Hmm, the test still fails for me but with a different book, and now I'm also getting an error from the redirect test.

**Comment by EdwinGuzman** [comment]
:sob:

**Comment by EdwinGuzman** [comment]
Which redirect test is it, the admin list page or the admin book page?

**Comment by aslagle** [comment]
Book page - 
✖ redirect
 – attempt to view a book page before signing in (7.578s)
   Testing if element <Element [name=@bookTitleSelector]> contains text: "Participatory Politics". Element could not be located in 1000 ms. - expected "Participatory Politics" but got: "null"
       at Object.attempt to view a book page before signing in (/Users/aslagle/dev/circulation-web/tests/browser/redirect.js:96:15)

**Comment by EdwinGuzman** [comment]
Hmm that seems like a timeout issue?

**Comment by aslagle** [comment]
It does look like that but I'm not sure why it started happening with this change.

The other test is now failing on genres, not book titles, I think because my first two books are in the same genre and it expects them to be different.

**Comment by EdwinGuzman** [comment]
Yes, the test is assuming that... It really isn't the best test or doesn't really help much so I am going to remove it. I will keep the title check because that does make more sense.

**Comment by EdwinGuzman** [comment]
I removed the genre check and just left the title check.

I commented out the redirect test. I can't be sure if it is a timeout issue or not but I cannot get it to run locally so I will just leave it commented out for now.

File: tests/browser/navigate.js [hunk 3]
```diff
@@ -157,81 +210,85 @@ module.exports = {
...
-  //                   var secondGenre = result.value;
-  //                   this.assert.notEqual(firstGenre, secondGenre);
-  //                 });
-  //             });
-  //         });
-  //     });
-  // },
-
-  after: function(browser) {
+  "navigate to two different book classifications tabs": function(browser) {
```
aslagle: This should be renamed since it no longer goes to the classifications tabs.

aslagle:
The redirect test probably needs more investigation but I think it's fine to leave that for a separate branch if you want.

EdwinGuzman:
That sounds like a good plan.

EdwinGuzman:
@aslagle I'm leaving the redirect test for the book page for another branch. Does this need anything else?

adriana-alter:
I am getting errors (even with the redirect test commented out)...

Screen Shot 2019-05-14 at 11 44 34 AM

EdwinGuzman:
You might want to increase the timeout for those tests. I guess it depends on how fast a machine is.

adriana-alter:
I will try increasing the timeout!

adriana-alter:
Increasing timeout does not fix :(

EdwinGuzman:
Firefox is weird. Try just the Chrome browser.

adriana-alter:
Everything is fine on Chrome!  (And preliminary Googling indicates that this is a known compatibility issue between Selenium2 and Firefox.) . If we don't care about Firefox, then I think this is good to go.

EdwinGuzman:
Yeah, I thought I mentioned somewhere that Firefox was buggy.

adriana-alter:
Not that I'm seeing.  But anyway.  We do not care about Firefox...?

EdwinGuzman:
I do, but I don't it's necessary for this PR. I'll update the readme, however.

adriana-alter:
Sounds good--update the readme, and I'll approve the PR.

PR #235, opened 2019-05-20
Lanes edit bug

adriana-alter:


src/components/LaneCustomListsEditor.tsx
```diff
@@ -182,26 +182,29 @@ export default class LaneCustomListsEditor extends React.Component<LaneCustomLis
...
     });
     if (this.props.onUpdate) {
-      this.props.onUpdate(this.props.customListIds || []);
+      this.props.onUpdate(ids);
     }
   }
 
   getCustomListIds(): number[] {
-    return this.state.customListIds;
+    let ids = this.state.customListIds.slice(0).concat(this.props.customListIds || []);
```
aslagle: I think you could remove the custom list ids from the state of this component? It seems redundant now.

EdwinGuzman:
This is very close. When I move between lanes to edit them, I now do see their own individual lists. I think the issue is when you save a lane and then go to a sublane. All of its children (and sub lanes) pick up the same changes made but don't actually save them. It appears as if a parent lane A's changes are all saved for A's children, but when you refresh the page A's children have the correct list. It seems to be updating the list for each lane correctly after saving any parent lane.

src/components/__tests__/LaneCustomListsEditor-test.tsx
```diff
@@ -332,9 +333,11 @@ describe("LaneCustomListsEditor", () => {
       }
     });
 
+    wrapper.setProps({customListIds: onUpdate.args[0][0]});
     expect((wrapper.instance() as LaneCustomListsEditor).getCustomListIds().length).to.equal(2);
     expect(onUpdate.callCount).to.equal(1);
-    (wrapper.instance() as LaneCustomListsEditor).reset();
+    (wrapper.instance() as LaneCustomListsEditor).reset([1]);
+    wrapper.setProps({customListIds: onUpdate.args[1][0]});
```
EdwinGuzman: Just add some comments on why you chose those arguments (i.e. what each array index from `onUpdate` is ).

PR #236, opened 2019-05-22
Specifying the node and npm version in package.json.

EdwinGuzman:


aslagle:
This looks fine if it works. Were you able to test it?

PR #237, opened 2019-05-24
React16

EdwinGuzman:
This fixes [SIMPLE-1964](https://jira.nypl.org/browse/SIMPLY-1964).

Creating a draft first because the tests for `Lanes-tests` need to be updated/fixed but that will be refactored by PR #233 . For now that suite is skipped.

I will soon write documentation on the update from enzyme 2 to 3 because that gave me the most trouble and headaches. There are minor API changes but it affects how we write tests - some are bugs found in Enzyme that haven't fully been fixed and some are just "better" ways to handle DOM lookups.

I'd like to revisit all the tests again but this may not be the PR for that.


src/components/ClassificationsForm.tsx
```diff
@@ -167,7 +167,7 @@ export default class ClassificationsForm extends React.Component<Classifications
             </div>
           }
         />
-        <Button callback={this.submit} content="Save" />
+        <Button callback={this.submit} content="Save" className="classification-save-btn" />
```
adriana-alter: Rationale for adding the className?

EdwinGuzman: For a test, but it seems I fixed the test without the need for this so I'll remove.

adriana-alter:
This looks good to me (except for the new merge conflicts from my Lanes branch--sorry about that!)

EdwinGuzman:
No, that's fine. I was waiting for the conflicts so I can continue updating this branch.

src/components/__tests__/InputList-test.tsx
```diff
@@ -160,20 +162,20 @@ describe("InputList", () => {
   });
 
   it("adds an autocompleted item", () => {
-    let languageSetting = {...setting, ...{format: "language-code"}};
+    let languageSetting = { ...setting, format: "language-code" };
```
EdwinGuzman: This was an unnecessary spread syntax here.

src/components/__tests__/Lanes-test.tsx
```diff
@@ -67,8 +67,9 @@ describe("Lanes", () => {
   };
 
   const getTopLevelLanes = () => {
-    let sidebar = wrapper.find(".lanes-sidebar");
-    let topLevelLanes = sidebar.children(DragDropContext).children(Draggable).children("div");
+    const sidebar = wrapper.find(".lanes-sidebar");
+    // let topLevelLanes = sidebar.children(DragDropContext).children(Draggable).children("div");
+    const topLevelLanes = sidebar.find("ul.droppable").at(0).children();
```
EdwinGuzman: This was the most annoying thing to update because now `.children()` doesn't actually return what you expect.

src/components/__tests__/LanesSidebar-test.tsx
```diff
@@ -285,54 +287,123 @@ describe("LanesSidebar", () => {
...
+    //        top
+    //       /   \
+    //      1     4
+    //     / \
+    //    2   5
+    //   /   / \
+    //  3   6   7
+    //      |
+    //      8
+    /** lane  8 */
```
EdwinGuzman: There was no test for lane 8.

src/components/__tests__/LanesSidebar-test.tsx
```diff
@@ -23,10 +23,10 @@ describe("LanesSidebar", () => {
     custom_list_ids: [2], inherit_parent_restrictions: false
```
EdwinGuzman: This file was the most complicated to update so more eyes on it is appreciated.

EdwinGuzman:
@aslagle @adriana-alter ready for review now.

EdwinGuzman:
Also, I wanted to use the new `ref` API but it's too big for some cases that it could be its own PR (along with other React 16 changes).

src/components/__tests__/LanesSidebar-test.tsx
```diff
@@ -23,10 +23,10 @@ describe("LanesSidebar", () => {
...
   };
   let lanesData: LaneData[];
-
   const getTopLevelLanes = () => {
-    return wrapper.children(DragDropContext).children(Draggable).children("div");
```
adriana-alter: Can you put in a comment saying what this returns now (since we're not used to the counterintuitive new `children()` implementation yet!)?

src/components/__tests__/LanesSidebar-test.tsx
```diff
@@ -49,56 +49,57 @@ describe("LanesSidebar", () => {
   });
   it("renders create top-level lane link", () => {
     let create = wrapper.find(".lanes-sidebar > div").find(Link).at(0);
-    expect(create.hasClass("create-lane")).to.be.true;
+    expect(create.render().hasClass("create-lane")).to.be.true;
```
adriana-alter: What's the advantage of calling `render()` on these?  It looks to me like it would return the same thing.  Just a matter of best practices...?

src/components/__tests__/LanesSidebar-test.tsx
```diff
@@ -49,56 +49,57 @@ describe("LanesSidebar", () => {
...
     topLevelLanes = getTopLevelLanes();
-    topLane1 = topLevelLanes.at(0);
-    topLane2 = topLevelLanes.at(1);
-    subLane1 = topLane1.find(Lane).find("> div");
-    expect(topLane1.hasClass("active")).to.be.true;
-    expect(topLane2.hasClass("active")).to.be.false;
-    expect(subLane1.hasClass("active")).to.be.false;
+    topLane1 = topLevelLanes.at(0).find("div").at(0);
+    topLane2 = topLevelLanes.at(1).find("div").at(0);
+    subLane1 = topLane1.find("li > div");
```
adriana-alter: I think the previous `.find(Lane)` implementation is a lot more readable, since it's not obvious that `li > div` is a sublane; if there are advantages to doing it this way instead (which I'm willing to believe!), how about a comment?

src/components/__tests__/LanesSidebar-test.tsx
```diff
@@ -176,56 +177,57 @@ describe("LanesSidebar", () => {
...
     expect(subLane1.text()).to.contain("Sublane 1");
     expect(subLane1.text()).to.contain("(3)");
     expect(subLane1.length).to.equal(1);
     let subLane1Expand = expectCollapsed(subLane1);
 
     // top lane 2 has no sublanes
-    let topLane2Sublanes = topLane2.find(Lane).find("> div");
-    let topLane2Draggables = topLane2.children(Droppable).children(Draggable).children("div");
+    let topLane2Sublanes = topLane2.find("li > div");
+    let topLane2Draggables = topLane2.find(Droppable).find(Draggable);
```
adriana-alter: I think you could use the `allChildren` and `draggableChildren` helper functions for some of these.

src/components/__tests__/LanesSidebar-test.tsx
```diff
@@ -285,54 +287,123 @@ describe("LanesSidebar", () => {
...
+    lane1Children = allChildren(lane1);
+    // The fourth child is the 8th lane since all lanes are expanded.
+    let sublane8 = lane1Children.at(4);
+    expect(isDraggable(sublane8)).to.be.false;
+    expand(sublane8);
+    expect(sublane8.text()).to.contain("sublane 8");
+    // Sublane 8 doesn't have children.
+    expect(allChildren(sublane8.render()).length).to.equal(0);
+    expect(draggableChildren(sublane8.render()).length).to.equal(0);
   });
```
adriana-alter: I like all the new trees!

src/components/__tests__/ClassificationsForm-test.tsx
```diff
@@ -89,7 +89,7 @@ describe("ClassificationsForm", () => {
 
     it("should not allow you to submit if you didn't select an audience or a fiction classification",
       () => {
-        let button = wrapper.find("button").findWhere(button => button.text() === "Save");
+        let button = wrapper.find("button").findWhere(button => button.text() === "Save").at(0);
```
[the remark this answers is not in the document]
adriana-alter: If it's a `Button`, it should have a className you could find it by instead...

src/components/__tests__/ClassificationsForm-test.tsx
```diff
@@ -337,13 +340,14 @@ describe("ClassificationsForm", () => {
       let button = wrapper.find(WithRemoveButton);
       let onRemove = button.props().onRemove;
       onRemove();
+      wrapper.update();
```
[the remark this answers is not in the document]
adriana-alter: It seems weird that the new version of Enzyme would require us to manually do things that the old one did automatically; maybe this is an indication that there's a different pattern we should be using?

src/components/__tests__/ClassificationsForm-test.tsx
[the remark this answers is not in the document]
EdwinGuzman: You'd still have to do the same thing if it's a React class, even if you search by class name.

src/components/__tests__/ClassificationsForm-test.tsx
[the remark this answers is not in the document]
EdwinGuzman: This is a pattern I saw often. There could be a different one as well.

src/components/__tests__/LanesSidebar-test.tsx
EdwinGuzman: I think I explained why in a different file, but the component isn't fully rendered and `hasClass` won't work by itself, _in most cases_.

src/components/__tests__/LanesSidebar-test.tsx
EdwinGuzman: I agree but `find(Lane)` doesn't always work since the component is already mounted. Finding by class name would be better for a shallow rendering of the component, so I tried to stick to classes and elements rather than React class names.

src/components/__tests__/ClassificationsForm-test.tsx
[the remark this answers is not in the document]
aslagle: This is annoying. Would `hostNodes` be more reliable since we don't have to assume the order of the DOM element and React component?

src/components/__tests__/CustomLists-test.tsx
```diff
@@ -205,6 +205,10 @@ describe("CustomLists", () => {
     expect(descendingButton.prop("checked")).to.equal(false);
 
     descendingButton.find("input").simulate("change");
+    radioButtons = wrapper.find(EditableInput);
```
aslagle: so how come this works without `hostNodes` or `at`? Because it's already a list, and each thing in the list now has two things in it?

src/components/__tests__/ResetAdobeId-test.tsx
```diff
@@ -125,27 +125,29 @@ describe("ResetAdobeId", () => {
...
     it("should enable the submit button when the checkbox is checked", () => {
       const input = wrapper.find(".reset-adobe-id input");
-      const button = wrapper.find(".reset-adobe-id").find(Button);
+      let button = wrapper.find(".reset-adobe-id").find(Button);
 
       expect(wrapper.state().checked).to.equal(false);
 
       input.simulate("change");
-      wrapper.update();
+      // wrapper.update();
```
aslagle: It is pretty weird that `wrapper.update` needed to be added some places but removed others....

EdwinGuzman: Forgot to remove this. No instance method was called directly and the state/prop were not updated directly. So if I understand it correctly, it's not needed for this case and re-finding the DOM is what is needed instead.

src/components/__tests__/ClassificationsForm-test.tsx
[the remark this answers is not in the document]
EdwinGuzman: Correct, it'll be the first or only node retrieved (if there are multiple).

src/components/__tests__/CustomLists-test.tsx
EdwinGuzman: Because you're not trying to find by class name.

src/components/__tests__/LanesSidebar-test.tsx
EdwinGuzman: It works 👌

package.json
```diff
@@ -32,14 +32,14 @@
     "library-simplified-reusable-components": "1.3.7",
     "numeral": "^2.0.6",
     "opds-feed-parser": "0.0.14",
-    "opds-web-client": "0.2.0",
+    "opds-web-client": "0.2.2",
```
aslagle: Would you mind updating this to 0.2.3 so I don't have to make a separate PR just for that? I published it yesterday.

src/components/__tests__/SitewideSettingEditForm-test.tsx
```diff
@@ -173,23 +173,26 @@ describe("SitewideSettingEditForm", () => {
     let fillOutFormFields = () => {
       wrapper.setProps({ item: settingData });
       let input = wrapper.find("input[name='value']");
-      let inputElement = input.get(0);
+      // Fixed message - Cannot add property value, object is not extensible
```
aslagle: I don't understand this comment - is this why you have to use `getDOMNode`?

aslagle:
I still don't totally understand some of the enzyme changes but I think this looks good overall.

package.json
EdwinGuzman: Will do.

src/components/__tests__/SitewideSettingEditForm-test.tsx
EdwinGuzman: Yes, I'll update the comment. Using `get(0)` for input elements has been updated to `getDOMNode()`.

src/components/__tests__/CustomListsSidebar-test.tsx
```diff
@@ -34,7 +34,7 @@ describe("CustomListsSidebar", () => {
   });
 
   it("renders a sidebar with a header and a create button", () => {
-    expect(wrapper.hasClass("custom-lists-sidebar")).to.be.true;
+    expect(wrapper.render().hasClass("custom-lists-sidebar")).to.be.true;
```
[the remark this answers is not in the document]
adriana-alter: That is...annoying.  Any reason not to do it as `expect(wrapper.find(".custom-lists-sidebar").length` instead?

adriana-alter:
This looks good!  (Well, much of the new Enzyme stuff does not look good.  But there's nothing we can do about that.)

package.json
EdwinGuzman: Funny story, using 0.2.3 outputs errors (+ a few more not in the screenshot):
Screen Shot 2019-06-04 at 1 58 23 PM

EdwinGuzman: I'll investigate and see if I can fix it in this PR or if not make another PR.

aslagle: Oh weird. It might have to do with adding a subtitle to BookData, but I thought I made that optional. If it's not easy to fix I'll do a separate PR, but didn't want to conflict with this one.

EdwinGuzman: Oh you know what it is, both the client and parser need to be updated because the `OPDSEntry` interface is retrieved directly from the parser module. It raised other errors but fixed.

aslagle: Oh right! I forgot that was separately imported here.

src/__tests__/editorAdapter-test.ts
```diff
@@ -21,13 +22,13 @@ describe("editorAdapter", () => {
...
       identifiers: [],
       links: [
-        { href: "hide", rel: "http://librarysimplified.org/terms/rel/hide", type: "type", title: "title" },
-        { href: "restore", rel: "http://librarysimplified.org/terms/rel/restore", type: "type", title: "title" },
-        { href: "refresh", rel: "http://librarysimplified.org/terms/rel/refresh", type: "type", title: "title" },
-        { href: "edit", rel: "edit", type: "type", title: "title" },
-        { href: "issues", rel: "issues", type: "type", title: "title" },
-        { href: "change-cover", rel: "http://librarysimplified.org/terms/rel/change_cover", type: "type", title: "title" },
-        { href: "cover", rel: "http://opds-spec.org/image", type: "image/png", title: "title" },
+        { href: "hide", rel: "http://librarysimplified.org/terms/rel/hide", type: "type", title: "title", role: "role" },
```
EdwinGuzman: @aslagle I thought `role` was optional but not having it here threw errors...

src/editorAdapter.ts
```diff
@@ -102,7 +95,7 @@ export default function adapter(data: OPDSEntry): BookData {
     title: data.title,
     authors: authors,
     contributors: data.contributors,
-    subtitle: subtitle,
+    subtitle: data.subtitle,
```
EdwinGuzman: Removed the previous way to retrieve the subtitle since that is now part of the OPDSEntry object.

package.json
```diff
@@ -31,8 +31,8 @@
     "isomorphic-fetch": "^2.2.1",
     "library-simplified-reusable-components": "1.3.7",
     "numeral": "^2.0.6",
-    "opds-feed-parser": "0.0.14",
-    "opds-web-client": "0.2.2",
+    "opds-feed-parser": "0.0.16",
```
EdwinGuzman: The 0.0.15 bump introduced the role attribute in the OPDSLink object.

src/__tests__/editorAdapter-test.ts
aslagle: We might want to fix that in the parser, I think it generally always includes everything but sets some things to undefined.

src/editorAdapter.ts
aslagle: Wow I hadn't realized this was here :)

Thanks for updating it.

PR #238, opened 2019-05-31
Add preview button instead of trying to preview when input changes

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2021


aslagle:
It seems pretty unlikely to me that anyone would type a URL letter by letter - this was meant for copy and paste.

EdwinGuzman:
That's true, but it's still a usability issue.

adriana-alter:
I agree that this is a very unlikely use case...but just on principle, I'd rather not use that as justification for implementing a pattern that (I'm pretty sure) violates accessibility guidelines...

aslagle:
It seems like it should be possible to fix the accessibility issue without slowing down the main use case - though I didn't look at the code, so maybe you did that :)

This is kind of a niche feature that's mainly for open bookshelf, so not a huge deal anyway.

aslagle:
Yeah that doesn't sound too bad.

EdwinGuzman:
The code looks fine but it's not working for me. Whenever I click on the "preview" button the page refreshes (I'm not sure if you intended to change this as well) which is fine but nothing happens. No error messages or successful preview shows up (I'm on the latest for everything) after clicking on "preview".

Expected result (from QA server):
1 - error
Screen Shot 2019-06-03 at 12 44 05 PM
2 - successful
Screen Shot 2019-06-03 at 12 45 33 PM

adriana-alter:
I figured it just didn't work in local; it doesn't work for me in local on `master`.  Does it for you??  (Fixing the page refresh thing meanwhile, though!)

EdwinGuzman:
This works for me on local on `master`. This includes displaying the error message and automatically fetching the image preview when the image is large enough (and the page does not refresh).

adriana-alter:
Weird.  I am now having other problems from pulling in the new version of `master` (stay tuned!), so maybe it's all part of the same thing...

adriana-alter:
A fun example:

Screen Shot 2019-06-06 at 11 01 09 AM

So, here's the catch: there is no `Collapsible-test` file.  There is in fact no Collapsible.  That's the thing we replaced with the reusable Panel.  Months ago.  What gives?  (Is it a GHOST?)

EdwinGuzman:
Delete your `lib` folder. The source was compiled correctly and that Collapsible component's tests passed but when it was removed the compiled code was probably still in the `lib` folder. Running the tests doesn't override that folder and instead just updates the compiled files. So now that you're updating this and the syntax changed for the tests, it's failing. This could be fixed with updating the test script to delete and recreate the `lib` folder.

adriana-alter:
Ghost successfully exorcised; thank you!

adriana-alter:
Note: this commit looks enormous because I pulled in the new version of `master`.  The only actual changes are in `BookCoverEditor` and `BookCoverEditor-test`.

EdwinGuzman:
That's the loveliest portrait of a lion ever.

adriana-alter:
🦁

EdwinGuzman:
Looks fine. I will retest soon.

PR #239, opened 2019-06-07
Updating language when deleting a custom list and minor fixes.

EdwinGuzman:
Fixes [SIMPLE-579](https://jira.nypl.org/browse/SIMPLY-579) and goes along with server branch: https://github.com/NYPL-Simplified/circulation/pull/1245.

I also changed and aligned some buttons which needed style fixes on the custom list and lane pages. One side-issue I fixed is [SIMPLY-2033](https://jira.nypl.org/browse/SIMPLY-2033) which you can see now on the master branch or on the production site. Fixed the dragging around of item issues just by checking that the array wasn't empty.
Screen Shot 2019-06-07 at 2 27 24 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Changelog
 
+### v0.3.1
+#### Fixed
+- Deleting a list now also deletes any lanes which contained only that list. Previously, there was no functionality in place to edit or remove any custom lanes when its custom list was deleted.
+- Minor styling and bug fixes on the custom list page.
+
 ### v0.3.0
 #### Updated
 - Using React 16 and Enzyme 3.9.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### v0.3.2
 #### Fixed
-- Deleting a list now also deletes any lanes which contained only that list. Previously, there was no functionality in place to edit or remove any custom lanes when its custom list was deleted.
+- Deleting a list now also deletes any lanes which contained only that list. Previously, there was no functionality in place to edit or remove any custom lanes when all their custom lists had been deleted.
 - Minor styling and bug fixes on the custom list page.
 
 ### v0.3.1
--- a/src/components/CustomLists.tsx
+++ b/src/components/CustomLists.tsx
@@ -12,6 +12,8 @@ import {
   CollectionsData,
   CollectionData as AdminCollectionData,
   LibraryData,
+  LaneData,
+  LanesData,
 } from "../interfaces";
 import Admin from "../models/Admin";
 import { FetchErrorData, CollectionData } from "opds-web-client/lib/interfaces";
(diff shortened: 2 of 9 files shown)


src/components/CustomListsSidebar.tsx
```diff
@@ -89,7 +89,7 @@ export default class CustomListsSidebar extends React.Component<CustomListsSideb
               <Button
                 callback={() => this.props.deleteCustomList(list)}
                 content={<span>Delete<TrashIcon /></span>}
-                className="right-align small"
+                className="right-align small danger"
```
EdwinGuzman: Made it a danger button to be consistent with other "delete" action buttons.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## Changelog
 
+### v0.3.1
+#### Fixed
+- Deleting a list now also deletes any lanes which contained only that list. Previously, there was no functionality in place to edit or remove any custom lanes when its custom list was deleted.
+- Minor styling and bug fixes on the custom list page.
```
adriana-alter: "...when all of their custom lists had been deleted"?

src/components/CustomLists.tsx
```diff
@@ -238,7 +238,10 @@ export class CustomLists extends React.Component<CustomListsProps, CustomListsSt
   }
 
   async deleteCustomList(list: CustomListData): Promise<void> {
-    if (window.confirm("Delete list \"" + list.name + "\"?")) {
+    const descriptionText = "Deleting this list can affect any custom lanes. If a \
+      lane only contains this list, then that lane will also be deleted.";
+    const confirmationText = `Delete list "${list.name}"? ${descriptionText}`;
```
adriana-alter: Could this message appear only if there are relevant custom lanes, and/or contain the names of those lanes?  Or is that too much of a hassle to be worth it (given that this use case is probably not all that common)?

aslagle: I agree, I think this message should only appear when relevant and ideally should mention the lane(s) that would be deleted. It would probably require passing lane info down from the server along with lists, or separately getting lane info on this page.

CHANGELOG.md
EdwinGuzman: Will update.

src/components/CustomLists.tsx
EdwinGuzman: Yeah, that's more complicated but can work that in.

EdwinGuzman:
Updated, now the prompt window will list down any lanes that will be deleted when that list is deleted.

Deleting a list where lanes are not affected:
Screen Shot 2019-06-10 at 10 12 38 AM

Deleting a list where lanes will be affected:
Screen Shot 2019-06-10 at 10 12 31 AM

src/components/CustomLists.tsx
```diff
@@ -238,12 +242,44 @@ export class CustomLists extends React.Component<CustomListsProps, CustomListsSt
...
       await this.props.deleteCustomList(String(list.id));
       this.props.fetchCustomLists();
     }
   }
 
+  async getDeletedLanes(listId: string | number): Promise<LaneData[]> {
+    let deletedLanes = [];
+    let lanes = this.props.lanes;
+    if (!lanes) {
+      await this.props.fetchLanes();
```
EdwinGuzman: Lanes are not needed on this page until a user decides to delete a list, so this is only called then. There is a bit of a lag the first time the user decides to delete a list, but it only happens once and I think it's more efficient than fetching all the lanes when they are not needed.

adriana-alter:
Don't forget, you were going to make that change in the Changelog description.  Otherwise, this looks very nice!

PR #240, opened 2019-06-14
Fixing list editor bugs for new lists.

EdwinGuzman:
Fixes [SIMPLY-2041](https://jira.nypl.org/browse/SIMPLY-2041).
First bug (also found in ticket):
Steps to reproduce:
* Create a new list `/admin/web/lists/.../create`.
* Give it a title. The "Save this list" button is enabled.
* Search for items. Now the save button is disabled. If you add any searched items to the entries list the save button is still disabled.

The other fix is redirecting to a new list's edit page after saving it. It seems that `successMessage` is the appropriate property instead of `responseBody`. This doesn't seem right though so is the server not returning the right thing? I noticed that it does get the right value but then becomes `null` right away.

### DIFF
--- a/src/components/CustomListEditor.tsx
+++ b/src/components/CustomListEditor.tsx
@@ -243,7 +243,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
       }
     }
 
-    return (titleChanged || entriesChanged || collectionsChanged);
+    return titleChanged || entriesChanged || collectionsChanged;
   }
 
   changeTitle(title: string) {
--- a/src/components/CustomListEditor.tsx
+++ b/src/components/CustomListEditor.tsx
@@ -57,6 +57,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     const listTitle = this.props.list && this.props.list.title ? this.props.list.title : "";
     const nextPageUrl = this.props.list && this.props.list.nextPageUrl;
     const opdsFeedUrl = `${this.props.library}/lists/${listTitle}/crawlable`;
+    const hasChanges = this.hasChanges();
     return (
       <div className="custom-list-editor">
         <div className="custom-list-editor-header">
@@ -77,10 +78,10 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
           <div className="save-or-cancel-list">
             <Button
               callback={this.save}
-              disabled={!this.hasChanges()}
+              disabled={!hasChanges}
               content="Save this list"
             />
-            { this.hasChanges() &&
+            { hasChanges &&
               <Button
                 className="inverted"
                 callback={this.reset}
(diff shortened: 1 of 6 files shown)


src/components/CustomListEditor.tsx
```diff
@@ -57,6 +57,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     const listTitle = this.props.list && this.props.list.title ? this.props.list.title : "";
     const nextPageUrl = this.props.list && this.props.list.nextPageUrl;
     const opdsFeedUrl = `${this.props.library}/lists/${listTitle}/crawlable`;
+    const hasChanges = this.hasChanges();
```
EdwinGuzman: Not related to any bug fixes but better to call this once than twice (and potentially doing multiple for loops).

src/components/CustomListEditor.tsx
```diff
@@ -168,9 +169,11 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
 
   componentWillReceiveProps(nextProps) {
     if (!nextProps.list) {
-      this.setState({ title: "", entries: [], collections: [] });
-    }
-    else if (nextProps.list && (nextProps.listId !== this.props.listId)) {
+      // If there's no list then a new one is being created, but there could already
+      // be a list title in the state.
+      this.setState({ title: this.state.title || "", entries: [], collections: [] });
```
aslagle: Wouldn't this mean that if you switched from editing an old list to creating a new one, you'd keep the title from the old list?

EdwinGuzman: Yes it does... but when you are creating a new list, the title input goes into edit mode so it doesn't pick up the old list's title. This is even when you do add a title or save an empty title (which shouldn't happen but different issue).

aslagle: What's the reason for having the state title here though?

EdwinGuzman: For the title of the current new list. If you create a new list and give it a title and then search for entries, the current list's state title becomes an empty string here.

aslagle: Oh I see. What about changing the if so it checks that the nextProps don't have a list but the current props also don't have one?

aslagle: Also it would be good to get some tests for this stuff.

EdwinGuzman: I think it could be broken apart into a creating component and an updating component because this seems messy right now. But this is not something that I think is crucial right now.

EdwinGuzman: Added some tests.

src/components/CustomListEditor.tsx
```diff
@@ -228,7 +235,8 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
         }
       }
     }
-    return titleChanged || entriesChanged || collectionsChanged;
+
+    return !!(this.state.title && this.state.title !== "") && (titleChanged || entriesChanged || collectionsChanged);
```
aslagle: I think I would separate the empty title check from hasChanges. If I'm reading this correctly, when you delete the title, the reset button will be disabled, so you can't get your old title back.

EdwinGuzman: Yes... that's true. I was thinking more of not letting the user do anything unless they have a title. So if they make it empty then the reset button won't show up and the save button is disabled, but once they add a new title the appropriate buttons show up.
I can update it to make it clearer.

EdwinGuzman: Yes... that's true. I was thinking more of not letting the user do anything unless they have a title. So if they make it empty then the reset button won't show up and the save button is disabled, but once they add a new title the appropriate buttons show up.
I can update it to make it clearer.

aslagle: I agree they should not be able to save without a title. But I think save could use a separate or additional function from hasChanges, so reset could still work.

EdwinGuzman:
Updated and displaying the revert button whenever the title is made empty (but still not allowing to save).

EdwinGuzman:
Yeah, there are services and components that map `responseBody` to `successMessage` when connecting to the redux store.

PR #242, opened 2019-06-25
Import form

adriana-alter:
Implementing the reusable Form component, as per https://jira.nypl.org/browse/SIMPLY-2040

### DIFF
--- a/src/components/BookCoverEditor.tsx
+++ b/src/components/BookCoverEditor.tsx
@@ -82,32 +82,33 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
                 alt="Current book cover"
                 />
             </div>
-            <Form
-              ref="parent-form"
-              className="cover-edit-form"
-              onSubmit={this.save}
-              disableButton={this.props.isFetching || !this.props.preview}
-              buttonContent="Save this cover"
-              errorText={this.props.fetchError && <ErrorMessage error={this.props.fetchError} />}
-              content={[
-                <h3>Change cover:</h3>,
-                <Panel
-                  headerText="Cover Metadata"
-                  openByDefault={true}
-                  content={this.renderCoverForm()}
-                  onEnter={this.save}
-                />,
-                (this.props.rightsStatuses &&
+            <div ref="parent-form" className="cover-edit-form">
+              <h3>Change cover:</h3>
+              <Panel
+                headerText="Cover Metadata"
+                openByDefault={true}
+                content={this.renderCoverForm()}
+                onEnter={this.save}
+              />
+              {
+                this.props.rightsStatuses &&
                   <Panel
                     headerText="Rights"
                     openByDefault={true}
                     onEnter={this.save}
                     content={this.renderRightsForm()}
                   />
-                )]
               }
-            />
-          </div>
+              <Button
+                content="Save this cover"
+                disabled={this.props.isFetching || !this.props.preview}
+                callback={this.save}
+              />
+            </div>
+        { this.props.fetchError &&
+          <ErrorMessage error={this.props.fetchError} />
+        }
+        </div>
         }
       </div>
     );
@@ -127,7 +128,7 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
             <ErrorMessage error={this.props.previewFetchError} />
           }
           content={
-            <fieldset>
+            <fieldset key="cover-image">
               <legend className="visuallyHidden">Cover Image</legend>
               <EditableInput
                 elementType="input"
@@ -206,7 +207,7 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
         className="edit-form"
         withoutButton={true}
         content={
-          <fieldset>
+          <fieldset key="rights">
             <legend className="visuallyHidden">Rights:</legend>
             <EditableInput
               elementType="select"
@@ -244,16 +245,18 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
     this.props.refreshCatalog();
   }
 
-  save(data: FormData) {
+  save() {
+    const data = new (window as any).FormData();
+
     const editUrl = this.props.book && this.props.book.changeCoverLink && this.props.book.changeCoverLink.href;
 
-    const imageForm = (this.refs["image-form"] as any).formRef.current;
+    const imageForm = (this.refs["image-form"] as HTMLFormElement).formRef.current;
     const imageFormData = new (window as any).FormData(imageForm);
     data.append("cover_file", imageFormData.get("cover_file"));
     data.append("cover_url", imageFormData.get("cover_url"));
     data.append("title_position", imageFormData.get("title_position"));
 
-    const rightsForm = (this.refs["rights-form"] as any).formRef.current;
+    const rightsForm = (this.refs["rights-form"] as HTMLFormElement).formRef.current;
     const rightsFormData = new (window as any).FormData(rightsForm);
     data.append("rights_status", rightsFormData.get("rights_status"));
     data.append("rights_explanation", rightsFormData.get("rights_explanation"));
--- a/src/components/BookCoverEditor.tsx
+++ b/src/components/BookCoverEditor.tsx
@@ -82,7 +82,7 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
                 alt="Current book cover"
                 />
             </div>
-            <div ref="parent-form" className="cover-edit-form">
+            <div ref="form-container" className="cover-edit-form">
               <h3>Change cover:</h3>
               <Panel
                 headerText="Cover Metadata"
--- a/src/components/BookCoverEditor.tsx
+++ b/src/components/BookCoverEditor.tsx
@@ -84,7 +84,7 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
             </div>
             <Form
               ref="parent-form"
-              className="cover-edit-form no-border"
+              className="cover-edit-form"
               onSubmit={this.save}
               disableButton={this.props.isFetching || !this.props.preview}
               buttonContent="Save this cover"
@@ -119,7 +119,7 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
         <p>Cover must be at least 600px x 900px and in PNG, JPG, or GIF format.</p>
         <Form
           ref="image-form"
-          className="no-border"
+          className="edit-form"
           onSubmit={this.preview}
           buttonContent="Preview"
           buttonClass="top-align"
@@ -203,7 +203,7 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
       <Form
         ref="rights-form"
         onSubmit={this.save}
-        className="no-border"
+        className="edit-form"
         withoutButton={true}
         content={
           <fieldset>
--- a/src/components/BookCoverEditor.tsx
+++ b/src/components/BookCoverEditor.tsx
@@ -9,7 +9,7 @@ import EditableInput from "./EditableInput";
 import { BookData, RightsStatusData } from "../interfaces";
 import { FetchErrorData } from "opds-web-client/lib/interfaces";
 import { State } from "../reducers/index";
-import { Panel, Button } from "library-simplified-reusable-components";
+import { Panel, Button, Form } from "library-simplified-reusable-components";
 
 export interface BookCoverEditorStateProps {
   bookAdminUrl?: string;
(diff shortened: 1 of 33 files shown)


src/components/ChangePasswordForm.tsx
```diff
@@ -39,60 +38,48 @@ export class ChangePasswordForm extends React.Component<ChangePasswordFormProps,
...
-            callback={this.save}
-            disabled={this.props.isFetching}
-          />
-        </form>
-      </div>
+        <Form
+          title="Change Password"
+          onSubmit={this.save}
+          content={formContent}
+          disableButton={this.props.isFetching}
```
EdwinGuzman: This doesn't work. The "Submit" button isn't disabled and you can submit an empty password field.

adriana-alter: I know; [see ticket](https://jira.nypl.org/browse/SIMPLY-2075).  It's not from this PR.

EdwinGuzman: I also get this error when I submit: `this.messageRef.current.focus is not a function`.

src/components/ComplaintForm.tsx
```diff
@@ -70,7 +74,7 @@ export default class ComplaintForm extends React.Component<ComplaintFormProps, a
       return;
     }
 
-    let data = {
+    data = {
```
EdwinGuzman: I'd name this something else since the param is `data`.

src/components/CustomListEditor.tsx
```diff
@@ -363,8 +367,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     return entryPointsElms;
   };
 
-  search(event) {
-    event.preventDefault();
+  search(data) {
```
EdwinGuzman: You're not using `data` so there's no need to have it here, but the callback function probably errors if you exclude it?

adriana-alter: Exactly.  Didn't seem worth coming up with a workaround, since it's a really uncommon use case, and it's not like having the unnecessary `data` is hurting anything.

src/components/ChangePasswordForm.tsx
adriana-alter: Good catch; will investigate!

src/components/CustomListEditor.tsx
EdwinGuzman: I rather not have unused params...

src/components/LibraryEditForm.tsx
```diff
@@ -43,68 +43,63 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
...
-        />
-      </form>
-    );
+    return (
+      <Form
+        hiddenName="uuid"
+        hiddenValue={this.props.item && this.props.item.uuid}
+        className="edit-form"
+        disableButton={this.props.disabled}
+        onSubmit={this.submit}
```
EdwinGuzman: It's odd that you have to pass in `this.submit` to the `Form` AND the `Panel`. I do get that for the panel it is for the enter button but it's just all the ways of submitting that will get confusing (forms can submit on enter as well, so which one is triggering first? etc).

src/components/ManagePatronsForm.tsx
```diff
@@ -46,35 +46,31 @@ export class ManagePatronsForm extends React.Component<ManagePatronsFormProps, {
...
+              className="form-control"
+              placeholder="Enter the patron's identifier"
+              required={true}
+              error={fetchError}
+            />
+          }
+          errorText={(fetchError && !patronExists) && <ErrorMessage error={fetchError} />}
+          successText={(!fetchError && patronExists) && `Patron found: ${patron.authorization_identifier}`}
+        />
+        { (!fetchError && patronExists) && <PatronInfo patron={patron} /> }
```
EdwinGuzman: No way to add this as a prop to the Form?

adriana-alter: No way to add the `PatronInfo` component, you mean...?  No, it's a separate component; not part of the form.  (There's possibly a case to be made for having it be part of the form, but that's a different, page-structure-related, conversation.)

src/components/ServiceEditForm.tsx
```diff
@@ -523,12 +526,12 @@ export default class ServiceEditForm<T extends ServicesData> extends React.Compo
...
   handleData(data) {
-    return data.append("libraries", JSON.stringify(this.state.libraries));
+    data && data.append("libraries", JSON.stringify(this.state.libraries));
+    return data;
   }
 
-  submit(event) {
-    event.preventDefault();
-    handleSubmit(this);
+  async submit(data: FormData) {
```
EdwinGuzman: You added the type here which is nice but missed it in other places.

adriana-alter: As I recall, adding the type was making TypeScript incorrectly throw errors; as between putting `data as any` everywhere or just leaving out the type, this seemed less bad...

src/components/ManagePatronsForm.tsx
EdwinGuzman: Ah nevermind. I thought it was a message and not a component.

src/components/ServiceEditForm.tsx
EdwinGuzman: Hmm but it doesn't throw errors here so why would it throw elsewhere?

src/components/IndividualAdminEditForm.tsx
```diff
@@ -344,13 +344,13 @@ export default class IndividualAdminEditForm extends React.Component<IndividualA
...
       roles = [{ role: "system" }];
     }
-    data.append("roles", JSON.stringify(roles));
+    data && data.append("roles", JSON.stringify(roles));
     return data;
   }
 
-  submit(event) {
-    event.preventDefault();
-    handleSubmit(this);
```
EdwinGuzman: If you're not using this anymore, can you remove the import? Not just here but in other files too.

src/components/__tests__/BookEditForm-test.tsx
```diff
@@ -302,29 +280,29 @@ describe("BookEditForm", () => {
       />
     );
 
-    let form = wrapper.find("form");
-    form.simulate("submit");
+    let form = wrapper.find(Form);
+    form.prop("onSubmit")(new (window as any).FormData(form.getDOMNode()));
```
EdwinGuzman: You can't simulate the submit instead of calling the component's function?

src/components/__tests__/ChangePasswordForm-test.tsx
```diff
@@ -49,23 +50,24 @@ describe("ChangePasswordForm", () => {
...
-    let success = wrapper.find(Alert);
+    let success = wrapper.find(".alert-success");
     expect(success.length).to.equal(0);
 
     wrapper.setState({ success: true, error: null });
-    success = wrapper.find(Alert);
+    success = wrapper.find(".alert-success");
     expect(success.length).to.equal(1);
-    expect(success.prop("bsStyle")).to.equal("success");
+    console.log(success.text());
```
EdwinGuzman: Extra console.

src/components/__tests__/LibraryEditForm-test.tsx
```diff
@@ -136,9 +135,10 @@ describe("LibraryEditForm", () => {
     });
 
     it("has a save button", () => {
-      let saveButton = wrapper.find(Button).at(1);
+      let form = wrapper.find(Form);
+      let saveButton = form.find(Button).at(1);
       expect(saveButton.text()).to.equal("Submit");
-      expect(saveButton.prop("callback")).to.equal(wrapper.instance().submit);
+      expect(form.prop("onSubmit")).to.equal(wrapper.instance().submit);
```
EdwinGuzman: The form has the on submit function call but doesn't the button have it as well?

src/stylesheets/custom_list_editor.scss
```diff
@@ -66,6 +71,12 @@
       }
     }
 
+    form {
```
EdwinGuzman: Why do you need this reset for?

EdwinGuzman:
Add the Jira ticket in the description and related PR links.

The search form for the catalog has a lot more margin/padding:
Screen Shot 2019-06-26 at 11 34 10 AM

src/components/ChangePasswordForm.tsx
adriana-alter: The messageRef error has been fixed by the latest commit in ReusableComponents.

src/stylesheets/custom_list_editor.scss
adriana-alter: Because the form is inside a panel...so without the reset, we get the panel's margin and padding AND the form's margin and padding...and that is too much margin and padding.  (The border reset is no longer necessary, though; will remove.)

Screen Shot 2019-06-26 at 1 01 26 PM

src/components/__tests__/BookEditForm-test.tsx
adriana-alter: This was the only way I could get it to work (after a lot of hours of trying other things and researching Enzyme quirks).

src/components/__tests__/LibraryEditForm-test.tsx
adriana-alter: In theory, yes.  But...Enzyme.  (I think the problem is that the button is a subcomponent of a subcomponent...)

Screen Shot 2019-06-26 at 1 08 23 PM

src/components/LibraryEditForm.tsx
adriana-alter: Turns out that, the way the page is now structured, the panel no longer needs the submit prop anyway.

src/components/__tests__/BookEditForm-test.tsx
EdwinGuzman: I'd prefer the previous way if possible...

src/components/__tests__/LibraryEditForm-test.tsx
EdwinGuzman: Yeah, I had trouble with Enzyme and getting those subcomponents to render and have the right properties. It's definitely possible but forgot the exact syntax. Although, I didn't try it with functions.

src/stylesheets/custom_list_editor.scss
EdwinGuzman: Is it _only_ for this panel? A reset rule like this seems like it can and should be used more often than in just one location.

src/components/__tests__/BookEditForm-test.tsx
adriana-alter: So would I, but the previous way does not work...

src/stylesheets/custom_list_editor.scss
adriana-alter: This is the only case that's come up so far, but I'll keep an eye out...

src/components/__tests__/LibraryEditForm-test.tsx
adriana-alter: Just spent another hour researching it; there really doesn't seem to be a way to do it with functions...

EdwinGuzman: Thanks for looking again.

src/components/ChangePasswordForm.tsx
EdwinGuzman: Is it possible to make the two input fields the same length? Looks off.

Screen Shot 2019-06-27 at 2 45 20 PM

src/components/BookCoverEditor.tsx
```diff
@@ -216,37 +187,79 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
...
     if (!file && !url && this.props.clearPreview) {
       this.props.clearPreview();
     }
     if ((file || url) && this.props.fetchPreview) {
-      this.props.fetchPreview(this.previewUrl(), formData);
+      this.props.fetchPreview(this.previewUrl(), data);
     }
   }
 
+  renderRightsForm() {
```
EdwinGuzman: This no longer works because the structure is all different on this page. I had a hard time trying to read this and now I can better see the issue in the console:
`validateDOMNesting(...): <form> cannot appear as a descendant of <form>.`

src/components/BookCoverEditor.tsx
```diff
@@ -216,37 +187,79 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
...
-  save () {
+  save(data: FormData) {
     const editUrl = this.props.book && this.props.book.changeCoverLink && this.props.book.changeCoverLink.href;
-    const coverForm = (this.refs["cover"] as any);
-    const formData = new (window as any).FormData(coverForm);
-    const coverFile = (this.refs["cover_file"] as any).getValue();
-    const coverUrl = (this.refs["cover_url"] as any).getValue();
-    const rightsForm = (this.refs["rights"] as any);
+
+    const imageForm = (this.refs["image-form"] as any).formRef.current;
```
EdwinGuzman: Do you still need to do `as any`? The typing for this (`HTMLFormElement ` or maybe the React ref specific one) should be available and you can use that instead of using `any`.

src/components/BookCoverEditor.tsx
```diff
@@ -82,68 +82,33 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
...
-              { this.props.rightsStatuses &&
+            <Form
+              ref="parent-form"
+              className="cover-edit-form"
+              onSubmit={this.save}
+              disableButton={this.props.isFetching || !this.props.preview}
+              buttonContent="Save this cover"
+              errorText={this.props.fetchError && <ErrorMessage error={this.props.fetchError} />}
+              content={[
+                <h3>Change cover:</h3>,
```
EdwinGuzman: I see a lot more `Each child in a list should have a unique "key" prop.` warnings in the console and it's because of these arrays. Each element should have a `key` attribute now whenever you pass an array into the `content` prop.

src/stylesheets/edit_form.scss
```diff
@@ -36,7 +38,7 @@
     }
   }
 
-  button.btn:not(.add-list-item) {
-    margin: 1em 0;
+  > button.btn:not(.add-list-item) {
```
EdwinGuzman: So specific! Try to make more general style rules for reusability.

EdwinGuzman:
It's a big PR so I just caught a few more things. 😵

EdwinGuzman:
You should update the PRs with the updated npm package version if you haven't already.

src/components/BookCoverEditor.tsx
```diff
@@ -82,32 +82,33 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
...
-              content={[
-                <h3>Change cover:</h3>,
-                <Panel
-                  headerText="Cover Metadata"
-                  openByDefault={true}
-                  content={this.renderCoverForm()}
-                  onEnter={this.save}
-                />,
-                (this.props.rightsStatuses &&
+            <div ref="parent-form" className="cover-edit-form">
```
EdwinGuzman: This is no longer a form so the ref could use a different name.

src/components/BookCoverEditor.tsx
```diff
@@ -82,7 +82,7 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
                 alt="Current book cover"
                 />
             </div>
-            <div ref="parent-form" className="cover-edit-form">
+            <div ref="form-container" className="cover-edit-form">
```
EdwinGuzman: Ah, I meant more like this is no longer a `<form>` so you can remove that from its name. But this is fine.

PR #244, opened 2019-07-03
Complaints button

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2074.  "Resolve" button in Complaints tab should be aligned like so:
Screen Shot 2019-07-03 at 3 38 59 PM

Not like so:
Screen Shot 2019-06-20 at 12 36 47 PM


src/components/Complaints.tsx
```diff
@@ -81,7 +81,7 @@ export class Complaints extends React.Component<ComplaintsProps, {}> {
                   <td className="complaint-count">{this.props.complaints[type]}</td>
                   <td className="complaint-resolve">
                     <Button
-                      className="btn-sm"
+                      className="btn-sm top-align"
```
EdwinGuzman: This is one reason why I didn't like having specific classes for all the alignments. It's easy to miss some.

EdwinGuzman:
This is fine, but can you hold off on merging one or two minor style fixes and making that a new version? It's better to bundle more fixes/updates together.

adriana-alter:
Yes; that's why the other one branches off of this.  I figured we'd add a few more things onto this version.

PR #245, opened 2019-07-05
Complaint names

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2101; renders names of complaint types in a more user-friendly format.

Before:
Screen Shot 2019-07-03 at 3 08 23 PM
Screen Shot 2019-07-03 at 3 38 59 PM

<hr></hr>

After:
Screen Shot 2019-07-05 at 12 04 10 PM
Screen Shot 2019-07-05 at 11 52 17 AM

### DIFF
--- a/src/components/ComplaintForm.tsx
+++ b/src/components/ComplaintForm.tsx
@@ -2,6 +2,7 @@ import * as React from "react";
 import EditableInput from "./EditableInput";
 import { PostComplaint } from "../interfaces";
 import { Button, Form } from "library-simplified-reusable-components";
+import { formatString } from "./sharedFunctions";
 
 export interface ComplaintFormProps {
   disabled?: boolean;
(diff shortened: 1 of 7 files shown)


src/components/ComplaintForm.tsx
```diff
@@ -34,6 +34,12 @@ export default class ComplaintForm extends React.Component<ComplaintFormProps, a
       "wrong-genre"
     ];
 
+    let formatType = (x) => {
```
EdwinGuzman: Use `const`. A better variable name other than `x` would be better, and its type, too. This seems like a good function that could be shared between components so maybe it can live in a general utils file?

src/components/Complaints.tsx
```diff
@@ -124,7 +124,7 @@ export class Complaints extends React.Component<ComplaintsProps, {}> {
   readableComplaintType(type: string) {
     let match = type.match(/\/terms\/problem\/(.+)$/);
     if (match) {
-      return match[1].replace("-", " ");
+      return match[1].substr(0, 1).toUpperCase() + match[1].substr(1).replace(/-/g, " ");
```
EdwinGuzman: This looks like a place you can use the same function above.

PR #247, opened 2019-07-11
Updating Redux and Typescript for updated compiler.

EdwinGuzman:
Resolves [https://jira.nypl.org/browse/SIMPLY-2111](SIMPLY-2111). Similar to [opds-web-client#232](https://github.com/NYPL-Simplified/opds-web-client/pull/232/).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.3.6",
+  "version": "0.3.7",
   "description": "Web Front-end for Library Simplified Circulation Manager Admin Interface",
   "repository": {
     "type": "git",
@@ -40,11 +40,11 @@
     "react-bootstrap": "^0.32.4",
     "react-color": "^2.14.1",
     "react-dom": "^16.8.6",
-    "react-redux": "^5.0.6",
+    "react-redux": "^7.1.0",
     "react-router": "^3.2.0",
     "recharts": "^1.0.1",
-    "redux": "^3.7.2",
-    "redux-thunk": "^2.1.0",
+    "redux": "^4.0.1",
+    "redux-thunk": "^2.3.0",
     "request": "^2.85.0"
   },
   "devDependencies": {
@@ -73,7 +73,7 @@
     "terser-webpack-plugin": "^1.3.0",
     "ts-loader": "^4.3.1",
     "tslint": "^3.10.2",
-    "typescript": "2.7.2",
+    "typescript": "2.9.1",
     "uglifyjs-webpack-plugin": "^2.1.2",
     "url-loader": "^0.5.7",
     "webpack": "^4.29.6",
--- a/src/components/BookDetailsTabContainer.tsx
+++ b/src/components/BookDetailsTabContainer.tsx
@@ -137,7 +137,7 @@ function mapDispatchToProps(dispatch) {
   };
 }
 
-let connectOptions = { withRef: true, pure: true };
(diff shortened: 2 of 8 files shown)


src/components/BookDetailsTabContainer.tsx
```diff
@@ -137,7 +137,7 @@ function mapDispatchToProps(dispatch) {
   };
 }
 
-let connectOptions = { withRef: true, pure: true };
+let connectOptions = { pure: true };
```
EdwinGuzman: This also had to be removed here.

package.json
```diff
@@ -73,7 +73,7 @@
     "terser-webpack-plugin": "^1.3.0",
     "ts-loader": "^4.3.1",
     "tslint": "^3.10.2",
-    "typescript": "2.7.2",
+    "typescript": "2.9.1",
```
EdwinGuzman: I'm not sure why you didn't encounter this but when I updated the same packages you did, I had to update the compiler because it didn't understand the redux typings. Even if I added `@types/redux` or `@types/react-redux`, it still didn't work. The solution was to update Typescript.

src/store.ts
```diff
@@ -1,16 +1,16 @@
...
-let reducers = combineReducers({
+let reducers: Reducer = combineReducers({
   editor: editorReducers,
   catalog: catalogReducers
 });
 
 /** Build a redux store with reducers specific to the admin interface
     as well as reducers from opds-web-client. */
-export default function buildStore(initialState?: any) {
+export default function buildStore(initialState?: any): Store<any> {
```
EdwinGuzman: This is the most annoying thing because I didn't want to use `any`. I created an interface for the combined state of the editor store and catalog store but then noticed that it doesn't work because for some `buildStore` calls, it expects _only_ the editor store. So `any` is the best for now unless if we want to further break it down or change the architecture in some tests and components (specifically if they use the context's `editorStore` property).

package.json
aslagle: Weird, but it's good to update this anyway.

src/store.ts
aslagle: Could it use a union type? I'm not sure exactly what the syntax is, but like `Store<EditorStoreType | CombinedType>`?

aslagle: Actually maybe if you made it generic - 
```buildStore<T>(initialState: T): Store<T>```

Is the initial state the same type as the store state?

EdwinGuzman: I'll try the generic. The initial state is the type as the admin's store state (not the catalog's store state). But even when specifying it, the compiler complains about a few properties not being defined in the store state.

aslagle: I don't think it would work unless those types are the same for any given call.

EdwinGuzman: No, the generic doesn't work...

aslagle: This seems to work for me: `buildStore(initialState?: State): Store<State>`, with `State` imported from reducers/index. Am I missing something?

EdwinGuzman: No, you're right. It works. I remembering trying that and it failed but I might've changed something else since then.

aslagle: I deleted my node modules to check this and now nothing works :(

EdwinGuzman: Try it again? I just re-installed everything and it all looks okay.

aslagle: I got it now - I didn't realize npm unlink removes things from package.json unless you tell it not to.

PR #248, opened 2019-07-11
Implement basic WYSIWYG editor in book summary form

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2112

Goes with https://github.com/NYPL-Simplified/reusable-components/tree/mousedown

### DIFF
--- a/src/components/BookEditForm.tsx
+++ b/src/components/BookEditForm.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
 import EditableInput from "./EditableInput";
+import EditorField from "./EditorField";
 import { Button, Form } from "library-simplified-reusable-components";
 import { BookData, ContributorData, RolesData, MediaData, LanguagesData } from "../interfaces";
 import WithRemoveButton from "./WithRemoveButton";
@@ -206,14 +207,10 @@ export default class BookEditForm extends React.Component<BookEditFormProps, Boo
           value={this.props.rating && String(Math.round(this.props.rating))}
           optionalText={false}
           />
-        <EditableInput
-          elementType="textarea"
-          disabled={this.props.disabled}
-          name="summary"
-          label="Summary"
-          value={this.props.summary}
-          optionalText={false}
-          />
+        <div className="editor form-group">
+          <label className="control-label">Summary</label>
+          <EditorField ref="summary" summary={this.props.summary} disabled={this.props.disabled}/>
+        </div>
       </fieldset>
     );
   }
@@ -252,6 +249,8 @@ export default class BookEditForm extends React.Component<BookEditFormProps, Boo
   }
 
   save(data: FormData) {
+    const summary = (this.refs["summary"] as any).getValue();
+    data.append("summary", summary);
     this.props.editBook(this.props.editLink.href, data).then(response => {
       this.props.refresh();
     });
--- a/src/components/EditorField.tsx
+++ b/src/components/EditorField.tsx
@@ -1,0 +1,82 @@
+import * as React from "react";
+import { Editor, EditorState, ContentState, RichUtils, convertFromHTML, compositeDecorator } from "draft-js";
+import { convertToHTML } from "draft-convert";
+import { Button } from "library-simplified-reusable-components";
+
+interface EditorFieldState {
+  editorState: EditorState;
+}
+
+interface EditorFieldProps {
+  summary: string;
+  disabled: boolean;
+}
+
+export default class EditorField extends React.Component<EditorFieldProps, EditorFieldState> {
+  constructor(props) {
+    super(props);
+    let blocksFromHTML;
+    try {
+      blocksFromHTML = convertFromHTML(JSON.parse(props.summary));
+    }
+    catch {
+      blocksFromHTML = convertFromHTML(props.summary);
+    }
+    const state = ContentState.createFromBlockArray(
+      blocksFromHTML.contentBlocks,
+      blocksFromHTML.entityMap,
+    );
+    this.state = {editorState: EditorState.createWithContent(state, compositeDecorator)};
+
+    this.onChange = this.onChange.bind(this);
+    this.handleKeyCommand = this.handleKeyCommand.bind(this);
+    this.getValue = this.getValue.bind(this);
+    this.changeStyle = this.changeStyle.bind(this);
+    this.renderButton = this.renderButton.bind(this);
+  }
+
+  onChange(editorState) {
+    this.setState({ editorState });
+  }
+
+  handleKeyCommand(command, editorState) {
+   const newState = RichUtils.handleKeyCommand(editorState, command);
+   if (newState) {
+     this.onChange(newState);
+     return "handled";
+   }
+   return "not-handled";
+  }
+
+  renderButton(style: string): JSX.Element {
+    let content = React.createElement(style[0].toLowerCase(), null, style[0]);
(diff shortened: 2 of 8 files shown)


src/components/BookEditForm.tsx
```diff
@@ -252,6 +249,8 @@ export default class BookEditForm extends React.Component<BookEditFormProps, Boo
   }
 
   save(data: FormData) {
+    const summary = (this.refs["summary"] as any).getValue();
```
EdwinGuzman: Can you use the updated way to use refs?

src/components/EditorField.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { Editor, EditorState, ContentState, RichUtils, convertFromHTML, compositeDecorator } from "draft-js";
+import { convertToHTML } from "draft-convert";
+import { Button } from "library-simplified-reusable-components";
+
+interface EditorFieldState {
+  editorState: EditorState;
+}
+
+interface EditorFieldProps {
+  summary: string;
```
EdwinGuzman: `content` is a better variable name for this so you can use this for a different textarea field that is not the summary.

src/components/EditorField.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export default class EditorField extends React.Component<EditorFieldProps, EditorFieldState> {
+  constructor(props) {
+    super(props);
+    let blocksFromHTML;
+    try {
+      blocksFromHTML = convertFromHTML(JSON.parse(props.summary));
+    }
+    catch {
```
EdwinGuzman: These, like the if/else, should be on the same line as the closing brace.

src/components/EditorField.tsx
```diff
@@ -0,0 +1,82 @@
...
+      blocksFromHTML = convertFromHTML(props.summary);
+    }
+    const state = ContentState.createFromBlockArray(
```
EdwinGuzman: Maybe `contentState` to not be confused with `this.state`.

src/components/EditorField.tsx
```diff
@@ -0,0 +1,82 @@
...
+      />
+    );
+  }
+
+  changeStyle(e, style: string) {
+    e.preventDefault();
+    this.onChange(RichUtils.toggleInlineStyle(this.state.editorState, style));
+  }
+
+  getValue() {
```
EdwinGuzman: Where are you using this function?

src/components/EditorField.tsx
```diff
@@ -0,0 +1,82 @@
...
+    return convertToHTML(this.state.editorState.getCurrentContent());
+  }
+
+  render() {
+    return (
+      <div className="editor-field">
+        { ["BOLD", "ITALIC", "UNDERLINE"].map(style => this.renderButton(style)) }
```
EdwinGuzman: If we ever need to add more styles, this inline array will be too long. Make the array a variable instead.

EdwinGuzman:
This is cool. I haven't used draftjs in a few years so I don't know the latest API but I trust you are using the correct ones. Maybe comments would be helpful (in the form of
```
/**
 *
 */
``` 
for functions).

A couple of things:
* The buttons should be a list of buttons.
* The full name instead of just the first letter is better for UX.
* I expect to write, click on a button to add a style, and then keep on typing. Something is wrong and this does not happen. I checked their demo and it works fine there so it should behave the same way here. It seems it only works when text is highlighted and the style added.
* Check with Willa to see if any more accessibility updates need to be added.

src/components/EditorField.tsx
adriana-alter: `BookEditForm` (line 254) calls it to get the editor field's value when submitting the form.

src/components/BookEditForm.tsx
```diff
@@ -21,6 +21,8 @@ export interface BookEditFormState {
 
 /** Edit a book's metadata in the edit tab on the book details page. */
 export default class BookEditForm extends React.Component<BookEditFormProps, BookEditFormState> {
+  private summaryRef = React.createRef<EditorField>();
```
EdwinGuzman: I'd change the name of this ref to not include "summary" so it can be more general.

adriana-alter: That seems confusing, given that the BookEditForm is in fact only using the EditorField for the summary.  (It's not internal to the EditorField; other parent components that are using it for non-summary things can call it whatever they want.)

EdwinGuzman: Oh right, I was thinking this was still the other component.

EdwinGuzman:
I'm getting inconsistencies when I click on one, two, or all buttons. I click on one to select it and then select another one and only the second style is used. Sometimes when you click to select and click to deselect, the style does not really become unselected. Sometimes clicking on all three styles works but other times it does not.

I also suggest to have a different hover color and active color. When I click on a button to deselect I don't get the feedback that it is off because the cursor is still on the button. So it appears as it is still selected.

src/components/EditorField.tsx
```diff
@@ -25,7 +24,7 @@ export default class EditorField extends React.Component<EditorFieldProps, Edito
       blocksFromHTML.contentBlocks,
       blocksFromHTML.entityMap,
     );
-    this.state = {editorState: EditorState.createWithContent(contentState, compositeDecorator)};
+    this.state = { editorState: EditorState.createWithContent(contentState, compositeDecorator) };
```
EdwinGuzman: 👍

src/components/EditorField.tsx
```diff
@@ -51,23 +50,19 @@ export default class EditorField extends React.Component<EditorFieldProps, Edito
     let buttonContent = React.createElement(style[0].toLowerCase(), null, style);
     let isActive = this.state.editorState.getSelection().getHasFocus() && this.state.editorState.getCurrentInlineStyle().has(style.toUpperCase());
     return (
-      <Button
+      <button
```
EdwinGuzman: This is tricky because now we can't use the reusable component. I think this is a good use case for updating the `Button` component to take in an `onMouseDown` prop.

src/stylesheets/editor.scss
```diff
@@ -28,10 +28,14 @@
       color: #403D37;
       font-weight: unset;
       transition: color 0, background 0;
-      &:hover, &.active {
-        background: #403D37;
-        color: white;
-      }
+    }
+    button.btn.inline:hover {
```
EdwinGuzman: Could you write this in a more scss way and have it like:
```
button {
  .btn {
    .inline {
      &:hover
// ...etc
```

Maybe not with all the deep nested levels but with some nesting in place.

src/stylesheets/editor.scss
```diff
@@ -28,10 +28,14 @@
...
+      background: $blue;
+      color: white;
+    }
+    button.btn.inline.active {
+      background: #403D37;
```
EdwinGuzman: Can you reuse an existing color for the active state? Or make it a variable instead (same with the above usage).

EdwinGuzman:
Working much better but minor comments. Updating the `Button` component and bringing it in here would be great. I'll dig into the docs for whatever new package is brought in for the next PR to help out; it seemed there was a hard time getting the state to work correctly for the added styles.

EdwinGuzman:
Let’s release this with the linter update together. You can merge your changes first and I’ll update my PR accordingly.

PR #250, opened 2019-07-18
tslint-react-a11y linter

EdwinGuzman:
Resolves [SIMPLY-2131](https://jira.nypl.org/browse/SIMPLY-2131). Check the ticket for the list of issues that the linter initially picked up and that were fixed in this PR.

Changes:
* Updated alignment to buttons that needed it
* Updating size of the button in the genre form:
Screen Shot 2019-07-17 at 3 58 01 PM
* added `role="button"` to anchors
* adding `aria-selected` to `option` elements in the `select` dropdown

For the third bullet point, it's more of a functional/stylistic change to update anchors to buttons. Since this was just to add the linter, I left updating those elements out of this since unless if it was simple. Eventually, those elements should be updated.
For the most part, the last bullet point worked okay but for some, I added `aria-selected={false}` on purpose. The current value is hard to get in some forms and updating those requires a lot more work. For some forms it's hard to get the current value because that is not stored in the state; current values are fetched after submitting. Maybe that is not the right pattern.


src/components/BookCoverEditor.tsx
```diff
@@ -100,6 +100,7 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
                   />
               }
               <Button
+                className="left-align"
```
adriana-alter: Why?  The button is already aligned correctly.

src/components/BookCoverEditor.tsx
```diff
@@ -115,6 +116,8 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
   }
 
   renderCoverForm() {
+    let titlePositionRef = this.refs["title_position"] as any;
+    let titlePositionValue = titlePositionRef && titlePositionRef.getValue();
```
adriana-alter: Nice :)

src/components/BookCoverEditor.tsx
```diff
@@ -200,6 +204,10 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
   }
 
   renderRightsForm() {
+    const copyrightUri = "http://librarysimplified.org/terms/rights-status/in-copyright";
+    const otherUri = "http://librarysimplified.org/terms/rights-status/unknown";
```
adriana-alter: How about a more descriptive variable name?  `unknownStatusUri` or something?

src/components/BookDetailsEditor.tsx
```diff
@@ -92,6 +92,7 @@ export class BookDetailsEditor extends React.Component<BookDetailsEditorProps, {
               <div className="form-group form-inline">
                 { this.props.bookData.hideLink &&
                   <Button
+                    className="left-align"
```
adriana-alter: Again: but it's aligned correctly as it is...

src/components/CustomListEditor.tsx
```diff
@@ -139,7 +139,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
                           checked={this.hasCollection(collection)}
                           label={collection.name}
                           value={String(collection.id)}
-                          onChange={() => {this.changeCollection(collection);}}
+                          onChange={() => { this.changeCollection(collection); }}
```
adriana-alter: Why?  Just the linter's preference, or is the functionality different?

src/components/ErrorMessage.tsx
```diff
@@ -55,7 +55,7 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, {}>
             }
             <span dangerouslySetInnerHTML={{ __html: errorMessageText }} /><br />
             { this.props.tryAgain &&
-              <a onClick={this.tryAgain.bind(this)}>Try again</a>
+              <a role="button" onClick={this.tryAgain.bind(this)}>Try again</a>
```
adriana-alter: I realize this isn't coming from this branch, but: what's the point of defining a `tryAgain` function that does nothing other than call `this.props.tryAgain`, rather than just giving `this.props.tryAgain` to the click handler?

src/components/Header.tsx
```diff
@@ -63,8 +63,8 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
     const currentPathname = (this.context.router &&
       this.context.router.getCurrentLocation() &&
       this.context.router.getCurrentLocation().pathname) || "";
-    let isLibraryManager =
-      this.context.library && this.context.admin.isLibraryManager(this.context.library());
+      let currentLibrary = this.context.library && this.context.library();
+    let isLibraryManager = this.context.admin.isLibraryManager(currentLibrary);
```
adriana-alter: Nice! :)

src/stylesheets/book_details_tab_container.scss
```diff
@@ -3,16 +3,8 @@
     padding-top: 2em;
   }
```
adriana-alter: Oh.  I see.  Disregard previous comments about button styling :)

src/components/BookCoverEditor.tsx
EdwinGuzman: I was thinking about that but the label says "other".

src/components/CustomListEditor.tsx
EdwinGuzman: Linter.

src/stylesheets/book_details_tab_container.scss
EdwinGuzman: Ha yeah. It was easier and probably better to update individually than set a global rule.

EdwinGuzman: The genre form's "add" button was the first one updated, and that caused the ripple effect in the other buttons.

src/components/BookCoverEditor.tsx
adriana-alter: Maybe the label shouldn't say "other" (although this would not be the branch in which to fix it)...

src/components/ErrorMessage.tsx
EdwinGuzman: Good point. If `ErrorMessage` is extended, the child component can override the class' `tryAgain` and inject any other behavior they want to. Probably that's the use case here.

src/components/BookCoverEditor.tsx
EdwinGuzman: Yeah, and I rather not get into what the label for an item's right should be. I'll update to uknown.

adriana-alter: Actually, maybe `unknown-status`.  `unknown` sounds like we don't know what the URI is.

src/components/BookCoverEditor.tsx
```diff
@@ -155,11 +158,12 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
...
+                ref="title_position"
               >
-                <option value="none">None</option>
-                <option value="top">Top</option>
-                <option value="center">Center</option>
-                <option value="bottom">Bottom</option>
+                <option value="none" aria-selected={titlePositionValue === "none"}>None</option>
+                <option value="top" aria-selected={titlePositionValue === "top"}>Top</option>
+                <option value="center" aria-selected={titlePositionValue === "center"}>Center</option>
+                <option value="bottom" aria-selected={titlePositionValue === "bottom"}>Bottom</option>
```
aslagle: This seems like it shouldn't be necessary - I'd expect screen readers to pick up the currently selected value on their own. Otherwise, they wouldn't be able to handle a plain HTML select with no JS. Does this actually change the behavior?

EdwinGuzman: This doesn’t change the behavior. I would agree that it’s not necessary and screen readers should pick it up, but unfortunately not all screenreaders behavior the same way. I’d say having this is a good fallback.

PR #251, opened 2019-07-23
Using buttons where they should be used, updating the Reusable Compon…

EdwinGuzman:
…ents package, cleaning up files and css.

This depends on [Resuable Components PR](https://github.com/NYPL-Simplified/reusable-components/pull/23) and resolves [SIMPLY-2136](https://jira.nypl.org/browse/SIMPLY-2136).

Updating elements that were either anchors or divs to buttons, but needed to update the `Button` component first to allow `aria-*` attributes as props.

A big but subtle change is the `Header`'s email button dropdown. The biggest changes are in the `WithEditButton` and `WithRemoveButton` components. In the book edit page it now appears like so:
Screen Shot 2019-07-23 at 11 26 04 AM
Screen Shot 2019-07-23 at 11 26 13 AM
Screen Shot 2019-07-23 at 11 26 18 AM

And in the Collections edit page:
Screen Shot 2019-07-23 at 11 26 32 AM

Also fixed/cleaned up styling issues:
Screen Shot 2019-07-19 at 4 32 57 PM

### DIFF
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -64,9 +64,9 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
           <ErrorMessage error={this.props.fetchError} tryAgain={this.refresh} />
         }
 
-        { this.state.customLists && this.state.customLists.map((list, idx) =>
+        { this.state.customLists && this.state.customLists.map(list =>
             <WithRemoveButton
-              key={`${list.id}-${idx}`}
+              key={list.id}
               disabled={this.props.isFetching}
               onRemove={() => this.removeList(list) }
               >
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -55,7 +55,7 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
 
   render(): JSX.Element {
     return (
-      <div>
+      <div className="custom-list-for-book">
         { this.props.book &&
           <h2>{this.props.book.title}</h2>
         }
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -64,9 +64,9 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
           <ErrorMessage error={this.props.fetchError} tryAgain={this.refresh} />
         }
 
-        { this.state.customLists && this.state.customLists.map(list =>
+        { this.state.customLists && this.state.customLists.map((list, idx) =>
             <WithRemoveButton
-              key={list.id}
+              key={`${list.id}-${idx}`}
               disabled={this.props.isFetching}
               onRemove={() => this.removeList(list) }
               >
@@ -92,7 +92,7 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
               value=""
               ref="addList"
               />
-            <Button className="left-align" callback={this.addList} content="Add" />
+            <Button className="left-align add-list-btn" callback={this.addList} content="Add" />
           </div>
         }
       </div>
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -92,7 +92,7 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
               value=""
               ref="addList"
               />
-            <Button className="left-align add-list-btn" callback={this.addList} content="Add" />
+            <Button className="left-align add-list-btn" callback={this.addList} content="Add" type="button" />
           </div>
         }
       </div>
--- a/src/components/ErrorMessage.tsx
+++ b/src/components/ErrorMessage.tsx
@@ -23,6 +23,13 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, {}>
         <Alert bsStyle="danger">
           <h4>You have been logged out.<br />
             <a target="_blank" href="/admin/sign_in_again">Log in again</a>
+            { this.props.tryAgain &&
+              <Button
+                className="left-align"
+                callback={this.tryAgain.bind(this)}
+                content="Try again"
+              />
+            }
           </h4>
         </Alert>
       );
--- a/src/components/ErrorMessage.tsx
+++ b/src/components/ErrorMessage.tsx
@@ -2,6 +2,7 @@ import * as React from "react";
 import * as ReactDOM from "react-dom";
 import { Alert } from "react-bootstrap";
 import { FetchErrorData } from "opds-web-client/lib/interfaces";
+import { Button } from "library-simplified-reusable-components";
 
 export interface ErrorMessageProps {
   error: FetchErrorData;
@@ -21,10 +22,7 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, {}>
       alertElement = (
         <Alert bsStyle="danger">
           <h4>You have been logged out.<br />
-            <a target="_blank" href="/admin/sign_in_again">Log in again</a><br />
-            { this.props.tryAgain &&
-              <a role="button" onClick={this.tryAgain.bind(this)}>Try again</a>
-            }
+            <a target="_blank" href="/admin/sign_in_again">Log in again</a>
           </h4>
         </Alert>
       );
@@ -55,7 +53,11 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, {}>
             }
             <span dangerouslySetInnerHTML={{ __html: errorMessageText }} /><br />
             { this.props.tryAgain &&
-              <a role="button" onClick={this.tryAgain.bind(this)}>Try again</a>
+              <Button
+                className="left-align"
+                callback={this.tryAgain.bind(this)}
+                content="Try again"
+              />
             }
           </h4>
         </Alert>
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -1,5 +1,4 @@
 import * as React from "react";
-import * as ReactDOM from "react-dom";
 import { connect } from "react-redux";
 import { Store } from "redux";
 import * as PropTypes from "prop-types";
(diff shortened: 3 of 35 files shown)


src/components/CustomListsForBook.tsx
```diff
@@ -64,9 +64,9 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
           <ErrorMessage error={this.props.fetchError} tryAgain={this.refresh} />
         }
 
-        { this.state.customLists && this.state.customLists.map(list =>
+        { this.state.customLists && this.state.customLists.map((list, idx) =>
             <WithRemoveButton
-              key={list.id}
+              key={`${list.id}-${idx}`}
```
EdwinGuzman: Just to fix warnings in the console.

src/components/ErrorMessage.tsx
```diff
@@ -21,10 +22,7 @@ export default class ErrorMessage extends React.Component<ErrorMessageProps, {}>
       alertElement = (
         <Alert bsStyle="danger">
           <h4>You have been logged out.<br />
-            <a target="_blank" href="/admin/sign_in_again">Log in again</a><br />
-            { this.props.tryAgain &&
```
EdwinGuzman: Removed this because it's redundant to have a link and a button next to each other that do similar behavior:
Screen Shot 2019-07-19 at 11 44 59 AM

src/components/Header.tsx
```diff
@@ -182,13 +184,14 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
...
-                <a
-                  className="account-dropdown-toggle"
-                  role="button"
+                <Button
+                  className="account-dropdown-toggle transparent"
+                  type="button"
                   aria-haspopup="true"
                   aria-expanded={this.state.showAccountDropdown}
-                  onClick={this.toggleAccountDropdown}
-                  >{ this.context.admin.email } &#9660;</a>
```
EdwinGuzman: Couldn't get the html entity to render correctly and is a problem with the `Button` component.

src/components/LanesSidebar.tsx
```diff
@@ -1,11 +1,6 @@
 import * as React from "react";
-import { Button } from "library-simplified-reusable-components";
 import AddIcon from "./icons/AddIcon";
-import PencilIcon from "./icons/PencilIcon";
 import ResetIcon from "./icons/ResetIcon";
-import VisibleIcon from "./icons/VisibleIcon";
-import HiddenIcon from "./icons/HiddenIcon";
-import XCloseIcon from "./icons/XCloseIcon";
```
EdwinGuzman: These were not removed when they were refactored.

src/components/WithEditButton.tsx
```diff
@@ -9,23 +10,24 @@ export interface WithEditButtonProps {
...
-        <a
-          className="sr-only"
-          role="button"
-          onClick={() => !this.props.disabled && this.props.onEdit()}
-          >edit</a>
+        <Button
+          className="edit-btn"
+          callback={this.onClick.bind(this)}
+          content={editContent}
+          type="button"
```
EdwinGuzman: The default type is `submit` which makes the form submit when you press "Enter". It's annoying but changing this to `type="button"` fixes it.

EdwinGuzman:
Tests will pass when the new Reusable Components package is published.

src/components/CustomListsForBook.tsx
```diff
@@ -92,7 +92,7 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
               value=""
               ref="addList"
               />
-            <Button className="left-align" callback={this.addList} content="Add" />
+            <Button className="left-align add-list-btn" callback={this.addList} content="Add" type="button" />
           </div>
         }
```
adriana-alter: Probably not coming from this branch, but this looks really squished to me; I think the remove button needs to be farther to the right.

Screen Shot 2019-07-23 at 12 09 27 PM

src/components/Header.tsx
adriana-alter: Let's use the down icon, in that case (like in the Panel); right now, there's no way to tell that it's a dropdown...

src/components/Lane.tsx
```diff
@@ -39,7 +39,7 @@ export default class Lane extends React.Component<LaneProps, LaneState> {
   }
 
   render(): JSX.Element {
-    let {lane, active, snapshot, provided, orderChanged, toggleLaneVisibility, parent, library, renderLanes} = this.props;
```
adriana-alter: Why aren't these needed anymore?

src/components/WithRemoveButton.tsx
```diff
@@ -9,23 +10,24 @@ export interface WithRemoveButtonProps {
...
-          className="sr-only"
-          role="button"
-          onClick={() => !this.props.disabled && this.props.onRemove()}
-          >remove</a>
+        <Button
+          className="remove-btn danger"
+          callback={this.onClick.bind(this)}
+          content={removeContent}
+          type="button"
+        />
```
adriana-alter: The alignment is off; the Add button is (slightly!) farther to the right than the Remove button.

Screen Shot 2019-07-23 at 12 17 35 PM

src/components/__tests__/ClassificationsForm-test.tsx
```diff
@@ -333,7 +334,8 @@ describe("ClassificationsForm", () => {
 
       expect(instance.validateAudience.callCount).to.equal(1);
       newGenres = wrapper.find(WithRemoveButton).map(name => name.text());
-      expect(newGenres).to.contain(instance.fullGenre("Folklore") + "remove");
+      console.log(newGenres);
```
adriana-alter: Leftover console statement

aslagle:
Some of those remove buttons seem excessive - they're now more prominent than the thing being removed.

src/components/CustomListsForBook.tsx
aslagle: Isn't the list id unique already?

src/components/ErrorMessage.tsx
aslagle: These don't do the same thing. After you log in, you might want to retry your original request, e.g. if you tried to submit a form but were logged out.

src/components/CustomListsForBook.tsx
EdwinGuzman: I thought it would be, too, but they were not unique in the tests...

src/components/ErrorMessage.tsx
EdwinGuzman: The button to "try again" still appears when it needs to. But if you have both at the same time for this situation and you try to log in again, you'll have to refresh anyway. Both at the same time is confusing. Maybe if there was a series of steps like, "click to log in again, then try again".

src/components/CustomListsForBook.tsx
EdwinGuzman: I can add more margin.

src/components/Lane.tsx
EdwinGuzman: They're needed for the component, but they are just not being used in the `render` function. No need to deconstruct them.

src/components/ErrorMessage.tsx
aslagle: I'm not sure if this is still the case, but the way it originally worked was that "log in again" would open a new tab, and once you logged in it would tell you to close the tab and try your request again. That way if you've already filled out a form you don't lose the changes.

EdwinGuzman:
Agreed, but I like the explicitness of having "delete" there. Maybe if the buttons were smaller and farther right, would that help?

src/components/ErrorMessage.tsx
EdwinGuzman: Yes, that behavior hasn't changed. But I wonder if anyone uses that workflow. But you're right about any unsaved changes. I'll add it back but would like to visit again later on to give users more details about the workflow.

src/components/WithRemoveButton.tsx
EdwinGuzman: Thankfully this only happens here.

adriana-alter:
What about going back to just having an icon, but having it be the garbage can instead of the X?

src/components/ErrorMessage.tsx
adriana-alter: I think maybe we need to get rid of the button and keep the link, not the other way around.

EdwinGuzman: But the link is the action that logs you back in, not the button.

aslagle: I think both are needed, but maybe adding some more text would make it clearer.

EdwinGuzman:
Maybe... I rather have the word "delete" or "remove" in there. If it's changed to the trash icon, the text should be updated to be consistent with the rest of the app.

EdwinGuzman:
Before I work on making this update, does this smaller edit/remove button work?
Screen Shot 2019-07-23 at 3 26 39 PM
Screen Shot 2019-07-23 at 3 26 13 PM

(@adriana-alter btw, the `small` class for the button doesn't really make it that much smaller).

adriana-alter:
That's a lot better.  I think it would help if there were significantly more space between the buttons and the thing they're for.  Can we just reuse the design from the EditableConfigList?  I definitely like the way the delete buttons are right-justified, and it might even make sense to have the bottom border.

src/components/WithRemoveButton.tsx
EdwinGuzman: The content of what's inside the `WithRemoveButton` component changes the distance. This isn't good but we'll see.

EdwinGuzman:
Updated everything that was mentioned.
Justified right looks fine
Screen Shot 2019-07-23 at 4 41 56 PM

adriana-alter:
That looks really nice!

src/components/CustomListsForBook.tsx
aslagle: I don't think this will hurt anything but it seems like something that could be fixed in the tests then.

EdwinGuzman: Okay, so this happens because when you add a new item it doesn't contain an id (the id gets added on the server-side when it's created). So for that moment until you save, react will complain. I don't think it's a big deal so I will leave it how it was before.

PR #252, opened 2019-07-26
Move self tests to diagnostics

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2099.

Goes with https://github.com/NYPL-Simplified/circulation/pull/1294

I'm still trying to figure out a bug involving receiving new self-test results, but in the meantime:

-- Instead of the "Diagnostics" page, there is now a "Troubleshooting" page, which contains both the Diagnostics and the Self-tests.  The Self-tests are organized/formatted so as to be visually consistent with the Diagnostics.

-- There is a link to the edit form for each service listed in the Self-tests section.

I'm on the fence re: whether to remove the self-tests from their current location in System Configuration (and replace them with a link to their new location, so people don't just think they've disappeared!), or whether to just have them in two places (to support more use cases).  Thoughts?

Screen Shot 2019-07-26 at 12 08 53 PM


EdwinGuzman:
What's the bug? I can run self-tests and they seem to update.

I do see a panel without a title:
Screen Shot 2019-07-26 at 12 55 14 PM

The whole point of moving the self-tests from their current location is to have them in one location. It doesn't make sense to have them in two locations because now it's confusing. Linking them to the troubleshooting page is better. I'd say don't do it for every collection/service but have a link at the top saying that any issues can be solved on the troubleshooting page, etc.

EdwinGuzman:
Also, move "troubleshooting" after "system configuration" in the navigation. Most of the work happens in the configuration page.

adriana-alter:
@EdwinGuzman Ready for review!

src/components/EditableConfigList.tsx
```diff
@@ -157,10 +158,22 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
     );
   }
 
+  renderLink(): JSX.Element {
+    const linkNames = {
+      "collection": "collections",
+      "patron authentication service": "patronAuthServices",
+      "search service": "searchServices"
+    };
+    return <h5>Problems with your {this.itemTypeName}{!this.limitOne && "s"}?  Please visit <a href={`http://localhost:6500/admin/web/troubleshooting/self-tests/${linkNames[this.itemTypeName]}`}>the troubleshooting page</a>.</h5>;
```
EdwinGuzman: I don't think this should be a heading, and I think it makes sense for this to go on top. If someone is expecting or wants to see the self tests, a link at the top to let them know right away that it's been move will be more helpful than at the bottom. Maybe it makes sense to have a temporary "self tests have been moved to..." message.

src/components/Collections.tsx
```diff
@@ -28,12 +27,12 @@ class CollectionEditForm extends ServiceWithRegistrationsEditForm<CollectionsDat
     the collection supports it. */
 export class Collections extends GenericEditableConfigList<CollectionsData, CollectionData, CollectionsProps> {
   EditForm = CollectionEditForm;
-  AdditionalContent = SelfTests;
   listDataKey = "collections";
   itemTypeName = "collection";
   urlBase = "/admin/web/config/collections/";
   identifierKey = "id";
   labelKey = "name";
+  hasSelfTests = true;
```
EdwinGuzman: One reason why Amy wanted me to name the variable `AdditionalContent` was to be generic for any component we wanted to pass in for any configuration. A different name that is not tied down to self tests is better in case you want to add a link for some other admin information.

src/components/Header.tsx
```diff
@@ -180,6 +173,13 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
                 >System Configuration</Link>
               </li>
             }
+            { isSystemAdmin &&
+              <li className="header-link">
+                <Link to={"/admin/web/troubleshooting"}
+                className={currentPathname.indexOf("/admin/web/troubleshooting") !== -1 ? "active-link" : ""}
```
EdwinGuzman: Indent this line.

src/components/EditableConfigList.tsx
```diff
@@ -157,10 +158,22 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
           </div>
         }
+        {
+          this.hasSelfTests && this.renderLink()
+        }
       </div>
     );
   }
 
+  renderLink(): JSX.Element {
```
EdwinGuzman: This is nice but maybe the classes that extend `EditableConfigList` should declare their own `renderLink` function so that `EditableConfigList` doesn't need to know about `collection`, `patron authentication service`, or `search service`.

src/components/SelfTestsTabContainer.tsx
```diff
@@ -99,10 +104,9 @@ function mapStateToProps(state, ownProps: SelfTestsTabContainerOwnProps) {
 
 function mapDispatchToProps(dispatch: Function, ownProps: SelfTestsTabContainerOwnProps) {
   let actions = new ActionCreator();
-  const fetchString = "fetch" + formatString(ownProps.tab);
-
+  const itemTypes = ["Collections", "PatronAuthServices", "SearchServices"];
   return {
-    fetchItems: () => dispatch(actions[fetchString]())
+    fetchItems: () => itemTypes.forEach(type => dispatch(actions["fetch" + type]()))
```
EdwinGuzman: It makes more sense the way you had it before by loading only the data you wanted instead of loading everything. If there are a lot of self tests it could slow the page down.

EdwinGuzman:
It is looking better. Left some code comments in the last commit.
Two things:
1) `DiagnosticsPage` and `SelfTestsPage` are pretty much the exact same file with the same props, state, and context, but only changing the tab container component that gets rendered. If later there is another tab that gets added then another file/component will need to be created. Why not create one class that takes in the tab container class you want. So instead of doing
```
  <DiagnosticsPage ... />
  <SelfTestsPage ... />
```
you can do 
```
  <TroubleShootingTypePage type={DiagnosticsTabContainer} />
  <TroubleShootingTypePage type={SelfTestsTabContainer} />
```

2) To get to the self tests there are now two clicks, one to open the panel and then another to expand the results in order to run. Maybe you can reduce the clicks by displaying the self test results when the open is open, so removing the "expand results" button.
Screen Shot 2019-08-05 at 10 19 44 AM

src/components/Collections.tsx
adriana-alter: But what if we need a link for self-tests AND a link for something else...?

EdwinGuzman: Would this `hasSelfTests` variable cover rendering both links in "renderLink"? Maybe a component that extends it and needs to render two links can implement its own renderLink function for both self-tests and a separate link?

src/components/SelfTestsTabContainer.tsx
adriana-alter: We can reassess if we ever get that many self-tests (which seems unlikely), but this is definitely the right way to do it for now.  The loading time this way is unnoticeable, whereas doing it the other way produces a long, annoying, very very noticeable delay and loading screen when you switch tabs.  Also, I like having it follow the same pattern as the Diagnostics page.

src/components/DiagnosticsServiceType.tsx
```diff
@@ -23,12 +23,12 @@ export default class DiagnosticsServiceType extends React.Component<DiagnosticsS
 
   render(): JSX.Element {
     let serviceTabs = this.props.services ?
-      <DiagnosticsServiceTabs
-        content={this.props.services}
-        tab={this.state.tab}
-        goToTab={this.goToTab}
-      /> :
-      <span>There are currently no {this.props.type.split("_").join(" ")} services.</span>;
+    <DiagnosticsServiceTabs
```
EdwinGuzman: The indentation was fine before.

src/components/SelfTestsTabContainer.tsx
```diff
@@ -0,0 +1,118 @@
...
+    } else {
+      return super.tabDisplayName(name);
+    }
+  }
+
+  getNames(category: string): string[] {
+    // The name used to look up data in the store: "collections", "patron_auth_services", "search_services".
+    let keyName = category.split(/(?=[A-Z])/).map(w => w.toLowerCase()).join("_");
+    // The name the SelfTests component is expecting: "collection", "patron_authentication_service", "search_service".
+    let typeName = keyName.replace("auth", "authentication").slice(0, -1);
```
EdwinGuzman: Why not leave it as `auth` instead of making a special case for it?

src/components/TroubleshootingCategoryPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  constructor(props) {
+    super(props);
+    let defaultSubtab = this.props.type === "diagnostics" ? "coverage_provider" : "collections";
+    this.state = { tab: this.props.subtab || defaultSubtab };
+    this.goToTab = this.goToTab.bind(this);
+  }
+
+  render(): JSX.Element {
+    let tabContainer = this.props.type === "diagnostics" ? DiagnosticsTabContainer : SelfTestsTabContainer;
```
EdwinGuzman: Maybe you don't foresee any more tabs to go into the trouble-shooting page, but if you do, then you'll have to keep updating this logic statement. I was hoping you'd just pass in the component as a child or prop, then you wouldn't have to worry about doing any checks here.

src/components/TroubleshootingCategoryPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+    return(
+      <div className={`${this.props.type}-page`}>
+       {
+         React.createElement(
+           tabContainer,
+           {
+             class: this.props.type === "diagnostics" ? "service-types" : "self-test-types",
+             store: this.context.editorStore,
+             csrfToken: this.context.csrfToken,
+             tab: this.props.subtab || this.state.tab,
```
EdwinGuzman: For `this.state.tab`, you already define it as `this.props.subtab` or `defaultSubtab`, so you don't need the first check.

src/components/EditableConfigList.tsx
```diff
@@ -158,10 +163,20 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
               />
           </div>
         }
+        {
+          this.hasSelfTests && <p>Problems with your {itemName}?  Please visit {link}.</p>
```
EdwinGuzman: Like in the other comment I made about the `AdditionalContent` component, this component shouldn't need to know about `selfTests`. That was one reason why it's called `AdditionalContent`. I think you can create something similar to that pattern for these links and not have the `hasSelfTests` variable in here. `linkName` seems fine and if a component that extends this class has it, then you can render the links. Same idea with the `renderLink` function. Instead of having the url for the troubleshooting page here, you can keep it generic by having each component with self tests define its own `renderLink` function.

src/components/SelfTestsTabContainer.tsx
adriana-alter: Because it's `authentication` in the URL, and it didn't seem worth making a separate server branch to change that.

src/components/DiagnosticsServiceType.tsx
adriana-alter: It went weird when I pulled in the new version of master.  Will revert.

src/components/SelfTestsTabContainer.tsx
EdwinGuzman: I rather have it consistent everywhere instead of having a special case for this one thing.

src/components/TroubleshootingCategoryPage.tsx
adriana-alter: It doesn't work without that.

src/components/SelfTestsTabContainer.tsx
adriana-alter: Okay, here you go! https://github.com/NYPL-Simplified/circulation/pull/1294

src/components/TroubleshootingCategoryPage.tsx
adriana-alter: I made the logic easier to update.  But I don't like the idea of implementing a pattern where the parent component needs to import its child's children.

src/components/TroubleshootingCategoryPage.tsx
EdwinGuzman: That's weird...

src/components/Collections.tsx
```diff
@@ -63,6 +62,15 @@ export class Collections extends GenericEditableConfigList<CollectionsData, Coll
     }
   }
 
+  renderLinks(): {[key: string]: JSX.Element} {
+    let linkBase = "/admin/web/troubleshooting/self-tests/collections";
+    let linkElement = <a href={linkBase}>the troubleshooting page</a>;
+    return {
+      "info": <>Self-tests for the collections have been moved to {linkElement}</>,
```
EdwinGuzman: I thought the react fragment didn't look good to you 👀
glad 👏 you 👏 made 👏 the 👏 change

EdwinGuzman: Seriously though, you can also interpolate the string so you don't have to make them into JSX elements. Either way is fine.

adriana-alter: I 👏 have 👏 no 👏 objection 👏 to 👏 React 👏 fragments 👏 in 👏 general!  Just thought it was confusing for whatever that specific case was (I don't even remember the details).

I tried interpolating and it didn't work :(

EdwinGuzman:
Just a cosmetic thing for the error message:

Screen Shot 2019-08-07 at 4 35 50 PM

EdwinGuzman:
I could've sworn this overflow issue before but maybe it's just for errors?
Screen Shot 2019-08-07 at 4 36 59 PM

src/components/Collections.tsx
EdwinGuzman: That makes sense. You can't interpolate an element.

src/components/TroubleshootingCategoryPage.tsx
EdwinGuzman: What doesn't work? I can navigate the tabs and go to them directly if I hit the url when I change it to `tab: this.state.tab`.

src/components/TroubleshootingCategoryPage.tsx
EdwinGuzman: Hmm it's also because `TroubleshootingCategoryPage` is a very specific component. If it was a generic `CategoryPage`, then importing that component's child component would be fine because it could potentially be _any_ component.

EdwinGuzman:
I like how reusable you made the `links` attribute for `GenericEdtiableConfig` now that it has nothing specific with self-tests. What do you think of doing it this way than and having each subclass define it's own `links`? Maybe it's more tedious to implement but I like how clean `EditableConfigList` is looking.

adriana-alter:
I'm not sure what you mean about the subclasses and `links`...?

Meanwhile:
Screen Shot 2019-08-12 at 3 30 06 PM

src/stylesheets/diagnostics.scss
```diff
@@ -71,6 +71,9 @@
       li {
         border-bottom: none;
         max-width: none;
+        .well {
```
EdwinGuzman: We should try not to add styles to these classes if we ever move away from bootstrap.

EdwinGuzman:
Okay, I can test now. Now that the layout is rearranged, I don't think the "Running new self tests" loading message is useful anymore.

adriana-alter:
It is extremely tedious to override the function in each class, but I think it's worth it for now, since only three things are using it, and it really does make `EditableConfigList` look a lot cleaner.  If we end up having more things that need to override it, we'll reassess.

I tried removing the loading message, and I thought the effect was weird and confusing; leaving it for now.

I changed the CSS so it doesn't rely on the bootstrap class name; good point.

PR #253, opened 2019-07-26
Scss cleanup

EdwinGuzman:
Resolves [SIMPLY-672](https://jira.nypl.org/browse/SIMPLY-672). For the most part, I broke apart big files into their React component. For some components, I left them as they were. Some are still slightly confusing because some names are used in multiple places and it's hard to figure out which one is the right rule (like `edit-form`).

Also, reduced some div output by using React 16's `<>` fragment component. Let me know what you think, I think it makes inspecting in the console easier and it looks cleaner to have less divs.

Everything should look the same but I did fix one styling issue with input fields extending past their container. There's also stuff like:
 
Screen Shot 2019-07-26 at 12 31 59 PM

which I did not touch.


src/components/BookDetailsEditor.tsx
```diff
@@ -69,9 +69,8 @@ export class BookDetailsEditor extends React.Component<BookDetailsEditorProps, {
 
   render(): JSX.Element {
     return (
-      <div>
-        { this.props.bookData && !this.props.fetchError &&
-          (<div>
+      <>
```
adriana-alter: It seems more confusing to do this as a fragment; why not just leave it as a `div` but give it a class name?

src/components/BookDetailsTabContainer.tsx
```diff
@@ -37,11 +37,7 @@ export class BookDetailsTabContainer extends TabContainer<BookDetailsTabContaine
 
   tabs() {
     const tabs = {};
-    tabs["details"] = (
-      <div className="details">
-        { this.props.children }
-      </div>
-    );
+    tabs["details"] = (<>{this.props.children}</>);
```
adriana-alter: Also seems more confusing...

adriana-alter:
This is really nice.  One overarching comment: what is the rationale for replacing all those divs with fragments?  I think the SCSS is a lot more readable (and tests down the road are easier to write) if we stick with the pattern of having components enclosed in a parent div with a classname.

EdwinGuzman:
Mostly because the divs don't add anything and then there's:
```
<div>
  <div>
    <div>
      <div class="container">
....
```

If it's confusing I can add it back in.

EdwinGuzman:
Btw, I only add back in the parent divs but left other fragments in place which reduce the html output. Later on, if we need to better and closer match HTML markup with USWDS, we'll have to use fragments. One that comes to mind that we've had a lot of issues in the past is the definition list element.

src/components/BookDetailsEditor.tsx
```diff
@@ -69,7 +69,7 @@ export class BookDetailsEditor extends React.Component<BookDetailsEditorProps, {
 
   render(): JSX.Element {
     return (
-      <>
+      <div className="bookDetailsEditor">
```
adriana-alter: Can this be "book-details-editor" to follow the pattern we use for other class names?

EdwinGuzman: Oh right right.

PR #255, opened 2019-08-12
Adjust catalog page styling

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2085; I made the catalog page elements that are coming from the OPDS component (breadcrumb links, lane titles, arrow buttons) visually consistent with the rest of the interface.

Screen Shot 2019-08-12 at 1 04 05 PM

### DIFF
--- a/src/stylesheets/colors.scss
+++ b/src/stylesheets/colors.scss
@@ -12,4 +12,5 @@ $dark-gray: #080807;
 $gray-tint: #F5F5F4;
 $white: #FFF;
 $blue: #1B7FA7;
+$blue-dark: #135772;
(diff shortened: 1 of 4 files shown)


src/stylesheets/colors.scss
```diff
@@ -12,4 +12,5 @@ $dark-gray: #080807;
 $gray-tint: #F5F5F4;
 $white: #FFF;
 $blue: #1B7FA7;
+$blue-dark: #135772;
```
EdwinGuzman: Kind of surprised we haven't used this before. Shouldn't this dark blue be used for more hover effects instead of just this then?

PR #257, opened 2019-08-15
Storybook

EdwinGuzman:
Resolves [SIMPLY-2162](https://jira.nypl.org/browse/SIMPLY-2162).

@adriana-alter just a draft for now but can you test it out and see if it works locally for you? Running `npm run storybook` works fine for me but when I try out `npm run build-storybook` and see that static site, there's a JS error. It's also preventing the `npm run deploy-storybook` to build correctly, and also preventing Travis from automatically building and deploying.

Here's the little that gets rendered on the `gh-pages` branch: http://nypl-simplified.github.io/circulation-web/?path=/story/*


adriana-alter:
Let me guess: `Super expression must either be null or a function, not undefined`?  Yes, happening to me too :(

adriana-alter:
On the bright side: the local version looks really nice so far!

EdwinGuzman:
I'm not going to close this in favor of #258 just yet. Locally, this works fine, but the static site (for deploying) fails to build correctly.

Here's my debugging process:
* `Autocomplete` builds fine but `ChangePasswordForm` does not build.
* What are the differences? `ChangePasswordForm` is more complex but the main point is that it imports `Form` from `reusable components`. Remove that call and everything works.
* So perhaps something is wrong with how `Reusable Components` is exporting its components, right? Not really. Whereas `Form` fails to be imported and rendered, `Button` works just fine.
* Without renaming any files, just copied the component from `Button` into `Form` but it _does not_ work. I can't tell if it was just bad file names or something in the component.
* We import the `opds-web-client` so maybe the components that import that package should fail. Nope.
* I tried updating the compilation target to be the same but it doesn't seem to be that.
* This was very frustrating but I will try to get around it again later on.

Will use Typedoc in #258 for now.

adriana-alter:
This sounds like the problem a few months ago that I resolved by switching over from Uglify to Terser.  Try running it without a minifier (just as an experiment) and see whether that fixes/changes the error message.

PR #258, opened 2019-08-16
Typedoc

EdwinGuzman:
I'm upset #257 did not work out, so here's an alternative. When building and running Storybook locally, everything works fine. But, when generating a static site, the build fails. I'll write more in the other PR. For now,

This will generate a Typescript-based document site: http://nypl-simplified.github.io/circulation-web/.

It's better than Storybook for documenting methods and interfaces, but there's no live preview of the components. Travis will deal with building and deploying the documentation site to Github pages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "test-chrome": "node_modules/.bin/selenium-standalone install && nightwatch -e chrome",
     "test-firefox": "node_modules/.bin/selenium-standalone install && nightwatch -e firefox",
     "dev": "webpack --watch --progress --display-modules --config webpack.dev.config",
-    "prod": "webpack -p --progress --display-modules --config webpack.prod.config"
+    "prod": "webpack -p --progress --display-modules --config webpack.prod.config",
+    "build-docs": "node_modules/.bin/typedoc --tsconfig tsconfig.json"
   },
   "dependencies": {
     "@nypl/dgx-svg-icons": "0.3.4",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -22,7 +22,8 @@
     "test-chrome": "node_modules/.bin/selenium-standalone install && nightwatch -e chrome",
     "test-firefox": "node_modules/.bin/selenium-standalone install && nightwatch -e firefox",
     "dev": "webpack --watch --progress --display-modules --config webpack.dev.config",
-    "prod": "webpack -p --progress --display-modules --config webpack.prod.config"
+    "prod": "webpack -p --progress --display-modules --config webpack.prod.config",
+    "build-docs": "node_modules/.bin/typedoc --tsconfig tsconfig.json"
```
aslagle: I thought npm already used `node_modules/.bin` as the path when running scripts?

src/stylesheets/entry_points_container.scss
```diff
@@ -50,6 +50,19 @@
   // EntryPointsTabs
   .entry-points-list {
     li {
+      a {
```
aslagle: Why are there styling changes in this PR?

package.json
EdwinGuzman: It does, but I like being explicit.

src/stylesheets/entry_points_container.scss
EdwinGuzman: Accidentally committed these.

PR #259, opened 2019-08-22
Refactor book edit form

adriana-alter:
This is just for stuff in the Details tab.  Two big changes:

1) The form for listing/adding/deleting contributors is its own component now.

2) Rather than copy-pasting the HTML for every single input field, BookEditForm now has a `renderTextField` function to generate most of them.

### DIFF
--- a/src/components/BookEditForm.tsx
+++ b/src/components/BookEditForm.tsx
@@ -1,10 +1,11 @@
 import * as React from "react";
 import EditableInput from "./EditableInput";
 import EditorField from "./EditorField";
-import { Button, Form } from "library-simplified-reusable-components";
+import { Form } from "library-simplified-reusable-components";
 import { BookData, ContributorData, RolesData, MediaData, LanguagesData } from "../interfaces";
-import WithRemoveButton from "./WithRemoveButton";
 import LanguageField from "./LanguageField";
+import { formatString } from "./sharedFunctions";
+import ContributorForm from "./ContributorForm";
 
 export interface BookEditFormProps extends BookData {
   roles: RolesData;
@@ -28,8 +29,6 @@ export default class BookEditForm extends React.Component<BookEditFormProps, Boo
     this.state = {
       contributors: (props.authors || []).concat(props.contributors || [])
     };
-    this.addContributor = this.addContributor.bind(this);
-    this.removeContributor = this.removeContributor.bind(this);
     this.renderForm = this.renderForm.bind(this);
   }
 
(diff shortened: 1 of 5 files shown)


src/components/ContributorForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+        name="contributor-name"
+        value={contributor && contributor.name}
+        ref={!contributor && "addContributorName"}
+        optionalText={false}
+      />
+    );
+  }
+
+  existingContributors() {
+    // The list of one or more existing contributors; each row has menu, input field, and delete button.
```
EdwinGuzman: These comments should be in jsdoc style, above the function name with `/** ... */`. This is more of Python's docstring style.

src/components/ContributorForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+          className="add-contributor small"
+          disabled={this.props.disabled}
+          callback={this.addContributor}
+          content="Add"
+        />
+      </span>
+    );
+  }
+
+  contributorSelect(contributorRole?: string, isNew = false): JSX.Element {
```
EdwinGuzman: Some of these methods are new and some are old but can you update them so that each param has a type?

src/components/ContributorForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+        <label>Authors and Contributors</label>
+        { this.existingContributors() }
+        { this.newContributor() }
+      </div>
+    );
+  }
+
+  newContributor(): JSX.Element {
+    // The component for adding a new contributor; menu, input field, and add button.
+    const newRoleRef = this.refs["addContributorRole"] as any;
```
EdwinGuzman: Use the React16 way of creating and using refs.

src/components/BookEditForm.tsx
```diff
@@ -1,10 +1,11 @@
 import * as React from "react";
 import EditableInput from "./EditableInput";
 import EditorField from "./EditorField";
-import { Button, Form } from "library-simplified-reusable-components";
+import { Form } from "library-simplified-reusable-components";
 import { BookData, ContributorData, RolesData, MediaData, LanguagesData } from "../interfaces";
-import WithRemoveButton from "./WithRemoveButton";
 import LanguageField from "./LanguageField";
+import { formatString } from "./sharedFunctions";
```
EdwinGuzman: I know this isn't new, but `sharedFunctions` isn't a component and should be in a `utils` folder.

src/components/BookEditForm.tsx
```diff
@@ -44,118 +43,36 @@ export default class BookEditForm extends React.Component<BookEditFormProps, Boo
     );
   }
 
+  renderTextField(name: string, placeholder?: string, value?: string, hasLabel = true): JSX.Element {
```
EdwinGuzman: This is very nice. It's very specific to this component/data but that's fine.

src/components/ContributorForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+export interface ContributorFormState {
+  contributors: ContributorData[];
+}
+
+export default class ContributorForm extends React.Component<ContributorFormProps, ContributorFormState> {
+  constructor(props) {
+    super(props);
+    this.state = { contributors: this.props.contributors || [] };
+    this.addContributor = this.addContributor.bind(this);
+    this.removeContributor = this.removeContributor.bind(this);
```
EdwinGuzman: I'm surprised you didn't have to add `this.getContributorRole` here.

src/components/ContributorForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  contributorSelect(contributorRole?: string, isNew = false): JSX.Element {
+    // The drop-down list of all the possible contributor roles.
+    return (
+      <EditableInput
+        elementType="select"
+        disabled={this.props.disabled}
+        name="contributor-role"
+        value={contributorRole || "Author"}
```
EdwinGuzman: Will this still work for the new contributor select? The old version has `value="Author"` hardcoded but now it is relying on `newRoleRef && newRoleRef.getValue();` which could have a value, no?

EdwinGuzman: Okay, I see. So this just updates the next input that is created to default the role back to Author. But now, if you select a value like "Composer" it will default to "Composer" for the next input. I don't think this is a big deal but it makes sense to default back to the most commonly used role.

src/components/ContributorForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+  roles: RolesData;
+  contributors?: ContributorData[];
+  disabled?: boolean;
+}
+
+export interface ContributorFormState {
+  contributors: ContributorData[];
+}
+
+export default class ContributorForm extends React.Component<ContributorFormProps, ContributorFormState> {
```
EdwinGuzman: I don't think this should be called `ContributorForm`. It's part of a form but it doesn't return an actual `<form>`. If this component is used in isolation, a form won't be rendered. I can't say I have a better name, but something that lets the developer know that what they expect to be rendered is going to be returned.

src/components/__tests__/ContributorForm-test.tsx
```diff
@@ -0,0 +1,105 @@
+import { expect } from "chai";
```
EdwinGuzman: 👍 nice tests. It'd be nice to test more functions like you did the `getContributorRole` but most just render components which are already tested so no need.

PR #262, opened 2019-08-30
Automatically update self-tests to show most recent results

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2261

### DIFF
--- a/src/components/SelfTests.tsx
+++ b/src/components/SelfTests.tsx
@@ -35,21 +35,35 @@ export interface SelfTestsProps extends React.Props<SelfTestsProps>, SelfTestsSt
 export interface SelfTestsState {
   runTests: boolean;
   error: FetchErrorData;
+  mostRecent?: ServiceData;
 }
 
 export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
-  constructor(props) {
+  constructor(props: SelfTestsProps) {
     super(props);
 
     this.state = {
       runTests: false,
       error: null,
+      mostRecent: this.props.item
     };
     this.runSelfTests = this.runSelfTests.bind(this);
   }
 
-  render() {
-    const integration = this.props.item;
+  componentDidMount() {
+    this.props.getSelfTests();
+  }
+
+  componentWillReceiveProps(nextProps: SelfTestsProps) {
+    let newTime = nextProps.item && nextProps.item.self_test_results && nextProps.item.self_test_results.start;
+    let oldTime = this.props.item && this.props.item.self_test_results && this.props.item.self_test_results.start;
+    if (!oldTime || (newTime > oldTime)) {
+      this.setState({ mostRecent: nextProps.item });
+    }
+  }
+
+  render(): JSX.Element {
+    const integration = this.state.mostRecent;
     const selfTestException = integration.self_test_results && integration.self_test_results.exception;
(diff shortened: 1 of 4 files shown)


src/components/SelfTests.tsx
```diff
@@ -35,21 +35,35 @@ export interface SelfTestsProps extends React.Props<SelfTestsProps>, SelfTestsSt
...
       error: null,
+      mostRecent: this.props.item
     };
     this.runSelfTests = this.runSelfTests.bind(this);
   }
 
-  render() {
-    const integration = this.props.item;
+  componentDidMount() {
+    this.props.getSelfTests();
```
EdwinGuzman: This is async, right? Would it be helpful to do async/await for this function?

src/components/SelfTests.tsx
```diff
@@ -35,21 +35,35 @@ export interface SelfTestsProps extends React.Props<SelfTestsProps>, SelfTestsSt
...
+  }
+
+  componentWillReceiveProps(nextProps: SelfTestsProps) {
+    let newTime = nextProps.item && nextProps.item.self_test_results && nextProps.item.self_test_results.start;
+    let oldTime = this.props.item && this.props.item.self_test_results && this.props.item.self_test_results.start;
```
EdwinGuzman: I was wondering if you could use `this.state.mostRecent` here since its value is `this.props.item` but it looks like you are using both `this.state.mostRecent` and `this.props.item` for different things. It seems unnecessary but I can see why it's not `this.state.mostRecent`. Seems like too much work is going on.

src/components/SelfTests.tsx
```diff
@@ -140,8 +152,19 @@ export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
 
 function mapStateToProps(state, ownProps) {
   const selfTests = state.editor.selfTests;
+  let item = ownProps.item;
+  if (selfTests && selfTests.responseBody && selfTests.responseBody.self_test_results) {
+    if (ownProps.type.includes(selfTests.responseBody.self_test_results.goal)) {
```
EdwinGuzman: Does `selfTests.responseBody.self_test_results.goal` only contain one goal?

EdwinGuzman:
I think the tests are lacking... could you write some for the `componentWillReceiveProps` update? I think it may be simple to pass in props and check the state.

src/components/SelfTests.tsx
adriana-alter: Yes; the goal can only be either `search` or `patron_auth`.

src/components/SelfTests.tsx
adriana-alter: I tried it; didn't make a difference.

src/components/SelfTests.tsx
adriana-alter: It does seem like a lot going on...but it was the only thing I found that worked.  Open to other ideas!  I don't think there's an actual downside to doing it like this, though.  (It's definitely important to keep the most recent result and the previously-displayed result separate; squishing them together was what was causing the bug before.)

EdwinGuzman:
Seems fine, just update the changelog.

EdwinGuzman:
The footer is not a big update that it needs to be its own version.

PR #263, opened 2019-09-05
Updating the link color in the tabs for catalog and book detail pages.

EdwinGuzman:
Resolves [SIMPLY-2280](https://jira.nypl.org/browse/SIMPLY-2280). Just updating link colors that were missed to be the default blue we are using. I created a new scss file since same elements get rendered in different pages (the catalog, the book detail, the hidden, and the complaints pages).

### DIFF
--- a/src/stylesheets/catalog_overrides.scss
+++ b/src/stylesheets/catalog_overrides.scss
@@ -1,0 +1,22 @@
+.collection-container {
+  .book {
+    .item-icon svg {
+      fill: $blue-dark;
+    }
+    .item-details {
+      color: $blue-dark;
+    }
+  }
+  .more-link div {
+    color: $blue;
+    svg {
+      fill: $blue;
+    }
+    &:hover {
+      color: $blue-dark;
+      svg {
+        fill: $blue-dark;
+      }
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


src/stylesheets/catalog_overrides.scss
```diff
@@ -0,0 +1,22 @@
+.collection-container {
+  .book {
+    .item-icon svg {
+      fill: $blue-dark;
```
EdwinGuzman: I decided to make the icon and the text for each book's title and author the dark blue when the book cover is hovered. Otherwise, if you update each color individually, it's confusing because you are already in a hovered state for the book cover, but then still need to hover over each individual piece of text for the updated hover color. Clicking anywhere goes to the same place, so to reduce confusion it's better to just make it the hover color.

adriana-alter: Did you mean for it to be `$blue` rather than `$blue-dark` on line 4, though?  I agree that everything should be the dark blue on hover.  But right now, nothing is happening on hover; the icon is already the hover color...

EdwinGuzman: Yes, updated.

EdwinGuzman:
@adriana-alter I'm not publishing a new version since it's small and can include the footer and self-tests update.

PR #265, opened 2019-09-11
First pass (no access_problem yet)

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2283

Screen Shot 2019-09-11 at 4 11 15 PM

Do we already know for sure what format the `access_problem` would be in, whether the format would always be the same, whether the problem would always be effectively the same (i.e. do we need the actual content of the `access_problem`, or is it safe to say that, if there is a problem, we should always display something to the effect of "there are terms but we're having trouble showing them"), etc.?  I've held off on implementing the error behavior for now.

### DIFF
--- a/src/components/LibraryRegistrationForm.tsx
+++ b/src/components/LibraryRegistrationForm.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
-import { LibraryData } from "../interfaces";
+import { LibraryData, LibraryRegistrationData } from "../interfaces";
 import EditableInput from "./EditableInput";
-import { Button, Form } from "library-simplified-reusable-components";
+import { Form } from "library-simplified-reusable-components";
 
 export interface LibraryRegistrationFormState {
   checked: boolean;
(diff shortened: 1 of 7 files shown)


src/components/LibraryRegistrationForm.tsx
```diff
@@ -28,27 +29,33 @@ export default class LibraryRegistrationForm extends React.Component<LibraryRegi
...
-  checkbox(library: LibraryData, termsLink: string): JSX.Element {
+  checkbox(library: LibraryData): JSX.Element {
+    let { access_problem, terms_of_service_html, terms_of_service_link } = this.props.registrationData;
     return (
       <section className="registration-checkbox">
-        <label>I have read and agree to the <a href={termsLink}>terms and conditions</a></label>
+        {
+          terms_of_service_html ?
+          <label dangerouslySetInnerHTML={{__html: terms_of_service_html}}></label> :
+          <label>I have read and agree to the <a href={terms_of_service_link}>terms and conditions</a>.</label>
```
EdwinGuzman: This is not really a label so a different element, like a paragraph, will be better. Or a div element for the first case.

EdwinGuzman:
You can make a PR a "draft" before submitting it whenever you are doing passes if that will help you (and lets me know that it is ready to review and not waiting on anything).

PR #267, opened 2019-09-12
Update circulation events report UI by library

EdwinGuzman:
This is the second part for #266 and goes along with [circulation #1316](https://github.com/NYPL-Simplified/circulation/pull/1316). Resolves [SIMPLY-2274](https://jira.nypl.org/browse/SIMPLY-2274).

For this PR, if a library short name is passed to the `CirculationEvents` component, then it will use that in its request URL. This means that `CirculationEvents` will always render in the `Dashboard` component regardless if it is in the context of a library or not.

If it is not in the context of a library, then it will query for events for all libraries.
If it is in the context of a library's dashboard, then it will query for events associated with that library.

This is okay but I think `CirculationEvents` could be refactored. The real-time stream of events may not be needed per library and was not part of the ticket so I removed it only when in the context of a library (along with the function call to the server).

### DIFF
--- a/src/components/CirculationEvents.tsx
+++ b/src/components/CirculationEvents.tsx
@@ -25,6 +25,7 @@ export interface CirculationEventsDispatchProps {
 export interface CirculationEventsOwnProps {
   store?: Store<State>;
   wait?: number;
+  library?: string;
 }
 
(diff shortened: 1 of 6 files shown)


EdwinGuzman:
We can first check with admins to see if it is useful or not. What we can do is refactor the main component and break out the form and the events table, so if it's not used, we can easily remove it.

PR #268, opened 2019-09-16
Crosslink location analytics

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2184


src/components/NeighborhoodAnalyticsForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+    super(props);
+    this.state = { selected: props.currentValue || null };
+    this.handleNeighborhoodChange = this.handleNeighborhoodChange.bind(this);
+    this.renderNeighborhoodForm = this.renderNeighborhoodForm.bind(this);
+  }
+
+  render(): JSX.Element {
+    return (
+      <Panel
+        headerText={`Patron Neighborhood Analytics: ${this.isEnabled() ? "En" : "Dis"}abled`}
```
EdwinGuzman: The only thing that I find off about this is that no other panel header does this dynamic title. So now we have a new pattern that is being introduced. If there's some other optional form in this field, then this may not be correct later.

src/components/NeighborhoodAnalyticsForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+          <p className="bg-danger">This feature will work only if it is also enabled in your <a href={url}>{name}</a>.</p>
+        }
+      </fieldset>
+    );
+  }
+
+  getPairedService(setting: SettingData): string[] {
+    // Whichever type of service we're currently dealing with--patron authentication or analytics--we need to provide a link to the other one.
+    const services = { "patronAuth": "patron authentication", "analytics": "analytics" };
+    let targetService = setting.key === "location_source" ? "patronAuth" : "analytics";
```
EdwinGuzman: This should say "local analytics" so admins don't get confused and think it's also for Google Analytics.

src/components/NeighborhoodAnalyticsForm.tsx
adriana-alter: I think it's worth it, since we're warning users that the feature will only work if they've enabled it on both analytics and patron auth; it seems important to give them a really easy way to tell whether it's currently enabled.  As far as I know, there won't be any other options in this field that would mess this up.  I don't love the inconsistency either, though, so if you have a different idea for how to get that information across...

src/components/NeighborhoodAnalyticsForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+  constructor(props: NeighborhoodAnalyticsFormProps) {
+    super(props);
+    this.state = { selected: props.currentValue || null };
+    this.handleNeighborhoodChange = this.handleNeighborhoodChange.bind(this);
+    this.renderNeighborhoodForm = this.renderNeighborhoodForm.bind(this);
+  }
+
+  render(): JSX.Element {
+    return (
+      <Panel
```
EdwinGuzman: `id` is a required prop.

src/components/NeighborhoodAnalyticsForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+        { setting.options.map(o =>
+            <option key={o.key} value={o.key} aria-selected={o.key === this.props.currentValue}>
+              {o.label}
+            </option>
+          )
+        }
+        </EditableInput>
+        {
+          this.isEnabled() &&
+          <p className="bg-danger">This feature will work only if it is also enabled in your <a href={url}>{name}</a>.</p>
```
EdwinGuzman: I don't think this should be a danger field because those are used when something is wrong. I do get that we want to display critical information but if everything is set it up and that message appears like that, it's possible to think that something is wrong. Maybe making it a normal info box will be enough.

adriana-alter: Yes.  Good point.  Do you think a warning box (yellow) would be ok, or is that still too alarming?

EdwinGuzman: Yeah, try it.

src/components/NeighborhoodAnalyticsForm.tsx
EdwinGuzman: Right, there are no current fields but it may be possible to add more in the future. It's fine but the inconsistency bothers me.

src/components/ServiceEditForm.tsx
```diff
@@ -126,6 +130,19 @@ export default class ServiceEditForm<T extends ServicesData> extends React.Compo
       />
     );
 
+    let neighborhoodForm = showNeighborhoodForm && (
+      <NeighborhoodAnalyticsForm
+        setting={this.protocolNeighborhoodSetting()}
```
EdwinGuzman: `this.protocolNeighborhoodSetting()` is already assigned to `showNeighborhoodForm` so you should use that instead of calling the function again.

src/components/ServiceEditForm.tsx
```diff
@@ -441,6 +458,19 @@ export default class ServiceEditForm<T extends ServicesData> extends React.Compo
     return [];
   }
 
+  protocolNeighborhoodSetting(): SettingData {
+    if (this.state.protocol && this.props.data && this.props.data.protocols) {
+      for (const protocol of this.props.data.protocols) {
+        if (protocol.name === this.state.protocol) {
+          let neighborhoodSetting = protocol.settings.find(s => ["neighborhood_mode", "location_source"].includes(s.key));
```
EdwinGuzman: This is very similar to `protocolSettings()`. Why not make an extra fields return value in `protocolSettings` so that it returns the nonrqeuired, required, and "other" fields (which in this case would be the neighborhood ones).
Then you can reuse the `["neighborhood_mode", "location_source"]` array and you will already have `protocol.settings` available.

EdwinGuzman:
`ServiceEditForm` is a generic component that should be able to render any form with any options for any setting. Now that component knows too much about `NeighborhoodAnalyticsForm` and its related values/functions.

I think it should work more like the `GenericEditableConfigList` component that takes in an `EditForm` generic named component and an `AdditionalContent` generic named component (this is how the self-test component worked before). In `PatronAuthServices`, for example, you can assign `NeighborhoodAnalyticsForm` to something more generic like `extraForm`, handle some more values to pass in `PatronAuthServices` and then pass in `extraForm` as a prop to `EditForm` in `GenericEditableConfigList`. That should help make `ServiceEditForm` less coupled with `NeighborhoodAnalyticsForm` and allow other forms to be added.

This is a big ask but I think it's a good update. I can see a few areas that managing data values will be tricky, but I can help you there.

adriana-alter:
I agree that this is a good idea, but I think it belongs in its own ticket/branch; 1) it may take a while, and I don't want to block the release of this feature, and 2) for consistency, we should also break out the libraries form into its own component and handle it the same way as the neighborhood form, which is out of scope for this branch.  What do you think?

src/components/NeighborhoodAnalyticsForm.tsx
adriana-alter: Hmmm.  I would like to change this feature so that it knows whether you've enabled the paired service, and can tell you right there in the form that you're currently on.  But I don't want to write that right now, because the refactor you suggested below means I'll just have to re-write it.  I vote we leave the current behavior as a placeholder, with the understanding that, by the time we add more fields, we will have replaced it with the new behavior.

EdwinGuzman:
It's fine to move out the libraries form out of `ServiceEditForm` but I don't think it should be handled the same way as neighborhood form. The library form is used often by different forms and services but the neighborhood form will only be used twice for two services. Since it's not being used often, I think that can be the "extraForm"(s) for specific services. I do think it'll take a while but we don't always go back to refactor components. I don't think there's a deadline so I would rather get this updated before merging.

src/components/AnalyticsServices.tsx
```diff
@@ -3,12 +3,15 @@ import { connect } from "react-redux";
 import ActionCreator from "../actions";
 import { AnalyticsServicesData, AnalyticsServiceData } from "../interfaces";
 import ServiceEditForm from "./ServiceEditForm";
+import NeighborhoodAnalyticsForm from "./NeighborhoodAnalyticsForm";
```
EdwinGuzman: The updates are looking good so far.

src/components/ServiceEditForm.tsx
```diff
@@ -400,86 +394,44 @@ export default class ServiceEditForm<T extends ServicesData> extends React.Compo
...
+  protocolSettings(protocol: ProtocolData) {
+    let requiredFields = [];
+    let nonRequiredFields = [];
+    let neighborhoodFields = [];
+    let neighborhoodSettings = ["neighborhood_mode", "location_source"];
+    let settings;
+    if (protocol) {
+      settings = this.state.parentId ? protocol.child_settings : protocol.settings;
+      settings && settings.forEach((s) => {
+        neighborhoodSettings.includes(s.key) ? neighborhoodFields.push(s) :
```
EdwinGuzman: Write a comment for this nested ternary statement.

src/components/ServiceEditForm.tsx
```diff
@@ -261,21 +262,21 @@ export default class ServiceEditForm<T extends ServicesData> extends React.Compo
                   onRemove={() => this.removeLibrary(library)}
                   ref={library.short_name}
                   >
-                  { this.props.data && this.props.data.protocols && this.protocolLibrarySettings() && this.protocolLibrarySettings().length > 0 &&
+                  { this.props.data && this.props.data.protocols && this.protocolLibrarySettings(protocol) && this.protocolLibrarySettings(protocol).length > 0 &&
```
EdwinGuzman: The nice thing about this refactor is that every time you call `this.protocolLibrarySettings()`, it no longer does a loop. You could assign `this.protocolLibrarySettings(protocol)` to a variable but it's not as necessary as it was before.

src/components/EditableConfigList.tsx
```diff
@@ -69,6 +79,8 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
   limitOne = false;
   links?: {[key: string]: JSX.Element};
   AdditionalContent?: new(props: AdditionalContentProps<T, U>) => React.Component<AdditionalContentProps<T, U>, any>;
+  ExtraFormSection?: new(props: ExtraFormSectionProps<T, U>) => React.Component<ExtraFormSectionProps<T, U>, any>;
```
EdwinGuzman: The `T` and `U` generics aren't needed but nice to have in case they become useful later.

PR #269, opened 2019-09-20
Router fix

EdwinGuzman:
Resolves [SIMPLY-2281](https://jira.nypl.org/browse/SIMPLY-2281).

I'm slightly concerned that it may be over-engineered but there was no straight forward way to refactor the list of links. I'm not sure what, how, or when but the three catalog links - "Catalog", "Complaints", "Hidden Books" - stopped correctly displaying as active when you were on that page. These links are part of an app inside the main app so maybe there's something involved there... but  I couldn't figure out why the Header which was inside the `opds-web-client` would not re-render when it received a new router context. So if you were on "Hidden Books" that link would initially be correctly active, but if you then went to "Complaints" it would not be active and "Hidden Books" would remain active. I updated them to be Bootstrap's `NavItem` components. This causes a page refresh which is unfortunate but correctly shows which page you are on.

Why not make them `Link` components like the rest of the navigation items? Doing so leads to the same issue. It does work better since it stays "within" the app and the loading time is faster, but the internal router doesn't update and so the Header doesn't actually know that it has been updated, and the links don't show up as active.

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -75,17 +75,20 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
     let currentLibrary = this.context.library && this.context.library();
     let isLibraryManager = this.context.admin.isLibraryManager(currentLibrary);
     let isSystemAdmin = this.context.admin.isSystemAdmin();
+    // Links that will be rendered in a NavItem Bootstrap component.
     const libraryNavItems = [
       { label: "Catalog", href: "%2Fgroups" },
       { label: "Complaints", href: "%2Fadmin%2Fcomplaints" },
       { label: "Hidden Books", href: "%2Fadmin%2Fsuppressed" },
     ];
+    // Links that will be rendered in a Link router component and are library specific.
     const libraryLinkItems = [
       { label: "Lists", href: "/admin/web/lists/" },
       { label: "Lanes", href: "/admin/web/lanes/", auth: isLibraryManager },
       { label: "Dashboard", href: "/admin/web/dashboard/"  },
       { label: "Patrons", href: "/admin/web/patrons/", auth: isLibraryManager },
     ];
+    // Links that will be rendered in a Link router component and are sitewide.
     const sitewideLinkItems = [
       { label: "Dashboard", href: "/admin/web/dashboard/",
         auth: (!this.context.library || !currentLibrary) },
@@ -185,7 +188,14 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
     this.setState({ showAccountDropdown });
   }
 
-  renderNavItem(item: HeaderNavItem, currentPathname: string, currentLibrary: string) {
+  /**
+   * renderNavItem
+   * Renders a NavItem Bootstrap component and is active based on the page's current path.
+   * @param {HeaderNavItem} item Object with the label and href for the navigation item.
+   * @param {string} currentPathname Page's current URL.
+   * @param {string} currentLibrary Active library.
+   */
+  renderNavItem(item: HeaderNavItem, currentPathname: string, currentLibrary: string = "") {
     const rootCatalogURL = "/admin/web/collection/";
     const href = item.href;
     const isActive = currentPathname.indexOf(href) !== -1;
@@ -202,6 +212,14 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
     );
   }
 
+  /**
+   * renderLinkItem
+   * Renders a Link Router component for library and sitewide navigation links
+   * and if the current admin has the correct authentication.
+   * @param {HeaderNavItem} item Object with the label and href for the navigation item.
+   * @param {string} currentPathname Page's current URL.
+   * @param {string} currentLibrary Active library.
+   */
   renderLinkItem(item: HeaderNavItem, currentPathname: string, currentLibrary: string = "") {
     const href = item.href;
     let isActive = currentPathname.indexOf(href) !== -1;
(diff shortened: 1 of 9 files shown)


EdwinGuzman:
Also, those header related console errors are hard to remove. It seems like `NavItem` isn't the correct element to use according to the docs, but `Nav.Link` crashes.

src/components/__tests__/Header-test.tsx
```diff
@@ -85,28 +84,25 @@ describe("Header", () => {
...
 
-      expect(catalogLinks.length).to.equal(3);
+      expect(navItems.length).to.equal(3);
 
-      let mainCatalogLink = catalogLinks.at(0);
-      expect(mainCatalogLink.prop("collectionUrl")).to.equal("/nypl/groups");
-      expect(mainCatalogLink.prop("bookUrl")).to.equal(null);
-      expect(mainCatalogLink.children().text()).to.equal("Catalog");
+      let mainNavItem = navItems.at(0);
+      expect(mainNavItem.prop("href")).to.contain("/nypl%2Fgroups");
```
adriana-alter: Why the extra `%` stuff in these now?

src/components/Header.tsx
```diff
@@ -230,6 +186,68 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
...
+          className={isActive ? "active-link" : ""}
+        >
+          {label}
+        </Link>
+      </li>
+    );
+
+    // Sometimes, some links should only be shown to admins who have
+    // specific privileges. If there is no restriction, always render the link.
+    if (auth !== undefined) {
```
adriana-alter: Does there need to be a separate check for `auth !== undefined`?  It looks to me like the `if (auth)` check will take care of this anyway.

src/components/Header.tsx
```diff
@@ -68,6 +74,30 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
...
     let isSystemAdmin = this.context.admin.isSystemAdmin();
+    // Links that will be rendered in a NavItem Bootstrap component.
+    const libraryNavItems = [
+      { label: "Catalog", href: "%2Fgroups" },
+      { label: "Complaints", href: "%2Fadmin%2Fcomplaints" },
+      { label: "Hidden Books", href: "%2Fadmin%2Fsuppressed" },
+    ];
+    // Links that will be rendered in a Link router component and are library specific.
+    const libraryLinkItems = [
+      { label: "Lists", href: "/admin/web/lists/" },
```
adriana-alter: Since all the links start with `/admin/web`, maybe just build that into `renderLinkItem` rather than having to specify it every time here...

adriana-alter:
Is there any real benefit to using Bootstrap here?  It's not like this is particularly complicated styling; if we just do it ourselves, that will get rid of the console errors...

src/components/__tests__/Header-test.tsx
EdwinGuzman: Those were always there in the URL for the catalog, complaints, and hidden books pages. Just easier to check against those because now those are different components.

src/components/Header.tsx
EdwinGuzman: Sure.

src/components/Header.tsx
EdwinGuzman: I'm not sure about the correct authentication for each admin level to see each link, but, for example, Lists and Dashboard don't have any type of admin check. If `if (auth)` is used, since there's no auth for those links, it will never be displayed regardless of whether you are allowed to see those links or not. Keeping the functionality the same, but admin levels and access is something we can bring up.

EdwinGuzman:
Nah, we should move away from Bootstrap and we should do it ourselves but something like that will require a bigger conversation with UX and maybe Amigos (and perhaps updating the default header that gets rendered on non-admin sites). So that's an easy but more complicated change which I don't want to get into right now without setting reqs and acceptance criteria.

src/components/Header.tsx
```diff
@@ -82,21 +82,21 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
...
         auth: (!this.context.library || !currentLibrary) },
-      { label: "System Configuration", href: "/admin/web/config/",
+      { label: "System Configuration", href: "config/",
         auth: this.context.admin.isLibraryManagerOfSomeLibrary() },
-      { label: "Troubleshooting", href: "/admin/web/troubleshooting/",
+      { label: "Troubleshooting", href: "troubleshooting/",
         auth: isSystemAdmin },
     ];
-    const accountLink = { label: "Change password", href: "/admin/web/account/" };
+    const accountLink = { label: "Change password", href: "account/" };
```
adriana-alter: If you wanted to--and this is very much just a matter of personal preference!--you could further simplify this function.  All the hrefs end with `/`...and (except for "System Configuration") each label is just an uppercased version of its href (which you can do very easily with the `formatString` function from `utils/sharedFunctions`.  You might want to leave the `sitewideLinkItems` array alone (because of the extra `auth` complication, and because it would be annoying to special-case "System Configuration")...but for the library links, you could just map over an array of the hrefs, build the ending `/` in for everything, and have the label be `formatString(href)`.  (Or not!  There's nothing wrong with the current implementation; this is just something that occurred to me when I looked at it.)

src/components/Header.tsx
```diff
@@ -82,21 +82,21 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
...
+      { label: "Patrons", href: "patrons/", auth: isLibraryManager },
     ];
     // Links that will be rendered in a Link router component and are sitewide.
     const sitewideLinkItems = [
-      { label: "Dashboard", href: "/admin/web/dashboard/",
+      { label: "Dashboard", href: "dashboard/",
         auth: (!this.context.library || !currentLibrary) },
-      { label: "System Configuration", href: "/admin/web/config/",
+      { label: "System Configuration", href: "config/",
         auth: this.context.admin.isLibraryManagerOfSomeLibrary() },
```
adriana-alter: Can you make a variable for this and the thing on line 93 (like the `isSystemAdmin` one, and the other stuff defined at the top of `render`)? (Just for consistency and readability.)

PR #270, opened 2019-10-07
Fixing Firefox bug where circulation events excel file doesn't download.

EdwinGuzman:
Resolves [SIMPLY-2307](https://jira.nypl.org/browse/SIMPLY-2307).

Minor description update for the location input text field.
Fixing a Firefox bug where it doesn't understand a virtually created anchor element and a click event unless it is part of the DOM.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.3.20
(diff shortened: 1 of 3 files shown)


leonardr:
:+1: on the wording change

PR #271, opened 2019-10-10
Re-aligned!

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2343

Before:
Screen Shot 2019-10-10 at 3 44 59 PM

After:
Screen Shot 2019-10-10 at 3 58 54 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.3.21
+#### Fixed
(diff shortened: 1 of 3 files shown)


EdwinGuzman:
Great. Was this only happening in that page and are other instances aligned too?

adriana-alter:
It was also happening for the languages on the library form, and the contributors on the book edit form.  Everything good now!

PR #273, opened 2019-10-22
Lists tab

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2254

Previous behavior: users could type something other than the name of an actual list into the input field, or just leave the field blank, and submit it with no problems.  As demonstrated by the "before" screenshot, the book would get saved to a "list" whose name was a blank string, or to something that wasn't one of the existing lists.  Among other things, this meant that if users wanted to add the book to an existing list but typed the list's name wrong, they would never know that the book hadn't actually been added to the intended list.  (The previous input field was an autocomplete field, so it did present the names of all the lists, but it still accepted invalid values.)

New behavior: there is a drop-down menu so users can choose from among the existing lists.  The menu choices are automatically narrowed down to include only lists that the book isn't on yet.  (The menu is hidden if the book is on all of the available lists.)  If users want to add the book to a list that doesn't exist yet, there is a link for them to create a new list.  The book's title is automatically copied to the clipboard when they follow the link, so that if it's really long (see the example in the "after" screenshot!), they can just paste it into the list creator's search field rather than having to type it out.  Also, the lists tab is now more consistent with the rest of the interface.

Before:
Screen Shot 2019-10-16 at 4 49 50 PM

After:
Screen Shot 2019-10-23 at 11 09 02 AM

In terms of code changes: the lists tab now uses the InputList component (also used for geographic areas and languages in the library edit form).  The InputList now has additional options, which make it more reusable: it can render a dropdown menu, and it can optionally call a callback when individual items are added or deleted (rather than everything being submitted at once, from a Submit button).

### DIFF
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -59,13 +59,13 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
         { this.props.book &&
           <h2>{this.props.book.title}</h2>
         }
-        <h3>Lists</h3>
+          <h3>Lists:</h3>
         { this.props.fetchError &&
           <ErrorMessage error={this.props.fetchError} tryAgain={this.refresh} />
         }
         <Form className="edit-form" content={[
           this.renderInputList(),
-          <a href={`/admin/web/lists/${this.props.library}/create`}>Create a new list</a>
+          this.renderLink()
         ]} onSubmit={this.save} />
       </div>
     );
@@ -75,26 +75,42 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
     return <option value={list.name} key={list.id} aria-selected={false}>{list.name}</option>;
   }
 
+  renderLink() {
+    if (this.props.allCustomLists && this.props.allCustomLists.length) {
+      return ([
+        <hr key="hr1"/>,
+        <span key="or">or</span>,
+        <hr key="hr2"/>,
+        <div key="list-creator-link" onClick={() => {(navigator as any).clipboard.writeText(this.props.book.title)}}>
+          <a href={`/admin/web/lists/${this.props.library}/create`}>Create a new list</a>
+          <p>(The book title will be copied to the clipboard so that you can easily search for it on the list creator page.)</p>
+        </div>
+      ]);
+    }
+  }
+
   renderInputList() {
     let allLists = this.props.allCustomLists || [];
-    return (
+    if (allLists.length) {
+      return (
         <ProtocolFormField
-          key={"lists"}
-          ref={"addList"}
-          setting={{
-            type: "menu",
-            format: "narrow",
-            key: "lists?",
-            label: null,
-            custom_lists: this.state.customLists,
-            required: true,
-            menuOptions: allLists.filter(l => l.name).map(l => this.makeSelect(l)),
-            urlBase: this.makeURL
-          }}
-          disabled={this.props.isFetching}
-          value={this.state.customLists && this.state.customLists.map(l => l.name)}
+        key={"lists"}
+        ref={"addList"}
+        setting={{
+          type: "menu",
+          format: "narrow",
+          key: "lists-input",
+          label: null,
+          custom_lists: this.state.customLists,
+          required: true,
+          menuOptions: allLists.filter(l => l.name).map(l => this.makeSelect(l)),
+          urlBase: this.makeURL
+        }}
+        disabled={this.props.isFetching}
+        value={this.state.customLists && this.state.customLists.map(l => l.name)}
         />
-    );
+      );
+    }
   }
 
   makeURL(listName): string {
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -13,7 +13,7 @@ import { State } from "../reducers/index";
 
 export interface CustomListsForBookStateProps {
   allCustomLists?: CustomListData[];
-  customListsForBook?: CustomListsData;
+  customListsForBook?: CustomListData[];
   fetchError?: FetchErrorData;
   isFetching?: boolean;
 }
@@ -36,7 +36,7 @@ export interface CustomListsForBookOwnProps {
 export interface CustomListsForBookProps extends CustomListsForBookStateProps, CustomListsForBookDispatchProps, CustomListsForBookOwnProps {};
 
 export interface CustomListsForBookState {
-  customLists?: CustomListsData;
+  customLists?: CustomListData[];
 }
 
 /** Tab on the book details page that shows custom lists a book is on and lets
@@ -111,7 +111,7 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
             urlBase: this.makeURL
           }}
           disabled={this.props.isFetching}
-          value={this.props.customListsForBook && this.props.customListsForBook.custom_lists.map(l => l.name)}
+          value={this.props.customListsForBook && this.props.customListsForBook.map(l => l.name)}
           altValue="This book is not currently on any lists."
           onSubmit={this.save}
         />
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -122,6 +122,7 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
           }}
           disabled={this.props.isFetching}
           value={this.state.customLists && this.state.customLists.map(l => l.name)}
+          altValue="This book is not currently on any lists."
         />
       );
     } else {
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -63,7 +63,10 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
         { this.props.fetchError &&
           <ErrorMessage error={this.props.fetchError} tryAgain={this.refresh} />
         }
-        <Form className="edit-form" content={this.renderInputList()} onSubmit={this.save} />
+        <Form className="edit-form" content={[
+          this.renderInputList(),
+          <a href={`/admin/web/lists/${this.props.library}/create`}>Create a new list</a>
+        ]} onSubmit={this.save} />
       </div>
     );
   }
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -87,7 +87,7 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
             menuOptions: allLists.filter(l => l.name).map(l => this.makeSelect(l))
           }}
           disabled={this.props.isFetching}
-          value={this.state.customLists.map(l => l.name)}
+          value={this.state.customLists && this.state.customLists.map(l => l.name)}
         />
     );
   }
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -10,8 +10,6 @@ import {
 } from "../interfaces";
 import { FetchErrorData } from "opds-web-client/lib/interfaces";
 import { State } from "../reducers/index";
-import { Button, Form } from "library-simplified-reusable-components";
-import { isEqual } from "../utils/sharedFunctions";
 
 export interface CustomListsForBookStateProps {
   allCustomLists?: CustomListData[];
@@ -44,7 +42,7 @@ export interface CustomListsForBookState {
 /** Tab on the book details page that shows custom lists a book is on and lets
     an admin add the book to lists or remove the book from lists. */
 export class CustomListsForBook extends React.Component<CustomListsForBookProps, CustomListsForBookState> {
-  constructor(props) {
+  constructor(props: CustomListsForBookProps) {
     super(props);
     this.state = {
       customLists: this.props.customListsForBook || []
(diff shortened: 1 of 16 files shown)


aslagle:
@adriana-alter would you mind if I connect you to someone from the Curation Corps to demo this before continuing on it? afaik they're the primary users of this tab and I'd like to make sure whatever you change is helping them.

adriana-alter:
@aslagle Absolutely!  Let me just figure out these failing tests firsts...

For what it's worth: the tab's intended functionality hasn't changed at all; I just made it more consistent with the rest of the interface, and made it so that you choose a list from a dropdown of available lists, to stop people from typo-ing list names, adding blank strings, adding books to lists that don't exist, etc.  (And I put in a link so that you can easily start a new list if you don't want to put the book on any of the existing ones.)

EdwinGuzman:
Very nice. It's working well.

When the book is question is added to every available list, there should be a message replacing the dropdown that says that the book has been added to every available list. At that point, the users can only delete or create a new list but are not told why those are the only options available, and why they can't add it to any more lists (maybe they don't know it was added to all the lists).

I agree with Amy that those who use the feature should review it. Like you said the functionality hasn't changed, but that doesn't mean a user will readily be able to use a new look of an existing feature (and users tend to not like UI changes anyway).

src/components/CustomListsForBook.tsx
```diff
@@ -59,87 +59,75 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
         { this.props.book &&
           <h2>{this.props.book.title}</h2>
         }
-        <h3>Lists</h3>
+          <h3>Lists:</h3>
```
EdwinGuzman: Maybe a more descriptive title like "Current Lists" or something. Unless if you make another heading (since this seems like a main heading) that says "Currently in these lists".

src/components/CustomListsForBook.tsx
```diff
@@ -59,87 +59,75 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
...
-              ref="addList"
-              />
-            <Button className="left-align add-list-btn" callback={this.addList} content="Add" type="button" />
-          </div>
-        }
+  renderLink(): JSX.Element {
+    // Link to the form for creating a new list
+    return(
+      <div role="button" key="list-creator-link" onClick={this.copyTitle}>
+        <a href={`/admin/web/lists/${this.props.library}/create`}>Create a new list</a>
```
EdwinGuzman: The `onClick` should be on the anchor element.

src/components/CustomListsForBook.tsx
```diff
@@ -59,87 +59,75 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
...
-        break;
-      }
+  renderInputList(): JSX.Element {
+    // The list of lists the book is already on, and the dropdown for adding it to others
+    let allLists = this.props.allCustomLists || [];
+    if (allLists.length > 0) {
+      return (
+        <ProtocolFormField
+          key={"lists"}
+          ref={"addList"}
```
EdwinGuzman: You can just do `key="lists"`

src/components/CustomListsForBook.tsx
```diff
@@ -167,9 +155,15 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
     this.props.refreshCatalog();
   }
 
-  save(lists: CustomListData[]) {
-    let url = this.listsUrl();
+  save() {
+    const listNames = (this.refs.addList as ProtocolFormField).getValue();
```
EdwinGuzman: Can you make that ref into the updated version of a ref instance?

src/components/CustomListsForBook.tsx
```diff
@@ -167,9 +155,15 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
...
+    const url = this.listsUrl();
     let data = new (window as any).FormData();
+    let lists = [];
+    listNames.map((name) => {
+      let list = this.props.allCustomLists.find(x => x.name === name);
```
EdwinGuzman: This is fine but I'm trying to think if there's a better way to keep track of the new/updated lists to save.

src/components/InputList.tsx
```diff
@@ -119,26 +176,41 @@ export default class InputList extends React.Component<InputListProps, InputList
...
       (this.refs["addListItem"] as any).refs["autocomplete"] :
       (this.refs["addListItem"] as any);
-
     const listItem = ref.getValue();
-    this.setState({ listItems: this.state.listItems.concat(listItem), newItem: "" });
-    ref.clear();
+    await this.setState({ listItems: this.state.listItems.concat(listItem), newItem: "" });
+    // Actually save the changes instead of just manipulating the state
+    if (this.props.onSubmit) { await this.props.onSubmit(); };
+    if (this.props.setting.type !== "menu") { ref.clear(); };
```
EdwinGuzman: I'd just put whatever is in curly braces into new lines for readability.

src/components/ProtocolFormField.tsx
```diff
@@ -27,12 +28,12 @@ export default class ProtocolFormField extends React.Component<ProtocolFormField
   }
 
   render(): JSX.Element {
-    const setting = this.props.setting;
+    const setting = this.props.setting as any;
```
EdwinGuzman: Why add `any`?

src/components/ProtocolFormField.tsx
```diff
@@ -1,19 +1,20 @@
...
 import ColorPicker from "./ColorPicker";
 import { Button } from "library-simplified-reusable-components";
 import InputList from "./InputList";
-import { SettingData } from "../interfaces";
+import { SettingData, CustomListsSetting } from "../interfaces";
 import { FetchErrorData } from "opds-web-client/lib/interfaces";
 
 export interface ProtocolFormFieldProps {
-  setting: SettingData;
+  setting: SettingData | CustomListsSetting;
```
EdwinGuzman: My only problem with this is that now `ProtocolFormField` knows about a specific type of setting instead of just the general one.

src/interfaces.ts
```diff
@@ -414,6 +415,18 @@ export interface CustomListsData {
   custom_lists: CustomListData[];
 }
 
+export interface CustomListsSetting {
```
EdwinGuzman: You can probably extend `SettingData` and then add the new values.

src/utils/sharedFunctions.ts
```diff
@@ -42,3 +42,7 @@ export function formatString(word: string, replacement?: string[], capitalize =
   }
   return word;
 }
+
+export function isEqual(array1, array2): boolean {
+  return array1.length === array2.length && array1.every(x => array2.indexOf(x) >= 0);
+}
```
EdwinGuzman: Add interface types for the params. `isEqual` may be too generic since this is specifically meant for arrays, but it's such a good and helpful function 👍 .
What if you are comparing something like ["a", "a", "b"] and ["b", "a", "a"]. I think I saw that the comparison would fail since there are duplicates but what if you want duplicates? I can't think of a case where that would actually happen in this app but just a thought.

src/components/InputList.tsx
```diff
@@ -1,109 +1,166 @@
...
+    const setting = this.props.setting as any;
+    // Hide the "Add" button if there are no options left
+    let showButton = !this.state.options || this.state.options.length > 0;
     return (
-      <div>
+      <div className="input-list">
         { this.props.labelAndDescription(setting) }
-        { this.state.listItems && this.state.listItems.map((listItem) => {
+        { this.props.altValue && !(this.state.listItems && this.state.listItems.length > 0) && <span>{this.props.altValue}</span> }
+        { this.state.listItems && this.state.listItems.filter(listItem => listItem).map((listItem) => {
```
EdwinGuzman: You're filtering for any list item that may be empty but when can that situation happen?

src/components/InputList.tsx
```diff
@@ -1,109 +1,166 @@
...
 
   render(): JSX.Element {
-    const setting = this.props.setting;
+    const setting = this.props.setting as any;
+    // Hide the "Add" button if there are no options left
+    let showButton = !this.state.options || this.state.options.length > 0;
     return (
-      <div>
+      <div className="input-list">
         { this.props.labelAndDescription(setting) }
```
EdwinGuzman: Since `setting` can now be of type `CustomListsSetting`, it's possible to not have a `label` in `setting` since it's optional. This makes it so an empty `<label>` renders and is not ideal so you should fix the underlying `labelAndDescription` function.
Screen Shot 2019-10-24 at 10 48 14 AM

src/components/InputList.tsx
```diff
@@ -1,109 +1,166 @@
...
-                  type: "text",
-                  description: null,
-                  disabled: this.props.disabled,
-                  value: value,
-                  name: setting.key,
-                  label: null,
-                  extraContent: this.renderToolTip(listItem, setting.format)
-                })
-              }
+              { this.renderListItem(setting, value, listItem) }
```
EdwinGuzman: The name can be misleading since it's not really a "list" item in the sense of an "li", which is what I was thinking of since it's "render...". It actually would be good to change the underlying DOM so that this is a list of `WithRemoveButton`s.

src/components/__tests__/CustomListsForBook-test.tsx
```diff
@@ -67,30 +67,79 @@ describe("CustomListsForBook", () => {
...
       expect(button.length).to.equal(1);
       expect(button.prop("content")).to.equal("Add");
     });
 
+    it("does not show the InputList if no lists exist", async () => {
+      let inputList = wrapper.find(InputList);
+      expect(inputList.length).to.equal(1);
+      wrapper.setProps({ allCustomLists: [] });
+      await new Promise((resolve) => setTimeout(resolve, 0));
+      wrapper.update();
```
EdwinGuzman: You don't need these lines.

src/components/__tests__/CustomListsForBook-test.tsx
```diff
@@ -115,101 +164,45 @@ describe("CustomListsForBook", () => {
...
-      expect(links.at(1).text()).to.equal("list 1");
-
+      expect(removables.length).to.equal(1);
+      expect(removables.find("a").text()).to.equal("list 2");
+
+      let menu = wrapper.find("select");
+      menu.getDOMNode().value = "list 1";
+      menu.simulate("change");
+      wrapper.find("button.add-list-item").simulate("click");
+      await new Promise((resolve) => setTimeout(resolve, 0));
```
EdwinGuzman: You can instead do `await wrapper.find("button.add-list-item").simulate("click");` instead of the dummy promise.

src/components/__tests__/CustomListsForBook-test.tsx
```diff
@@ -115,101 +164,45 @@ describe("CustomListsForBook", () => {
...
-      let removable = wrapper.find(WithRemoveButton);
-      expect(removable.length).to.equal(1);
-      let removeButton = removable.find(".remove-btn").hostNodes();
+    it("removes a list", async () => {
+      expect(editCustomListsForBook.callCount).to.equal(0);
+      let removables = wrapper.find(WithRemoveButton);
+      expect(removables.length).to.equal(1);
+      let removeButton = removables.find(Button);
       removeButton.simulate("click");
+      await new Promise((resolve) => setTimeout(resolve, 0));
```
EdwinGuzman: Same as the above comment: you can use await with the simulate method.

src/components/InputList.tsx
adriana-alter: If the admin has previously submitted an empty input field by accident (which the interface's current behavior allows; hence the third item in the first screenshot), then a blank string got saved as the name of one of the lists the book is on.  I don't know how frequently this has actually happened (probably not very!), but it still seems worth filtering those out.  Going forward, there won't be a way for the situation to happen again.

src/utils/sharedFunctions.ts
adriana-alter: No, that passes!  (I'll add a test for that specific example.)  The case with duplicates where it fails is if the arrays are the same length (because duplicates) and all the items from Array #2 are also in Array #1, but Array #1 has an additional item that Array #2 doesn't have (["a", "b", "c"] vs. ["a", "a", "b"]).

src/components/ProtocolFormField.tsx
adriana-alter: Because TypeScript (incorrectly) yells at me otherwise.

src/components/ProtocolFormField.tsx
adriana-alter: The only way I see around this is to add the `CustomListsSetting` stuff to `Setting Data` (as optional) and just call everything SettingData, and that seems like a bigger problem.  Other ideas?  (How much is this bothering you?)

adriana-alter:
Note: I have just done a demo of the new UI for Bri from the Open Bookshelf curation corps, and she thinks that everything looks good; we're all set to move forward with this.

src/components/InputList.tsx
EdwinGuzman: Right, but shouldn't this not happen now that users don't have the option to input a false list name?

adriana-alter: Yes, but for users who have already done it, using the old version.  We don't want the already-existing bad values to get displayed.

EdwinGuzman: I hope they don't wonder, "What happened to my list?" (even though it was a bad value).

adriana-alter: This is just filtering out the blank values.  Fake list names will still show up.  I can't imagine that anyone is going to wonder "what happened to my empty string?"...

src/utils/sharedFunctions.ts
```diff
@@ -43,6 +43,6 @@ export function formatString(word: string, replacement?: string[], capitalize =
   return word;
 }
 
-export function isEqual(array1, array2): boolean {
+export function isEqual(array1: Array<any>, array2: Array<any>): boolean {
```
EdwinGuzman: This is fine but you should comment the function with the style we use and specify the function is only to compare arrays.

src/components/ProtocolFormField.tsx
EdwinGuzman: Not a lot but I think these generic components would be better if they didn't depend on specific interfaces. We can change it if needed.

PR #274, opened 2019-11-05
Automatically search the book title on the list creator page

adriana-alter:
Resolves the problem I found on QA: the title doesn't get copied when you click the link to the list creator form.  (I am guessing this is CSRF-related.)

New implementation: when you go to the list creator form, it automatically runs a search for your book title.

### DIFF
--- a/src/components/CustomListsForBook.tsx
+++ b/src/components/CustomListsForBook.tsx
@@ -10,6 +10,7 @@ import {
 } from "../interfaces";
 import { FetchErrorData } from "opds-web-client/lib/interfaces";
 import { State } from "../reducers/index";
+import { Link } from "react-router";
 
 export interface CustomListsForBookStateProps {
   allCustomLists?: CustomListData[];
@@ -52,7 +53,6 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
     this.refresh = this.refresh.bind(this);
     this.save = this.save.bind(this);
     this.makeURL = this.makeURL.bind(this);
-    this.copyTitle = this.copyTitle.bind(this);
   }
 
   render(): JSX.Element {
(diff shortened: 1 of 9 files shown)


src/components/CustomListsForBook.tsx
```diff
@@ -82,17 +82,15 @@ export class CustomListsForBook extends React.Component<CustomListsForBookProps,
     // Link to the form for creating a new list
     return(
       <div key="list-creator-link">
-        <a onClick={this.copyTitle} href={`/admin/web/lists/${this.props.library}/create`}>Create a new list</a>
-        <p>(The book title will be copied to the clipboard so that you can easily search for it on the list creator page.)</p>
+        <Link to={{
+          pathname: `/admin/web/lists/${this.props.library}/create`,
+          state: { bookTitle: this.props.book && this.props.book.title }
+        }}>Create a new list</Link>
+        <p>(The book title will be automatically copied and searched on the list creator page.)</p>
```
EdwinGuzman: Is it still copied or just searched for?

EdwinGuzman:
I think it's fine but maybe running a search automatically is bit jarring. The note you added is great though because it gives the user a heads up.

src/components/CustomListsForBook.tsx
adriana-alter: It's used to populate the search field, but it isn't copied to the clipboard; that's the part that doesn't work on QA :(

EdwinGuzman: Right, so the text shouldn't say "automatically copied" then.

adriana-alter: Ah.  I just meant "copied" in the sense of "a copy shows up in the search field"; I don't think anyone is going to assume this literally means "copied to clipboard" and then be surprised and distressed when it isn't there...right?

PR #275, opened 2019-11-06
Adding react-axe and adding a script to run it.

EdwinGuzman:
Resolves [SIMPLY-2376](https://jira.nypl.org/browse/SIMPLY-2376). Similar to the registry admin PR, it should only be run when wanting to test/fix accessibility issues but not active development.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -52,6 +52,14 @@ class CirculationWeb {
     let customListPagePath = "/admin/web/lists(/:library)(/:editOrCreate)(/:identifier)";
     let lanePagePath = "/admin/web/lanes(/:library)(/:editOrCreate)(/:identifier)";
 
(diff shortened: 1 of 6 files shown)


src/index.tsx
```diff
@@ -52,6 +52,14 @@ class CirculationWeb {
     let customListPagePath = "/admin/web/lists(/:library)(/:editOrCreate)(/:identifier)";
     let lanePagePath = "/admin/web/lanes(/:library)(/:editOrCreate)(/:identifier)";
 
+    // `react-axe` should only run in development and testing mode.
+    // Running this is resource intensive and should only be used to test
+    // for accessibility and not while active development.
```
adriana-alter: Same comment as in registry_admin PR: "while actively developing" or "during active development".

PR #278, opened 2019-11-15
React axe fixes

EdwinGuzman:
This partly resolves [SIMPLY-2385](https://jira.nypl.org/browse/SIMPLY-2385). Partly because I didn't want to make the PR too big and there are some issues that need more thought to fix.

Some that come to mind:
1) The SVG icons in the List page (the pyramids) all have the same ids. It may be better to just not have ids or pass in unique ids somehow.
2) Similar "duplicate id" issue for panels. If you go to the Troubleshooting page, for example, a lot of panels have ids of "undefined-panel". I thought it could be that we are simply not passing in an `id` prop. So, I tried passing in the name of the collection since that's what is being displayed, but then you have multiple instances of `axis-360-panel`. Even if you add an index, I kept seeing multiple `axis-360-0-panel`, `axis-360-1-panel`, etc. Any ideas would be helpful.
3) Axe keeps saying that the `ul.nav.nav-tabs` is not right since Bootstrap (it seems) adds `::before` and `::after` css pseudo elements. Not sure how to fix that unless if we step away from bootstrap but it will change a lot of the interface.

As for this PR, it depends on the [opds PR 242](https://github.com/NYPL-Simplified/opds-web-client/pull/242). So some issues are fixed there that pop up here. There's more information about which issues have been completed in the spreadsheet: https://docs.google.com/spreadsheets/d/1dWq1NLukCYJJx70id0V2wNst4Gjm_q6fRJZTRJUTVh0/edit#gid=1557754872
It's mostly adding aria-labels to forms, updating headings, and fixing some duplicate id issues.


src/components/BookEditForm.tsx
```diff
@@ -44,13 +44,16 @@ export default class BookEditForm extends React.Component<BookEditFormProps, Boo
   }
 
   renderTextField(name: string, placeholder?: string, value?: string, hasLabel = true): JSX.Element {
+    const formattedName = formatString(name);
+    const ariaLabel = !hasLabel ? `Field for ${formattedName}` : null;
```
adriana-alter: Why would there be a situation in which we would want the aria-label to be null?  Isn't `hasLabel` just about whether a label gets visually rendered?

src/components/ConfigPage.tsx
```diff
@@ -33,14 +33,16 @@ export default class ConfigPage extends React.Component<ConfigPageProps, {}> {
     return (
       <div className="config">
         <Header />
```
adriana-alter: I get why it's like this, but it seems bad to have `<main>` inside `<div>`.  Should the whole thing be the `<main>`, or would that be worse because the header and footer would be in there?

src/components/ClassificationsForm.tsx
```diff
@@ -90,6 +90,7 @@ export default class ClassificationsForm extends React.Component<Classifications
                       type="text"
                       disabled={this.props.disabled}
                       name="target_age_min"
+                      aria-label="Target age minimum"
```
adriana-alter: This and the other target age aria-label be in the imperative--i.e. "Enter a minimum target age"--for consistency with the aria-labels on input fields elsewhere.

src/components/ClassificationsTable.tsx
```diff
@@ -22,8 +22,8 @@ export default class ClassificationsTable extends React.Component<Classification
             </tr>
           </thead>
           <tbody>
-            { this.props.classifications.map(classification =>
-              <tr key={classification.source + ":" + classification.name}>
+            { this.props.classifications.map((classification, idx) =>
+              <tr key={classification.source + "-" + idx}>
```
adriana-alter: Why the change?  React specifically advises against using the idx as a key except as a last resort.  (I know the key here doesn't consist just of the idx...but still.)

src/components/CustomListEditor.tsx
```diff
@@ -81,6 +81,8 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
               ) : null
             }
             <input
+              placeholder="Harry Potter"
```
adriana-alter: Did you mean to leave this in...?  (I mean, if we need a placeholder, this is a good one.  But it seems like a way to confuse admins.)

src/components/CustomListEntriesEditor.tsx
```diff
@@ -190,17 +190,17 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
                 </div>
               }
             </div>
+            <p>Drag search results here to add them to the list.</p>
```
adriana-alter: Any particular reason for moving this?  (I don't have a problem with it; just curious.)

src/components/DiagnosticsServiceTabs.tsx
```diff
@@ -34,14 +34,17 @@ export default class DiagnosticsServiceTabs extends TabContainer<DiagnosticsServ
     // If any of the timestamps has an exception, display a warning.  Otherwise, display the number of timestamps.
     let badge = hasException ?
         <span className="badge danger">!</span> :
-        <ToolTip trigger={<span className="badge">{timestampArray.length}</span>} text="Total number of timestamps for this service"/>;
+        <ToolTip
```
adriana-alter: Extra space :(

src/components/DiagnosticsServiceTabs.tsx
```diff
@@ -34,14 +34,17 @@ export default class DiagnosticsServiceTabs extends TabContainer<DiagnosticsServ
...
+          text="Total number of timestamps for this service"
+        />;
 
     return <section><span className="service-name">{super.tabDisplayName(name)}</span>{badge}</section>;
   }
 
   renderCollections(collections: Array<DiagnosticsCollectionData>) {
     // If the collection has any timestamps with exceptions, it should start out expanded.
-    return Object.keys(collections).map((collectionName) =>
+    return Object.keys(collections).map((collectionName, idx) =>
```
adriana-alter: Previous question about using idx in keys holds for this too (and everywhere else it's happening!).

src/components/EditableInput.tsx
```diff
@@ -38,75 +39,83 @@ export default class EditableInput extends React.Component<EditableInputProps, E
   }
 
   render() {
+    const {
+      type, elementType, optionalText, required, description, clientError, error,
+      label, extraContent,
+    } = this.props;
```
adriana-alter: Very nice!

src/components/GenreForm.tsx
```diff
@@ -44,9 +44,11 @@ export default class GenreForm extends React.Component<GenreFormProps, GenreForm
         <div className="form-inline">
           <select
             name="genre"
+            aria-label="Select table for genres"
```
adriana-alter: This is kind of a confusing way to phrase that...

src/components/BookEditForm.tsx
EdwinGuzman: Yes, `hasLabel` is whether there is a visual label or not. If there is no visual label, the "fallback" would be to render an `aria-label` attribute. If it's null it just won't appear and that's okay because a visual label will be rendered.

src/components/SaveButton.tsx
```diff
@@ -4,7 +4,7 @@ export interface SaveButtonProps {
   disabled: boolean;
   submit?: any;
   text?: string;
-  type?: string;
+  type?: "submit" | "reset" | "button";
 }
```
adriana-alter: Nice!

src/components/ClassificationsForm.tsx
EdwinGuzman: Good point.

src/components/Stats.tsx
```diff
@@ -46,7 +46,7 @@ export class Stats extends React.Component<StatsProps, {}> {
     const totalStats = this.props.isLoaded && this.props.libraries && this.props.libraries.length > 1 && this.props.stats && this.props.stats["total"];
 
     return (
-      <div>
+      <>
```
adriana-alter: But...why?  Fragments seem less accessible...

src/components/TextWithEditMode.tsx
```diff
@@ -7,6 +7,7 @@ export interface TextWithEditModeProps extends React.Props<TextWithEditMode> {
   text?: string;
   placeholder: string;
   onUpdate?: (text: string) => void;
+  "aria-label"?: string;
```
adriana-alter: Why is this optional?

src/components/ClassificationsTable.tsx
EdwinGuzman: `classification.name` can be undefined. If it is, there can be multiple `${classification.source}:undefined` keys.

src/components/ConfigPage.tsx
EdwinGuzman: Right, the whole thing cannot be `<main>` because it shouldn't include the header or the footer. Technically it should be legal to do, but because there's only one header and because we don't intend to have sub-headers, then it will complain. I think a better change will be updating the whole app so that the header and footer always render and only the main content gets updated based on the route.

src/components/CustomListEditor.tsx
EdwinGuzman: I may have thought it was a test. "Book title" is better. _deletrius_

src/components/CustomListEntriesEditor.tsx
EdwinGuzman: `p` element shouldn't be in a `ul` element.

src/components/DiagnosticsServiceTabs.tsx
EdwinGuzman: Hm?

src/components/DiagnosticsServiceTabs.tsx
EdwinGuzman: Oh, I forgot to remove it. This is a case where the `Panel` element here needs to have an `id` prop. Currently, all the panels will have an id of `undefined-panel`. Using the index here doesn't help.

src/components/ClassificationsTable.tsx
adriana-alter: Ah.  That's bad.  How about telling it to use the name if there is one, and the idx if not?  Or is life too short for this?

src/components/EditableInput.tsx
EdwinGuzman: This was actually accidental. Once I finished updating this, I realized I wanted to update a different function, but left it because it's slightly more readable.

src/components/GenreForm.tsx
EdwinGuzman: True...

src/components/DiagnosticsServiceTabs.tsx
adriana-alter: Blank line after line 37.  Or is Github just being weird?

src/components/SelfTestsCategory.tsx
```diff
@@ -26,7 +26,18 @@ export default class SelfTestsCategory extends React.Component<SelfTestsCategory
     return (
       <div className="self-tests-category has-additional-content">
         <ul>
-          { this.props.items && this.props.items.map(i => <Panel style={getClassName(i)} key={i.name} openByDefault={onlyChild} headerText={i.name} content={[link(i), selfTests(i)]} />)}
+          { this.props.items && this.props.items.map((item, idx) =>
+              <li>
+                <Panel
+                  style={getClassName(item)}
+                  key={item.name}
+                  id={`${item.name.replace(/\s+/g, "")}-${idx}`}
```
EdwinGuzman: I should probably revert this as well and then fix all the Panel instances in one PR.

src/components/Timestamp.tsx
```diff
@@ -28,6 +28,7 @@ export default class Timestamp extends React.Component<TimestampProps, {}> {
     // If the timestamp has an exception, it should start out expanded.
     return (
       <Panel
+        id="timestamp"
```
EdwinGuzman: But some `id` props in some instances are okay and won't be duplicated.

src/components/SelfTestsCategory.tsx
adriana-alter: This sounds like a good idea.

src/components/ClassificationsTable.tsx
EdwinGuzman: Yeah, not bad to do.

src/components/DiagnosticsServiceTabs.tsx
EdwinGuzman: Ah, that should be Github.

src/components/SaveButton.tsx
EdwinGuzman: Yeah, I'm not sure why this issue didn't come up before 🤷‍♂  🤷‍♂

src/components/Stats.tsx
EdwinGuzman: Fragments don't render anything. This is to reduce all the divs that get rendered. In this case, it doesn't add anything and seeing so many nested divs is confusing.

src/components/TextWithEditMode.tsx
EdwinGuzman: That can change. 👍

EdwinGuzman:
I'm going to leave this open until other accessibility fixes are in to bundle them all together.

PR #279, opened 2019-11-15
Test fixes node 10

EdwinGuzman:
Resolves [SIMPLY-2412](https://jira.nypl.org/browse/SIMPLY-2412).

Minor update to sinon but major updates to chai and mocha. That update doens't actually fix anything but figured it'd be good to update now. A big change is updating the Node version in Travis to use Node v10.

Wrote some comments in the updated tests (`CustomList-test`). It seems that fully mounting a component didn't allow the `window.confirm` to work. I think it works in _one_ case but every other case uses shallow. But, the tests need to fully mount or else the `button` which is clicked is unavailable. So, mocked this by directly calling the component instance's function with different list data.


src/components/__tests__/CustomLists-test.tsx
```diff
@@ -82,386 +81,439 @@ describe("CustomLists", () => {
...
-      />,
-      { context: { admin: librarian }}
-    );
-
-    let entryPoints = wrapper.instance().getEnabledEntryPoints(libraries);
-
-    expect(entryPoints.length).to.equal(1);
-    expect(entryPoints).to.eql(["Audio"]);
+  // These tests are for mocking the confirm function in the window object.
+  // It relying works when shallow mounting a component but not when doing a
```
adriana-alter: Typo in comment; supposed to say "reliably"?

adriana-alter:
This is really nice!  Just fix the typo in the comment, and you're all set.

src/components/__tests__/CustomLists-test.tsx
EdwinGuzman: Oops, yes.

PR #280, opened 2019-11-18
Opds 0.3.0

EdwinGuzman:
Resolves [SIMPLYE-2415](https://jira.nypl.org/browse/SIMPLY-2415).

Updating the `opds-web-client` to 0.3.0, updating Typescript, and updating related packages.


src/components/AccountPage.tsx
```diff
@@ -16,7 +16,7 @@ export default class AccountPage extends React.Component<{}, {}> {
   context: AccountPageContext;
 
   static contextTypes: React.ValidationMap<AccountPageContext> = {
-    editorStore: PropTypes.object.isRequired,
+    editorStore: PropTypes.object.isRequired as React.Validator<Store>,
```
EdwinGuzman: These are required now... it seems like Typescript doesn't know the type of `object` that is required even though it was declared in the context type. So it needs to be more explicit and validated. There are other occasions where this happens with other interfaces.

src/components/BookCoverEditor.tsx
```diff
@@ -66,10 +66,10 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
         </h2>
         { this.props.isFetching &&
           <div className="cover-fetching-container">
-            <h4>
+            <h3>
```
EdwinGuzman: This makes this text slightly larger but I don't think it's too jarring. This pattern occurs multiple places and it would be better to separate it into its own component and with updated styling.

src/components/CirculationEventsDownloadForm.tsx
```diff
@@ -103,6 +103,7 @@ export default class CirculationEventsDownloadForm extends React.Component<Circu
               ref={this.type}
               name="type"
               onChange={this.toggleLocationsInput}
+              onBlur={this.toggleLocationsInput}
```
EdwinGuzman: `tslint-react-a11y` was updated and now it required appropriate input elements to not depend on just `onChange`. It's an accessibility issue but don't know the full idea behind it.

PR #281, opened 2019-11-19
Updating components that had missing or duplicate ids, and added a co…

EdwinGuzman:
…mponent to make the updating loading message consistent.

Continuation of PR #278 . This mostly fixes the missing or duplicate id issue in the `Panel` component and updates all the SVG icons to be consistent and have an `aria-hidden` attribute and not an `aria-describedby` or `aria-labelledby` attribute.


package.json
```diff
@@ -30,7 +30,7 @@
     "@nypl/dgx-svg-icons": "0.3.4",
     "bootstrap": "^3.3.6",
     "draft-convert": "^2.1.5",
-    "draft-js": "^0.11.0",
+    "draft-js": "0.11.2",
```
EdwinGuzman: There is one aria attribute issue coming from draftjs but there's no solution, it seems. It's an open issue in their repo and trying to pass or add a manual label doesn't work. One of the children components has an `aria-describedby` attribute but there's no element with the id to describe it 🤷‍♂

src/components/BookCoverEditor.tsx
```diff
@@ -64,14 +65,9 @@ export class BookCoverEditor extends React.Component<BookCoverEditorProps, {}> {
...
-        { this.props.isFetching &&
-          <div className="cover-fetching-container">
-            <h4>
-              Updating
-              <i className="fa fa-spinner fa-spin"></i>
-            </h4>
-          </div>
-        }
+
+        <UpdatingLoader show={this.props.isFetching} />
```
EdwinGuzman: Made this a component to make this all consistent.

src/components/CustomListEditor.tsx
```diff
@@ -132,7 +132,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
             <div className="custom-list-filters">
               <Panel
                 headerText="Add from collections"
-                id="add-from-collections-panel"
```
EdwinGuzman: The `Panel` component already adds `-panel` to the end.

src/components/DiagnosticsServiceTabs.tsx
```diff
@@ -43,19 +44,23 @@ export default class DiagnosticsServiceTabs extends TabContainer<DiagnosticsServ
...
+          openByDefault={collectionToRender.some((ts) => ts.exception)}
+          content={this.renderTimestamps(collectionToRender)}
+        />
+      );
+    });
   }
 
   renderTimestamps(timestamps: Array<TimestampData>): JSX.Element {
     let tsList = timestamps.map(timestamp =>
-      <li className="timestamp-holder" key={timestamp.toString()}>
```
EdwinGuzman: `timestamp` is an object so this threw an error.

src/components/icons/AddIcon.tsx
```diff
@@ -1,12 +1,10 @@
...
-  render(): JSX.Element {
-    return (
-      <svg xmlns="http://www.w3.org/2000/svg" width="48.01772" height="48.01658" viewBox="0 0 48.01772 48.01658" aria-hidden="true">
-        <title>Add Icon</title>
-        <path d="M26.98371,20.9161V2.97433a2.97433,2.97433,0,0,0-5.94867,0v17.903l-18.04011-.1176a2.97542,2.97542,0,0,0-.039,5.95072l18.07914.11708V45.04224a2.97433,2.97433,0,1,0,5.94867,0V26.86579l18.03908.11683a2.9744,2.9744,0,0,0,.04108-5.94867Z"/>
-      </svg>
-    );
-  }
-}
+const AddIcon = (): JSX.Element => (
```
EdwinGuzman: Also converted all of these SVG components from React classes to functions. Not necessary but we don't use state for these components so it's fine.

src/components/BookCoverEditor.tsx
EdwinGuzman: Another issue is that the `h4` sometimes made for inconsistent headings. On this page it goes from h2, h4, h3 and that's an accessibility issue.

package.json
adriana-alter: Eek.  How likely is this to actually affect users in any meaningful way (not that I have a back-up plan...)?

src/components/ClassificationsForm.tsx
```diff
@@ -56,6 +56,7 @@ export default class ClassificationsForm extends React.Component<Classifications
         }
 
         <Panel
+          id="classifications-form"
```
adriana-alter: It seems confusing to call a panel "form" (even though Panel will tack `-panel` on).  I assume the panel just contains a form, rather than being used as a form itself?  Any reason not to just go "classifications-panel"?

src/components/DiagnosticsServiceTabs.tsx
```diff
@@ -43,19 +43,23 @@ export default class DiagnosticsServiceTabs extends TabContainer<DiagnosticsServ
...
-        headerText={collectionName}
-        openByDefault={collections[collectionName].some((ts) => ts.exception)}
-        content={this.renderTimestamps(collections[collectionName])}
-      />
-    );
+    return Object.keys(collections).map((collectionName) => {
+      const collectionToRender = collections[collectionName];
+      return (
+        <Panel
+          id={`diagnostics-${collectionToRender[0].id}`}
```
adriana-alter: Why would we want the id from `collectionToRender[0]`; isn't that just a timestamp...?  I think using `collectionName` would be better.

src/components/SelfTestsCategory.tsx
```diff
@@ -26,15 +26,17 @@ export default class SelfTestsCategory extends React.Component<SelfTestsCategory
     return (
       <div className="self-tests-category has-additional-content">
         <ul>
-          { this.props.items && this.props.items.map((item) =>
-              <li key={item.name}>
+          { this.props.items && this.props.items.map((item) => {
+              return <li key={item.name}>
```
adriana-alter: Why the change to curly braces and `return`--just stylistic preference, or was something not getting returned correctly?

src/components/Timestamp.tsx
```diff
@@ -8,18 +8,19 @@ export interface TimestampProps {
 
 export default class Timestamp extends React.Component<TimestampProps, {}> {
   render(): JSX.Element {
-    let exception = <section className="well exception"><pre>{this.props.timestamp.exception}</pre></section>;
-    let achievements = <section className="well"><pre>{this.props.timestamp.achievements}</pre></section>;
+    const { timestamp } = this.props;
```
adriana-alter: Any reason not to go `const timestamp = this.props.timestamp`?  (I don't have any objection to this; just curious.)

src/components/UpdatingLoader.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from "react";
+
+export interface LoadButtonProps {
+  text?: string;
+  show: boolean;
+}
+
+function UpdatingLoader(props: LoadButtonProps) {
```
adriana-alter: Any reason to do this as a functional component?  Given that we don't have functional components anywhere else (off the top of my head, at least!), I'd rather keep things consistent.  If there's an argument to be made for having this be a functional component, then I imagine there are other similar components that we should convert too, but that would be a whole other PR...

src/components/__tests__/UpdatingLoader-test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    );
+
+    const container = wrapper.find(".updating-loader-container");
+    const loader = wrapper.find(".updating-loader");
+    expect(container.length).to.equal(1);
+    expect(loader.length).to.equal(0);
+  });
+
+  it("should render a container and the updating message", () => {
+    const wrapper = mount(
```
adriana-alter: No test for rendering the spinner? :(

package.json
EdwinGuzman: Not very likely but something we should keep aware of in their repo https://github.com/facebook/draft-js/issues/1739

src/components/ClassificationsForm.tsx
EdwinGuzman: That's true but not all panels have forms. I was trying to go more with the name of the component but didn't want to do camel case. "classifications-panel" is fine.

src/components/DiagnosticsServiceTabs.tsx
EdwinGuzman: No, the id isn't a timestamp. The problem with using `collectionName` is that you can still have multiple `diagnostics-axis-360-panel` on the page.

src/components/SelfTestsCategory.tsx
EdwinGuzman: You're right. I forgot to revert this.

src/components/Timestamp.tsx
EdwinGuzman: I guess just a habit of destructuring even if it's one variable.

src/components/UpdatingLoader.tsx
EdwinGuzman: Yes, I think if we can go more of the functional component route (which is what the latest React suggestion is), it'd be nice but not necessary. They have gone from functional to class and now back to functional (but hooks seems nice to use). I don't think there's an urgent need to convert components that _can_ be converted if it's out of the way. But if you want to refactor any components to use this pattern and it makes sense (i.e. components that don't have state), then I'm for it.

src/components/__tests__/UpdatingLoader-test.tsx
EdwinGuzman: :imp:

PR #282, opened 2019-11-26
Updating the context so that pathFor is supported.

EdwinGuzman:
This fixes the issue found when trying to import the `logic-components` branch from the `opds-web-client` into this app. This was an issue raised in PR [opds-web-client #248](https://github.com/NYPL-Simplified/opds-web-client/pull/248). I haven't updated the version or changelog, but will do so after some feedback. It seems this is the minimum required to make everything work.

### DIFF
--- a/src/components/__tests__/ContextProvider-test.tsx
+++ b/src/components/__tests__/ContextProvider-test.tsx
@@ -1,16 +1,20 @@
 import { expect } from "chai";
 
 import * as React from "react";
-import { shallow } from "enzyme";
+import { shallow, mount } from "enzyme";
 import * as jsdom from "jsdom";
 
 import ContextProvider from "../ContextProvider";
 import Admin from "../../models/Admin";
+import * as PropTypes from "prop-types";
+import { stub } from "sinon";
 
-class FakeChild extends React.Component<any, any> {}
+class FakeChild extends React.Component<{}, {}> {}
 
 describe("ContextProvider", () => {
   let wrapper;
+  let instance;
+  let pathFor;
 
   beforeEach(() => {
     wrapper = shallow(
@@ -18,17 +22,18 @@ describe("ContextProvider", () => {
         csrfToken="token"
         roles={ [{ "role": "system" }] }
         email="email"
-        >
+      >
         <FakeChild />
       </ContextProvider>
     );
+    instance = wrapper.instance();
+    pathFor = instance.pathFor;
   });
 
   it("provides child context", () => {
-    let context = wrapper.instance().getChildContext();
+    let context = instance.getChildContext();
     expect(context.editorStore.getState().editor).to.be.ok;
     expect(context.editorStore.getState().catalog).to.be.ok;
-    expect(context.pathFor).to.equal(wrapper.instance().pathFor);
     expect(context.csrfToken).to.equal("token");
     expect(context.settingUp).not.to.be.ok;
     expect(context.admin instanceof Admin).to.be.ok;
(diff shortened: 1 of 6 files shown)


adriana-alter:
This seems very reasonable.  Update the tests to match, and I'll be better able to see what's going on.

EdwinGuzman:
Oh hmmm the tests won't pass until the `opds-web-client` is published first. So waiting on that (but they do pass locally).

adriana-alter:
No, I meant, can you write tests to reflect that there's now supposed to be the PathForProvider wrapper?

EdwinGuzman:
I see, yeah.

@kristojorg It seems that the `OPDSCatalog` component always needs the pathFor provider. If that's the case, might it be better to include it and call it inside of `OPDSCatalog` instead of wrapping it like below (app.tsx in `opds-web-client`) or is there another reason for separating it?
```
<PathForProvider pathFor={config.pathFor}>
  <OPDSCatalog {...mergedProps} />
</PathForProvider>
```
That way, that same pattern doesn't have to be written in `circulation-web` here.

kristojorg:
Ok I see where I made a mistake. Basically, `pathFor` is always provided by the application wrapping `opds-web-client`. What I did in my PR was to create the `PathForProvider` which wrapping applications could use to provide the `pathFor` via both legacy and new context APIs. The mistake I made is that internally, `opds-web-client` renders `StoreContextProvider` which I also created and is similar to `PathForProvider` in that it creates the store and passes it down via old and new APIs. 

The issue is that `StoreContextProvider` requires the `PathFor` context, and previously it had retrieved it from the legacy API, but I instead switched the component to retrieve it from the new API. I forgot, however, that `circulation-web` would only be providing it with the legacy. 

In any case, the solution is simple: just use the `PathForProvider` I created in `opds-web-client` to pass it down the tree, instead of relying on the old API. This is similar to what you've done @EdwinGuzman , but instead of retrieving pathFor in the `CatalogPage` via the old API and then passing it in to `PathForProvider`, we can just use `PathForProvider` in the top level `ContextProvider` to pass it down via both legacy and new API. I'll submit a PR doing this, which should fix it and be cleaner. 

Hopefully this answers your question about rendering `PathForProvider` within `OPDSCatalog`... the value is supposed to be provided by a wrapping application, so doing it within `OPDSCatalog` wouldn't really allow for that.

EdwinGuzman:
@kristojorg thanks for the other PR and helping me fix how the PathForProvider should be used. I updated this PR with the updated `opds-web-client` package, and updated some tests, @adriana-alter .

src/components/__tests__/ContextProvider-test.tsx
```diff
@@ -41,64 +46,98 @@ describe("ContextProvider", () => {
     expect(children.length).to.equal(1);
   });
 
-  describe("pathFor", () => {
+  it("should pass down the pathFor function down to the child as context", () => {
+    const hasAccess = "has access to pathFor context";
+    let pathForStub = stub();
+    class MockContextProvider extends ContextProvider {
```
EdwinGuzman: @kristojorg I used a similar pattern from your `opds-web-client` PR to test the provider, but I extended this class here so I could stub and make sure the `pathFor` class method is testable and the same.

kristojorg: Ya interesting way of doing it. I'm a bit confused on the advantage of it, would love an explanation so I know for moving forward

EdwinGuzman: Not sure if there's an advantage but it was easier to extend and stub the class method that will get passed down to `Child` and test the text that gets rendered, rather the way it was done in `opds-web-client`. That's mostly because `pathFor` is a prop for the `PathForProvider`, but in this case it's a class method. That made testing `this.context?.pathFor === pathFor` difficult.

PR #283, opened 2019-12-02
added PathForProvider to ContextProvider

kristojorg:
This PR is a different approach to fixing the same problem as in https://github.com/NYPL-Simplified/circulation-web/pull/282. In that PR, the `pathFor` was read out of context in `CatalogPage`, and then passed in to `PathForProvider`, which in turn provides `pathFor` in both the legacy and new context APIs. That's a bit duplicative, so instead I just used the `PathForProvider` inside of the top level `ContextProvider` where the pathFor context was initially defined. 

I removed code from the `ContextProvider` which added `pathFor` via legacy context API because this is now handled by the `PathForProvider`.

Note I was not able to run this code because I don't have the proper setup to run `circulation-web`, so @EdwinGuzman maybe you can take over from here?

### DIFF
--- a/src/components/ContextProvider.tsx
+++ b/src/components/ContextProvider.tsx
@@ -5,6 +5,7 @@ import buildStore from "../store";
 import { PathFor } from "../interfaces";
 import { State } from "../reducers/index";
 import Admin from "../models/Admin";
+import PathForProvider from "opds-web-client/lib/components/context/PathForContext";
 
 export interface ContextProviderProps extends React.Props<any> {
   csrfToken: string;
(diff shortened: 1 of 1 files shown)


EdwinGuzman:
Yes, I'll try it out. Do you need to update any tests for this change?

kristojorg:
@EdwinGuzman I honestly am not sure, I more meant for this to be a jumping off point for you. So long as you are testing `ContextProvider` to see if it is really providing these values, I think you're good. You could add a test to look for `pathFor` via the new context API as well if you'd like

EdwinGuzman:
Closing this for #282 , thanks!

PR #284, opened 2019-12-12
Metadata self tests

adriana-alter:
Resolves [2420](https://jira.nypl.org/browse/SIMPLY-2420); adding a metadata tab to the self-tests section on the Troubleshooting page.

Goes with these [circulation](https://github.com/NYPL-Simplified/circulation/pull/1366) and [core](https://github.com/NYPL-Simplified/server_core/pull/1144) branches (also called `metadata-self-tests`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.4.3
+#### Added
+- Added a Metadata Services tab to the self-tests section of the Troubleshooting page.
+
 ### v0.4.2
 #### Updated
 - Updated the `opds-web-client` package to v0.3.2.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 #### Added
 - Added a Metadata Services tab to the self-tests section of the Troubleshooting page.
 
-### v0.4.2
+### v0.4.2`
 #### Updated
 - Updated the `opds-web-client` package to v0.3.2.
 - Updated the ContextProvider based on the `opds-web-client`'s updated context APIs.
@@ -22,7 +22,7 @@
 #### Updated
 - Updated npm packages including chai, mocha, and sinon.
 - Updated version of Node to run in Travis to Node v10.
-#### Fixed
+#### Fixed``
 - Fixed tests that were passing in Node v8 but failing in Node v10. Specifically, stubbing window.confirm did not work on components that were fully mounted. A shallow mount helped fix this issue, but the tests needed to be updated accordingly.
 
 ### v0.3.26
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 #### Added
 - Added a Metadata Services tab to the self-tests section of the Troubleshooting page.
 
-### v0.4.2`
+### v0.4.2
 #### Updated
 - Updated the `opds-web-client` package to v0.3.2.
 - Updated the ContextProvider based on the `opds-web-client`'s updated context APIs.
@@ -22,7 +22,7 @@
 #### Updated
 - Updated npm packages including chai, mocha, and sinon.
 - Updated version of Node to run in Travis to Node v10.
-#### Fixed``
+#### Fixed
 - Fixed tests that were passing in Node v8 but failing in Node v10. Specifically, stubbing window.confirm did not work on components that were fully mounted. A shallow mount helped fix this issue, but the tests needed to be updated accordingly.
 
 ### v0.3.26
--- a/src/components/SelfTests.tsx
+++ b/src/components/SelfTests.tsx
@@ -40,6 +40,9 @@ export interface SelfTestsState {
 }
 
 export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
+  static defaultProps = {
+    sortByCollection: false
+  };
   constructor(props: SelfTestsProps) {
     super(props);
 
--- a/src/components/SelfTests.tsx
+++ b/src/components/SelfTests.tsx
@@ -28,6 +28,7 @@ export interface SelfTestsDispatchProps {
 export interface SelfTestsOwnProps {
   store?: Store<State>;
   type: string;
+  sortByCollection: boolean;
 }
 
 export interface SelfTestsProps extends React.Props<SelfTestsProps>, SelfTestsStateProps, SelfTestsDispatchProps, SelfTestsOwnProps {}
@@ -48,7 +49,7 @@ export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
       mostRecent: this.props.item
     };
     this.runSelfTests = this.runSelfTests.bind(this);
-    this.displayMetadata = this.displayMetadata.bind(this);
+    this.displayByCollection = this.displayByCollection.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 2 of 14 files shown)


src/components/__tests__/ContextProvider-test.tsx
```diff
@@ -61,7 +61,7 @@ describe("ContextProvider", () => {
         pathFor: PropTypes.func.isRequired
       };
       render() {
-        const hasContext = this.context?.pathFor === pathForStub;
```
adriana-alter: My linter doesn't like the `?` syntax; it was blocking me from running the tests.

src/stylesheets/self_tests.scss
```diff
@@ -53,6 +53,10 @@
           margin-left: auto;
         }
 
+        .panel-success > .panel-heading {
+          background: #809F69;
```
adriana-alter: There's an issue with getting nested panels to be the right color.  A better solution would be to change this in ReusableComponents, but a quick fix for now...

adriana-alter:
Resolves [2420](https://jira.nypl.org/browse/SIMPLY-2420); adding a metadata tab to the self-tests section on the Troubleshooting page.

Goes with these [circulation](https://github.com/NYPL-Simplified/circulation/pull/1366) and [core](https://github.com/NYPL-Simplified/server_core/pull/1144) branches (also called `metadata-self-tests`

src/components/__tests__/ContextProvider-test.tsx
EdwinGuzman: Did you remove the `node_modules` folder and reinstall everything?

EdwinGuzman:
I currently have four services: two NoveList Select, Metadata Wrangler, and New York Times. On the initial attempt to run tests for the Metadata Wrangler, all the services pick up the same failed Self Test Results. When the page is refreshed, only the appropriated service has the failed Self Tests Results (this is good). But the failed Self Test Results should not be copied to the other services (whether the Self Test Results failed or passed).

Initial:
Screen Shot 2019-12-12 at 4 52 47 PM
This is fine since I expect Metadata Wrangler tests to have failed results.

If "Run tests" for the Metadata Wrangler is hit, this happens. Notice that the "Novelist Api" panel is still grey and the Self Test Result is clearly not for the Novelist API service.
 
Screen Shot 2019-12-12 at 4 52 56 PM

EdwinGuzman:
The screenshot only shows the data copied over to one other service, but the other two services now shown do have the same and incorrect Self Test Results.

adriana-alter:
-- Fixed bug (turned out to be from missing parentheses!)
-- Put in a different message for services where the reason no self-tests were found is that self-tests have never been run on them before, not because something is wrong; i.e. if it's just a new service, don't display something that looks like a scary error message (screenshot is current behavior).
-- Slight refactoring so I could see what I was doing with putting in the message.

Screen Shot 2019-12-13 at 12 29 20 PM

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 ## Changelog
 
-### v0.4.2
+### v0.4.3
+#### Added
+- Added a Metadata Services tab to the self-tests section of the Troubleshooting page.
+
+### v0.4.2`
```
EdwinGuzman: Master has been updated and is now at `0.4.3` and that version hasn't been published yet, so this can be bundled together. You don't need the backtick after `0.4.2`.

CHANGELOG.md
```diff
@@ -18,7 +22,7 @@
 #### Updated
 - Updated npm packages including chai, mocha, and sinon.
 - Updated version of Node to run in Travis to Node v10.
-#### Fixed
+#### Fixed``
```
EdwinGuzman: More backticks?

src/components/SelfTests.tsx
```diff
@@ -120,13 +113,50 @@ export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
...
+    let sorted = {};
+    // The initial set-up test's collection property is always undefined
+    let isInitial = (c: string) => c === "undefined";
+    // Group the results by their collections
+    results.forEach(r => sorted[r.collection] = results.filter(result => result.collection === r.collection));
+    let style = (resultsForCollection: SelfTestsResult[]) => resultsForCollection.every(r => r.success) ? "success" : "danger";
+    collectionList = Object.keys(sorted).map((collection) => {
+      let resultsForCollection = sorted[collection];
+      return (
+        <Panel
```
EdwinGuzman: This is missing an `id` prop.

src/components/SelfTests.tsx
```diff
@@ -120,13 +113,50 @@ export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
...
+          key={collection}
+          headerText={isInitial(collection) ? "Initial Setup" : collection}
+          openByDefault={isInitial(collection)}
+          content={resultsForCollection.map((result: SelfTestsResult, idx: Number) => this.makeResult(result, idx, isFetching))}
```
EdwinGuzman: I'd make the map result into a new variable to make it more readable.

src/components/__tests__/SelfTests-test.tsx
```diff
@@ -295,4 +311,54 @@ describe("SelfTests", () => {
...
+    const baseResult = collections[0].self_test_results.results[0];
+    let results = [];
+    let makeResult = (c: string, success = true) => {
+      return {...baseResult, ...{
+        collection: c,
+        name: `Test ${results.length < collectionNames.length ? 1 : 2}`,
+        success: success
+      }};
+    };
+    let f = (results) => wrapper.instance().displayMetadata(results, false);
```
EdwinGuzman: This could use a better name than `f`

src/components/SelfTests.tsx
```diff
@@ -120,13 +113,50 @@ export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
...
+    return <SelfTestResult key={`${result.collection}-${idx}`} result={result} isFetching={isFetching} />;
+  }
+
+  displayMetadata(results: SelfTestsResult[], isFetching: boolean): JSX.Element[] {
+    let collectionList: JSX.Element[];
+    let sorted = {};
+    // The initial set-up test's collection property is always undefined
+    let isInitial = (c: string) => c === "undefined";
+    // Group the results by their collections
+    results.forEach(r => sorted[r.collection] = results.filter(result => result.collection === r.collection));
```
EdwinGuzman: When I `console.log(results)`, there is no `collection` property for every item in results. Is that right?

EdwinGuzman:
Can you give an example and screenshot of having multiple collections? It seems the biggest change is adding a panel component inside the `SelfTest` component to wrap around different collections in a metadata test, right?

I'm seeing these errors in the console on initial load:
Screen Shot 2019-12-16 at 11 17 20 AM

I'd to see `SelfTests` be refactored so that it does not know about `metadata` or `"Metadata Wrangler"`. Right now it is specifically branching out to render a specific DOM if that's the goal/protocol but if more of service that are similar are introduced, then it we'll have more checks for them. Could the `shouldSortByCollection` be done in the `SelfTests` parent? `displayMetadata` also seems like a function that can be used for more than just metadata, so a more generic name will be good to have.

src/components/SelfTests.tsx
adriana-alter: The first `results` item should have no `collection` property; this is because it's the initial setup test, which isn't specific to a collection.  That's the only one with no `collection`...right??

EdwinGuzman: Yeah, all the metadata services have an "Initial setup" as their first group of tests.

adriana-alter:
Still trying to figure out the console errors.

Note: the server-side stuff has been merged into master.

EdwinGuzman:
Very nice refactor. Do you also get the errors in the console or was that just me?

EdwinGuzman:
I'm guessing that now, because there is no first initial collection, that the sub-panels won't show up? I now see the same as before but the results are not in an "Initial Setup" panel.

adriana-alter:
I get the errors in the console, but not consistently, which is...interesting...

adriana-alter:
I've just discovered that I get the console errors for the other self test categories too.  I'm pretty sure it's a server-side issue.  In any event, it's a not-this-branch issue.

Yes, that's why no sub-panels for the NYT integration.

EdwinGuzman:
😰

PR #285, opened 2019-12-12
Updating and adding a default string to the DraftJS editor when an empty value is passed.

EdwinGuzman:
Resolves [SIMPLY-2439](https://jira.nypl.org/browse/SIMPLY-2439).

On error in the console that appears is misleading in production. In local development, it was an error with the not being able to get the `key` for some value. Not sure exactly how DraftJs tries to create an editor element, but having an empty summary breaks when trying to parse the string to pass it to `ContentState.createFromBlockArray` from `blocksFromHTML.contentBlocks`.

There are two scenarios this seems to happen:
1) A book already has no summary and the page breaks
2) An admin removes the summary from a book and saves the editor 

The second one is the one I tested and found that DraftJS adds an automatic `"<p></p>"` string as a default so tested against that and the empty string too.

There is now a default string that gets rendered from `EditorField`, but in `BookEditForm`, it does not get saved. So the content that appears will only serve as a reminder to admins to update the content. Maybe that should be more explicit and say something like `"Update the summary for this book. This content will not be saved."`?

### DIFF
--- a/src/components/__tests__/BookEditForm-test.tsx
+++ b/src/components/__tests__/BookEditForm-test.tsx
@@ -2,7 +2,7 @@ import { expect } from "chai";
 import { stub } from "sinon";
 
 import * as React from "react";
-import { shallow, mount } from "enzyme";
+import { mount } from "enzyme";
 
 import BookEditForm from "../BookEditForm";
 import EditableInput from "../EditableInput";
@@ -307,8 +307,11 @@ describe("BookEditForm", () => {
     expect(editBook.args[0][1].get("summary")).to.contain(bookData.summary);
   });
 
-  it("refreshes book after editing", (done) => {
+  it("refreshes book after editing", async () => {
     let editBook = stub().returns(new Promise((resolve, reject) => {
+      resolve();
+    }));
+    let refreshStub = stub().returns(new Promise((resolve, reject) => {
       resolve();
     }));
     wrapper = mount(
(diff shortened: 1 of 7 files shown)


src/components/__tests__/BookEditForm-test.tsx
```diff
@@ -307,24 +307,56 @@ describe("BookEditForm", () => {
...
-        refresh={done}
+        refresh={refreshStub}
         editBook={editBook}
       />
     );
 
     let form = wrapper.find(Form);
-    form.prop("onSubmit")(new (window as any).FormData(form.getDOMNode()));
+    await form.prop("onSubmit")(new (window as any).FormData(form.getDOMNode()));
+    expect(refreshStub.callCount).to.equal(1);
```
EdwinGuzman: It seems this was an empty test without actually testing anything. Added an expect.

aslagle:
Thanks, I had the same problem https://jira.nypl.org/browse/SIMPLY-2435

src/testHelper.ts
```diff
@@ -10,6 +10,7 @@ const win = doc.defaultView;
 
 global["document"] = doc;
 global["window"] = win;
+global["HTMLElement"] = win.HTMLElement;
```
EdwinGuzman: DraftJS complained about this. It seems like something internal to the package when trying to write specific tests.

EdwinGuzman:
@aslagle  thanks, I'll link the tickets together.

adriana-alter:
What do you mean about reminding admins to update the content?  The editor doesn't erase the existing summary or anything...right?  (And if you go to the edit page but don't change anything, then you shouldn't need to be told that there aren't any changes.) . Or do you mean there's an issue where blank summaries on the details tab are rendered differently from blank summaries in the editor field (if so, I can't reproduce it)?  Or...?

EdwinGuzman:
The editor does not erase the existing summary. If you go into the edit tab and remove the summary content, when you refresh the page the `EditorField` component throws an error because DraftJS's functions can't handle empty strings or undefined string. The empty strings (both `""` and `"<p></p>"`) crash in `convertFromHTML`. Somehow, I'm guessing in run-time, `undefined` is passed as the `content` prop and that crashes too (not sure how since it's suppose to only accept strings and would normally not compile and crash before that).

For my comment, I was talking about this:

Screen Shot 2019-12-12 at 4 32 12 PM

If the summary was intentionally removed, not updated, and saved, then this will popup next time. Same case if there was no summary to begin with. Besides serving as a reminder to admins, it's also to allow the DraftJS functions to work properly.

adriana-alter:
Ohhh.  Yes, placeholder text is a good idea!

PR #288, opened 2019-12-16
List page scroll

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2426 (footer on list manager page blocks content)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### v0.4.3
 #### Fixed
+- Fixed a display bug on the List Manager page.
 - Fixed the display color for the Terms & Conditions link in the footer.
 - Fixed a bug where an empty book summary caused creating a DraftJS instance to crash the app.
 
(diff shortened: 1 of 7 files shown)


EdwinGuzman:
This seems fine overall but if you have 50 items in a list, then the page is very long. The scroll box before was to prevent that, so not sure if we should have a very long page or a longish box that scrolls with up to 50 items.

adriana-alter:
I think the long page is okay.  (Especially if we're going to overhaul the list curation soon anyway...)

EdwinGuzman:
That may not be any time soon though!

adriana-alter:
True.  But the scroll box seems to be more trouble than it's worth.  Do you think it's actually a problem to have a long page, or...?

EdwinGuzman:
I do think it's a problem to have a long page in this case because you may want to see the rest of the page while scrolling through the lists of book. So maybe you want to compare a book in one list with the other list. In that case you can't because there's no bounding box.

adriana-alter:
Ah.  Fair.  I'll see what I can do!

PR #290, opened 2020-01-06
Narrow facets

adriana-alter:
Mostly resolves https://jira.nypl.org/browse/SIMPLY-589.  It works perfectly for filtering out saved value, i.e. values from props, but it doesn't reflect unsaved changes.  This is because of some weird things involving the "list setting with options" flavor of `ProtocolFormField`.  After spending a LOT of hours trying to force it to work (no, really--something like 10 hours), I have concluded that the list setting thing needs to be overhauled anyway, or possibly even replaced with `InputList`, and that's well outside the scope of this ticket...so it's extremely not worth sinking any more time into it here.


EdwinGuzman:
I'm a bit concerned about this because, component-wise, `FiltersForm` is the same as `LibraryEditForm`'s `renderFieldset` but without the fieldset (which is missing and should be added back for consistency). So it's now making `LibraryEditForm` less general with a specific use case while creating another component for it. I think it would be okay to have that specific case if `FiltersForm` didn't seem to have special cases for `ProtocolFormField`. Now it seems there are three different functions that all just call that component slightly differently props.

Maybe it would be better to just call `ProtocolFormField` once, but in the if/else update the settings object and other props accordingly instead.

The "Lanes & Filters" panel is now open by default even though it is optional and was previously collapsed.

To go with the problem this is trying to solve, maybe introducing state somewhere to keep track of the props would be better. I think I would go with the approach of a new component that renders:
```
 <checkboxes />
 <dropdown />
```
and in this component keep track of the state through any changes made by checkboxes. The values trickly up from that child and when they get updated, they can go down to dropdown and re-render. Might this be an option to try?

adriana-alter:
Goes with [this core branch](https://github.com/NYPL-Simplified/server_core/pull/1156)

src/components/EditableInput.tsx
```diff
@@ -133,6 +133,7 @@ export default class EditableInput extends React.Component<EditableInputProps, E
         value = this.getValue();
       }
       this.setState({ value, checked });
+      return [value, checked];
```
EdwinGuzman: What is this used for?

src/components/InputList.tsx
```diff
@@ -19,6 +19,8 @@ export interface InputListProps {
   onEmpty?: string;
   title?: string;
   capitalize?: boolean;
+  onChange?: (value: any) => {};
+  locked?: boolean;
```
EdwinGuzman: I think the `readOnly` name which is what it's being used for makes more sense than `"locked"`.

src/components/InputList.tsx
```diff
@@ -203,11 +208,16 @@ export default class InputList extends React.Component<InputListProps, InputList
     // All the possibilities, regardless of whether they've already been selected.
     let allOptions = (this.props.setting as any).menuOptions;
     if (!allOptions) {
-      return;
+      try {
+        allOptions = this.props.setting.options.map(o => <option key={o.key} value={o.key} aria-selected={false}>{o.label}</option>);
+      }
+      catch {
```
EdwinGuzman: A comment here about what's happening would be good.

src/components/InputList.tsx
```diff
@@ -203,11 +208,16 @@ export default class InputList extends React.Component<InputListProps, InputList
...
+        return;
+      }
     }
     // Items that have already been selected, and should be eliminated from the menu.
-    let listItems = this.state ? this.state.listItems : this.props.value;
-    // Items that haven't been selected yet.
+    let listItems = this.state ? this.state.listItems : (this.props.value || this.props.setting.default);
+    // // Items that haven't been selected yet.
```
EdwinGuzman: Double comment start.

src/components/LibraryEditForm.tsx
```diff
@@ -103,49 +104,66 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
...
-        { fields.map(setting =>
+        { fields.map(setting => (setting.paired &&
+          this.renderPairedMenus(setting, fields))
+        ||
+          !setting.skip &&
           <ProtocolFormField
             key={setting.key}
             ref={setting.key}
-            setting={setting}
+            setting={setting.format === "narrow" ? {...setting, ...{type: "menu"}} : setting}
```
EdwinGuzman: There are a lot of props that are long and inline and I think it'll be easier to read if they were assigned to variable names.

src/components/LibraryEditForm.tsx
```diff
@@ -103,49 +104,66 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
...
+      />
+    );
+  }
+
+  renderFieldset(fields: SettingData[]) {
     return (
       <fieldset>
         <legend className="visuallyHidden">Additional Fields</legend>
-        { fields.map(setting =>
+        { fields.map(setting => (setting.paired &&
```
EdwinGuzman: So if it's a "paired" menu then it'll render the combo, otherwise, it will render a field if it's not skipped? That `||` in the middle is hard to read.

src/components/PairedMenus.tsx
```diff
@@ -0,0 +1,100 @@
...
+  dropdownValue: string;
+}
+
+export default class PairedMenus extends React.Component<PairedMenusProps, PairedMenusState> {
+  constructor(props: PairedMenusProps) {
+    super(props);
+    let inputListValues = {};
+    let existingInput = this.existingInput();
+    this.props.inputListSetting.options.forEach(o => inputListValues[o.key] = existingInput.includes(o.key));
+    this.state = { inputListValues: this.existingInput(), dropdownValue: this.props.dropdownSetting ? (this.getValueFromItem(this.props.dropdownSetting) as string) : "" };
```
EdwinGuzman: Can you use the existing `existingInput` value or do you need to call this method twice? Also, this is a long line so it's fine to break it down into separate lines.

src/components/PairedMenus.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+export interface PairedMenusState {
+  inputListValues: string[];
+  dropdownValue: string;
+}
+
+export default class PairedMenus extends React.Component<PairedMenusProps, PairedMenusState> {
+  constructor(props: PairedMenusProps) {
+    super(props);
+    let inputListValues = {};
```
EdwinGuzman: I don't think you actually use this in the constructor besides creating and populating the object. It's in the state but that's only the name, so this variable itself doesn't seem to be used.

EdwinGuzman:
I like the new update and I think it's easier to visually reason about when you go through the UI.

I would update it so that:
1) when you delete an item and the dropdown appears, a description of what clicking on "add" will do appears
Screen Shot 2020-02-03 at 5 16 58 PM

2) When all options are deleted the "default availability" dropdown is still there but it's blank. You can save this option but it reverts to the original state (which I think is good). But, maybe it should return an error instead or warn you to have one option for that set of paired menus.
Screen Shot 2020-02-03 at 4 35 16 PM

Otherwise, it looks good! The changes are big and I'm still trying to figure out the changes.

src/components/EditableInput.tsx
adriana-alter: Makes the new value available to the parent component, so the parent component can update state accordingly.

src/components/InputList.tsx
adriana-alter: It seemed potentially confusing to give variables the same name as HTML attributes...

adriana-alter:
What kind of description of what clicking on "add" will do?  We don't have that for any of the other input menus; people can figure out that clicking "add" will add the thing...no?  Or did you mean something else?

adriana-alter:
The PairedMenus tests, which I definitely wrote, seem to have mysteriously disappeared :(  Will deal with that next.

src/components/InputList.tsx
EdwinGuzman: I think it's confusing if it's _not_ the same name, unless if it's used for something else.

src/components/LibraryEditForm.tsx
```diff
@@ -145,20 +145,34 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
...
+      }
+      return setting;
+    };
+    let settingValue = (setting: SettingData) => {
+      // Does the library already have a saved value for this setting?
+      if (this.props.item?.settings) {
+        return this.props.item.settings[setting.key];
+      }
+      return setting.default;
+    };
```
EdwinGuzman: 👍

EdwinGuzman:
New update looks good. This paired menu UI makes sense but I do worry it may be confusing to some. Not for this round but having more eyes on it would be great.

src/components/EditableInput.tsx
EdwinGuzman: Can you give an example of this? I'm not finding a place where a parent gets these values after an onChange event.

src/components/PairedMenus.tsx
```diff
@@ -0,0 +1,103 @@
...
+  onDropdownChange(value: string) {
+    this.setState({...this.state, ...{dropdownValue: value}});
+  }
+
+  getValueFromItem(setting: SettingData) {
+    let value = this.props.item && this.props.item.settings ?
+      this.props.item.settings[setting.key] : setting.default;
+    return value;
+  }
+
```
EdwinGuzman:
```
/**
 * Renders a list of input options to ....
 */
```

src/components/PairedMenus.tsx
```diff
@@ -0,0 +1,103 @@
...
+      />
+    );
+  }
+
+  renderDropdown() {
+    let setting = {...this.props.dropdownSetting};
+    let available = this.state.inputListValues && setting.options.filter(o => this.state.inputListValues.includes(o.key));
+    if (available.length === 0) {
+      return (
+        <p className="bg-warning">In order to set a default value, you must add at least one option.</p>
```
EdwinGuzman: 👍

src/components/ProtocolFormField.tsx
```diff
@@ -17,24 +17,24 @@ export interface ProtocolFormFieldProps {
...
   title?: string;
+  onChange?: any;
+  locked?: boolean;
 }
 
 /** Shows a form field for editing a single setting, based on setting information
     from the server. */
 export default class ProtocolFormField extends React.Component<ProtocolFormFieldProps, {}> {
-  constructor(props) {
+  constructor(props: ProtocolFormFieldProps) {
```
EdwinGuzman: Did typescript complain about this?

src/components/__tests__/InputList-test.tsx
```diff
@@ -367,5 +387,26 @@ describe("InputList", () => {
       let message = wrapper.find(".add-list-item-container span");
       expect(message.text()).to.equal("You've run out of options!");
     });
+    it("renders option elements if necessary", () => {
```
EdwinGuzman: Nice

EdwinGuzman: Do you need to test what happens when there is a `menuOptions` value and what gets rendered?

src/components/__tests__/LibraryEditForm-test.tsx
```diff
@@ -189,6 +222,15 @@ describe("LibraryEditForm", () => {
       nameInput.simulate("change");
     };
 
+    it("submits data from the FiltersForm", () => {
```
EdwinGuzman: What's the `FiltersForm`?

src/components/ProtocolFormField.tsx
adriana-alter: No, I was adding types to some other things (since typescript was complaining about them), and figured I might as well do this while I was at it.

src/components/__tests__/LibraryEditForm-test.tsx
adriana-alter: Something that no longer exists.  Good catch.

adriana-alter:
This commit does not address the latest round of feedback; it's just tests for the PairedMenus component.  Will address feedback in next commit!

src/components/EditableInput.tsx
adriana-alter: Um...okay, now I'm not finding it either.  And taking the line out doesn't seem to break anything.  Must not be using it anymore; will remove.

src/components/__tests__/InputList-test.tsx
adriana-alter: That's what the "it renders a dropdown menu" test is doing.

EdwinGuzman:
Updates look good but I noticed that the tests for the PairedMenu component is not in the commit... you said before that they disappeared so maybe they were not added when committing?

adriana-alter:
Do you mean...you can't see them now??  (I really don't want to write them a third time!) . They're in the commit before this last one.  The "PairedMenus-test" commit.

EdwinGuzman:
They are not... if you click on that commit, there's only one file that changed and that's the component itself. Weird...
Screen Shot 2020-02-10 at 12 30 44 PM

adriana-alter:
This is extremely weird.  How about now?

EdwinGuzman:
I can see them now.  Writing you commit in all caps made it work :)

adriana-alter:
It's MAGIC!

PR #291, opened 2020-01-15
Set default neighborhood analytics value to empty string

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2456.  (To test: try making a new local analytics integration, without touching the Patron Neighborhood Analytics form.)


EdwinGuzman:
Works fine, though I don't remember this failing before when I created a local analytics without and changes to the Patron Neighborhood Analytics form.

PR #292, opened 2020-01-17
Add automatic capitalization option for InputList

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2482; makes it so you can input lowercase values for the geographic settings.  (More broadly, makes it so InputList has an option for automatically capitalizing input.)

Goes with https://github.com/NYPL-Simplified/circulation/pull/1375 (which just adds `"geographic": True` to the two relevant settings).

### DIFF
--- a/src/components/InputList.tsx
+++ b/src/components/InputList.tsx
@@ -5,7 +5,7 @@ import { SettingData, CustomListsSetting } from "../interfaces";
 import ToolTip from "./ToolTip";
 import { LocatorIcon } from "@nypl/dgx-svg-icons";
 import { Button } from "library-simplified-reusable-components";
-import { isEqual } from "../utils/sharedFunctions";
+import { isEqual, formatString } from "../utils/sharedFunctions";
 
 export interface InputListProps {
   createEditableInput: (setting: SettingData, customProps?: any, children?: JSX.Element[]) => JSX.Element;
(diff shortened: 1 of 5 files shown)


adriana-alter:
Not that I'm seeing.  I'm not thrilled with it either.  I can take it out and just have all 2-character strings get capitalized--I can't imagine a realistic scenario in which someone would enter a 2-character string that shouldn't be all-caps--but that also doesn't seem ideal.  The only potentially better option I can think of is to do it on the back end by making `capitalize` an object--like, `capitalize: { handleTwoLetters: true }`--but I'm not sure that's worth it.  What do you think?

src/components/InputList.tsx
```diff
@@ -238,6 +240,14 @@ export default class InputList extends React.Component<InputListProps, InputList
     }
   }
 
+  capitalize(value: string): string {
+    // If it's a state/province abbreviation, the whole thing should be uppercase.  Otherwise, just capitalize
+    // the first letter of each word.
+    return value.split(" ").map(x =>
+      x.length === 2 && this.props.setting.format === "geographic" ? x.toUpperCase() : formatString(x)
```
EdwinGuzman: There's already a special case for `language-code` so it'd be better if this was updated to be more generic and have those two cases out, eventually. I think for now it should be fine. Maybe it could be a util function called here instead of having the function here, in case this needs to be done for other values elsewhere.

adriana-alter: Sounds good!

If it ever comes up elsewhere, I'll make it a util function.  For now, we seem to be good with `formatString` (util function which capitalizes the first letter of a word).

PR #293, opened 2020-01-27
Search for items in the Lists admin without a language filter.

EdwinGuzman:
Resolves [SIMPLY-1213](https://jira.nypl.org/browse/SIMPLY-1213).

This was originally completed to allow librarians and admins to search for items without a language filter. That part was done in `opds-web-client` for the search form but the custom lists search is a different component so it never adds the `&language=all` query for searches.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.4.7
+#### Updated
+- Added the query param `&language=all` for searches done in the List admin page to allow searches without a language filter.
+
 ### v0.4.6
 #### Updated
 - Implemented an optional case-insensitive configuration for InputList.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.4.4",
+  "version": "0.4.7",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.4.6",
+  "version": "0.4.7",
   "description": "Web Front-end for Library Simplified Circulation Manager Admin Interface",
   "repository": {
     "type": "git",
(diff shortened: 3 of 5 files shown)


adriana-alter:
Would it also work to search for specific languages (i.e. language=something-other-than-all)?  If so, let's add a test for that.

Otherwise, looks good!

EdwinGuzman:
No, it doesn't seem like the server responds to any language value except for `all`. There's no UI for admins to search by language at the moment and not sure if that's something we will add. This is more of an override to get all items rather can searching by language.

@leonardr , does this sound right?

leonardr:
Yes, the current feature is just an override.

I think https://jira.nypl.org/browse/SIMPLY-564 would provide a useful way around this that could be used by both librarians and patrons.

christhepianist:
Is there a ticket to allow searching by language in the Lists admin? It would be very helpful for librarians creating lanes for subsets of their communities.


Christine Peterson

Christine Peterson
Continuing Education Librarian for Technology
Amigos Library Services
800-844-8482 x2891

________________________________

leonardr:
Since https://jira.nypl.org/browse/SIMPLY-564 changes the way the search engine handles queries, it would be usable everywhere.

christhepianist:
For patrons, this might work, however, for librarians creating queries, they will expect a language filter. The difference between a title “The Italian Language” and wanting something written in Italian.


Christine

leonardr:
Once we have OPDS 2.0 support we can discuss an explicit filter feature, but I would like folks to try the SIMPLY-564 solution before we commit to that.

I say this because our existing system does a good job of handling cases where a search query may or may not contain a genre ("billionaire romance" / "Modern Romance"), fiction status ("fiction dinosaurs"/"The Art of Fiction"), or age group ("grade 5 dogs" / "Tales of a Fourth Grade Nothing").

564 would expand this to treat the name of a language as a _potential_ keyword. For "The Italian Language", we would create two [hypotheses](https://github.com/NYPL-Simplified/Simplified/wiki/Elasticsearch#hypotheses): one that you are looking for an Italian book called _The_ which is classified under Language, and one that you are looking for a book called _The Italian Language_. Obviously the second hypothesis would win out.

Whereas if you searched for "italian romance", the hypothesis that you are looking for books in Italian and classified under Romance would probably win out over the hypothesis that you are looking for a book called _Italian Romance_.

christhepianist:
Happy to try it out.


Chris

killerpuppytails:
🎉

PR #295, opened 2020-02-06
Get footer text and link from server values

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2527


src/components/Footer.tsx
```diff
@@ -1,32 +1,39 @@
...
+        <p><a href={tos_link_href} target="_blank">{this.decode(tos_link_text)}</a></p>
       </footer>
     );
   }
+
+  decode(text) {
+    // The current Terms of Service link text has an apostrophe in it.  If we don't go through
+    // this extra step, the browser renders the HTML entity code for the apostrophe.
+    let element = document.createElement("textarea");
+    element.innerHTML = text;
```
EdwinGuzman: The "react" way of doing this is through `dangerouslySetInnerHTML`. It'll be good to use it to also remember that we are rendering something that can be bad, like an entity code.

adriana-alter:
When I try to run the tests in local, I am blocked by this fun error message from tslint-react-a11y: ` Link text or the alt text of image in link should be at least 4 characters long. If you are not using <a> element as anchor, please specify explicit role, e.g. role='button'` . Have you come across this before?  I'm not seeing a way to disable it.  (And the link test IS at least 4 characters long!)

EdwinGuzman:
Ha I haven't and that's pretty funny... I'll try it soon. The problem right now is that `TOSContext` file is not committed.

src/components/Footer.tsx
```diff
@@ -1,31 +1,22 @@
...
 process. */
 export default class Footer extends React.Component<FooterProps, {}> {
+  static contextType = TOSContext;
+
   render() {
-    let links = this.props.links || [[
-      "Terms of Service for presenting e-reading materials through NYPL's SimplyE mobile app",
-      "https://www.librarysimplified.org/SimplyETermsofService2.html"
-    ]];
+    let [text, href] = this.context;
```
EdwinGuzman: An alternative you can do is convert this to a functional component and use the `useContext` hook:
```
const Footer = (props: FooterProps) => {
    let [text, href] = React.useContext(TOSContext);
// ....
```

src/components/TOSContext.tsx
```diff
@@ -0,0 +1,3 @@
+import * as React from "react";
```
EdwinGuzman: It's the all caps commit that made it work.

EdwinGuzman:
The test warning is because now that `dangerouslySetInnerHTML` is being used, there's no explicit text in the anchor element (<a ... /> instead of <a ...>text string</a>). 

You can revert back to the previous `decode` function or try something like https://github.com/mathiasbynens/he . This seems very silly but we can't turn the error off.

adriana-alter:
I brought back `decode`...and it's STILL NOT WORKING :(   I'll give the other thing a try!

EdwinGuzman:
Your decode function or the other library shouldn't make a difference if you're setting the text for the link. I tried placing your decode function back to test the linter and it worked for me...

PR #297, opened 2020-02-14
Sort by title

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2539 and https://jira.nypl.org/browse/SIMPLY-2538

Screen Shot 2020-02-14 at 12 54 11 PM

### DIFF
--- a/src/components/CustomListEntriesEditor.tsx
+++ b/src/components/CustomListEntriesEditor.tsx
@@ -12,6 +12,7 @@ import {
   BookIcon,
 } from "@nypl/dgx-svg-icons";
 import CatalogLink from "opds-web-client/lib/components/CatalogLink";
+import EditableInput from "./EditableInput";
 import { formatString } from "../utils/sharedFunctions";
 
 export interface Entry extends BookData {
@@ -63,7 +64,7 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
     this.clearState = this.clearState.bind(this);
   }
 
-  sortBy(books, attr: string) {
+  sortBy(books: BookData[], attr: string) {
     let parse = (value: string | string[]) => {
       if (attr === "authors") {
         return value[0].split(" ").reverse()[0];
@@ -121,14 +122,21 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
     }
 
     let sortDropdown = (
-      <fieldset className="sort-results-by">
-        <label>Sort by:</label>
-        <select onChange={(e) => this.setState({ sortBy: e.currentTarget.value.toLowerCase() })}>
-          <option>Default</option>
-          <option>Title</option>
-          <option>Authors</option>
-        </select>
-      </fieldset>
+      <EditableInput
+        key="sortBy"
+        elementType="select"
+        name="sortBy"
+        className="sort-results-by"
+        label="Sort search results by:"
+        onChange={(value) => this.setState({ sortBy: value.toLowerCase() })}
+      >
+        {
+          ["default", "title", "authors"].map((x) => {
+            return <option key={x} value={x} aria-selected={this.state.sortBy === x}>{formatString(x)}</option>;
+          })
+
+        }
+      </EditableInput>
     );
 
     return (
@@ -139,6 +147,7 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
               <h4>Search Results</h4>
               { searchResults && (this.searchResultsNotInEntries().length > 0) &&
                 [<Button
+                  key="addAll"
                   className="add-all-button"
                   callback={this.addAll}
                   content={<span>Add all to list<ApplyIcon /></span>}
--- a/src/components/CustomListEntriesEditor.tsx
+++ b/src/components/CustomListEntriesEditor.tsx
@@ -12,6 +12,7 @@ import {
   BookIcon,
 } from "@nypl/dgx-svg-icons";
 import CatalogLink from "opds-web-client/lib/components/CatalogLink";
+import { formatString } from "../utils/sharedFunctions";
 
 export interface Entry extends BookData {
   medium?: string;
@@ -37,6 +38,7 @@ export interface CustomListEntriesEditorState {
   deleted: Entry[];
   added: Entry[];
   totalVisibleEntries?: number;
+  sortBy?: string;
 }
 
 /** Drag and drop interface for adding books from search results to a custom list. */
@@ -59,6 +61,17 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
     this.loadMore = this.loadMore.bind(this);
     this.loadMoreEntries = this.loadMoreEntries.bind(this);
     this.clearState = this.clearState.bind(this);
+  }
+
+  sortBy(books, attr: string) {
+    let parse = (value: string | string[]) => {
+      if (attr === "authors") {
+        return value[0].split(" ").reverse()[0];
+      } else if (attr === "title") {
+        return formatString((value as string), ["The ", "A ", "An ", ""], false);
+      }
+    };
+    return books.sort((x, y) => (parse(x[attr]) > parse(y[attr])) ? 1 : -1);
   }
 
   render(): JSX.Element {
@@ -107,6 +120,17 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
       }
     }
 
+    let sortDropdown = (
+      <fieldset className="sort-results-by">
+        <label>Sort by:</label>
+        <select onChange={(e) => this.setState({ sortBy: e.currentTarget.value.toLowerCase() })}>
+          <option>Default</option>
+          <option>Title</option>
+          <option>Authors</option>
+        </select>
+      </fieldset>
+    );
+
     return (
       <DragDropContext onDragStart={this.onDragStart} onDragEnd={this.onDragEnd}>
         <div className="custom-list-drag-and-drop">
@@ -114,11 +138,12 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
             <div className="droppable-header">
               <h4>Search Results</h4>
               { searchResults && (this.searchResultsNotInEntries().length > 0) &&
-                <Button
+                [<Button
                   className="add-all-button"
                   callback={this.addAll}
                   content={<span>Add all to list<ApplyIcon /></span>}
-                />
+                />,
+                sortDropdown]
               }
             </div>
             <Droppable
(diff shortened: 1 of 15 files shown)


src/components/CustomListEntriesEditor.tsx
```diff
@@ -107,18 +125,38 @@ export default class CustomListEntriesEditor extends React.Component<CustomListE
...
+      <EditableInput
+        key="sortBy"
+        elementType="select"
+        name="sortBy"
+        className="sort-results-by"
+        label="Sort search results by:"
+        onChange={(value) => this.setState({ sortBy: value.toLowerCase() })}
+      >
+        {
+          ["default", "title", "authors"].map((x) => {
```
EdwinGuzman: `["default", "title", "authors"]` should be a variable.

src/components/__tests__/CustomListEntriesEditor-test.tsx
```diff
@@ -118,6 +117,54 @@ describe("CustomListEntriesEditor", () => {
     expect(results.at(2).text()).to.contain("author 3");
   });
 
+  it("sorts search results", () => {
```
EdwinGuzman: I think this would be better if you broke it up into separate and smaller test suites to define each component state and behavior, i.e. "it should render the default state", "it should sort by author", etc.

EdwinGuzman:
Looks good but doesn't the server already return sorted results by author or title? It may be better to let the server handle this and let the client make the appropriate api request.

EdwinGuzman:
No, just checked, the server doesn't sort results by author or title. I think it'll be a good change to add this to the server instead of the client.

adriana-alter:
I'm not convinced it's a better idea to make a lot of extra calls to the server; this feature is designed for users to get the search results and then sort them one or more times afterwards (as opposed to submitting a search with a sorting filter as a parameter).  What's the advantage to sending this back and forth between the server and the front end a bunch of times, rather than just letting the front end quickly take care of it?

EdwinGuzman:
I do agree that it's not helpful to make a lot of calls, but sorting lists by author and title seems like something that the mobile clients will also benefit from. So, having a unified way of doing it in the server makes sense to me.

From the other jira tickets, it looks like you will need to support language and boolean queries, which will need to be done in the server, no? So it might be helpful to architect it so you can get all the queries in an API call.

adriana-alter:
I'm...still not convinced!  Yes, the language and boolean queries will need to be done in the server, but those are a whole separate thing; it seems wrong to lump search filters together with sorting mechanisms.  The language/boolean information will need to be sent to the server along with the search query (I figured I would build an "advanced search options" form for them); that's very different from the sorting stuff, which happens after the server has already returned the search results.  And I don't think it's worth building this in an unnecessarily convoluted way just because we might hypothetically need similar functionality on the mobile client one day.  How strongly do you feel about this?  Is it time to get a third opinion?

EdwinGuzman:
Hmm that's fine for now like this but if there will be sorting by author and title in the mobile clients, we'd want the same implementation. I'm fine with this but also update the small things I pointed out.

We can ask @leonardr what he thinks about this, too.

leonardr:
Because our search query includes fuzzy matches, it's common for searches to get thousands of results. My go-to search is `penguin` and you can see that [there are over 9000 results](http://qa.circulation.librarysimplified.org/QANYPL/search/?q=penguin&after=9000) for that query on NYPL's QA server, results that pretty quickly stop having anything to do with penguins. 

So, on its own, telling Elasticsearch to sort by title isn't going to give good results. Grabbing the first page of results and sorting _that_ by title client-side (I assume this branch grabs the first page) is a good first step. But it would be better to have something available on the server side for two reasons. First, on the server side we have `sort_title` and `sort_author` available, which saves a lot of complicated code dealing with "The", "An", "A" -- and a whole bunch of even worse code when you search by author. Second, we don't know how many 'good' results there will be. There might be only one or there might be thousands. So grabbing one page is a second-best solution.

A server-side solution would involve changing the search controller to allow sorting by title and author, but to institute a strict cutoff on the level of quality. That way you can pull pages until you run out of results without worrying that the first result alphabetically has nothing to do with the search term.

aslagle:
What is the use case for this feature?

leonardr:
When a librarian is creating a custom list, they run searches that are more general than the ones patrons run. Patrons are very likely to be looking for a specific book; librarians will often be looking for a number of related books that they're considering for inclusion in a custom lane. (e.g. the "Animals" lane in Open eBooks)

When patrons browse a lane they get the choice of ordering by title or author. The thinking here is that it makes sense to give librarians the opportunity to order their ad-hoc "lane" in the same ways. They can go down the lane, adding books as they go. Among other things, alphabetical order gives the librarian a way to leave off the task and come back later.

I do not think this feature will be super-useful for mobile, though it might make sense when the search is clearly a topic search.

aslagle:
This seems very odd since the order will change whenever you load another page of search results. It's hard to imagine how someone would use it to leave and come back later.

EdwinGuzman:
Re: Loading another page of search results - when a new set of search results is fetched, the entire list of search results is updated and re-sorted. Something I noticed, is that when you are at the bottom and click on "load more" it's possible (depending on the search term) that you don't notice any change because the last few results stay the same. This can be confusing because one can think the list wasn't updated.
Screen Shot 2020-02-19 at 2 23 16 PM
In this example, the last three results stayed the same.

killerpuppytails:
Good point @EdwinGuzman - we should include some copy by the load more button to inform librarians that new search results might be sorted into the middle of the list .... and maybe add some visual indicator as to which ones were added for that new load?

leonardr:
I agree that it's hard to close a big lane, reopen it, and get back to the place you were. That's also a problem on mobile, and the solution I've proposed for mobile ([alphabetic signposts](https://jira.nypl.org/browse/SIMPLY-415)) would also help here.

I also agree that the implementation in this branch doesn't work when there's more than one page of results. When I get back to work (hopefully tomorrow) I'll work with Adriana to make a server-side branch that will simplify the client-side work and meet the use case better.

adriana-alter:
We have a whole new UI!  (I anticipate there eventually being multiple fieldsets in the search options box.)

Screen Shot 2020-02-28 at 12 57 37 PM

EdwinGuzman:
Looks good so far but I feel like the long search button doesn't make it look like a button and it feels awkward. The UI for the "sort by" seems okay for now, but for an advance search tab, I think it's better to have everything available upfront instead of hiding selections under different tabs. You'd then have to review each tab to make sure your search is correct rather than seeing it all at one. But now that I look at it, is it suppose to be tabs or just look like a tab? It seems a big misleading right now.

adriana-alter:
Ah.  It's not a tab, and it wasn't supposed to look like a tab; I just thought it was a nice way to display the fieldset legend.  I see your point.

EdwinGuzman:
I'd just follow what is already available, like what's in the Registry Admin.

adriana-alter:
What do you mean--the RegistryAdmin filters section?  That has toggles...and I don't want to use toggles for this because then it looks like you can pick more than one, which seems confusing for users...

src/components/CustomListSearch.tsx
```diff
@@ -0,0 +1,129 @@
...
+  constructor(props: CustomListSearchProps) {
+    super(props);
+    this.submitSearch = this.submitSearch.bind(this);
+    this.sort = this.sort.bind(this);
+    this.state = { sortBy: null };
+  }
+
+  componentDidMount() {
+    if (this.props.startingTitle) {
+      (this.refs["searchTerms"] as HTMLInputElement).value = this.props.startingTitle;
```
EdwinGuzman: Make updating this to the new ref api a different task/PR.

src/components/CustomListSearch.tsx
```diff
@@ -0,0 +1,129 @@
...
+  renderSearchBox(): JSX.Element {
+    return (
+      <fieldset key="search">
+        <legend className="visuallyHidden">Search for titles</legend>
+        {
+          this.props.entryPoints?.length ? (
+            <div className="entry-points">
+              <span>Select the entry point to search for:</span>
+              <div className="entry-points-selection">
+                {this.props.getEntryPointsElms(this.props.entryPoints)}
```
EdwinGuzman: It looks like now the "All" option is part of the result from the server. So now, this function renders "All" twice. Can you update it?

Screen Shot 2020-02-28 at 3 05 38 PM

EdwinGuzman: Also, that function is defined in `CustomListEditor` but is only used in this file, so maybe you can bring in that function here instead of passing it as a prop.

EdwinGuzman: Ooof, I see that the `CustomListEditor` keeps the entrypoint value in its state, so maybe this actually isn't so doable (but do delete the extra "All").

src/components/__tests__/CustomListEntriesEditor-test.tsx
```diff
@@ -1,5 +1,5 @@
 import { expect } from "chai";
-import { stub } from "sinon";
+import { spy, stub } from "sinon";
```
EdwinGuzman: Not using spy.

src/stylesheets/custom_list_editor.scss
```diff
@@ -107,3 +107,62 @@
     }
   }
 }
+
+.search-options {
```
EdwinGuzman: Yeah, this might be a good opportunity to reuse what's in the Registry Admin for fieldsets?

EdwinGuzman:
Oh, not toggles but the fieldset you're using for choosing a stage. That same wrapping style.

adriana-alter:
Ohhh.  Yes, good idea!

src/components/CustomListSearch.tsx
adriana-alter: I don't have the entry point menu; how are you even getting it to show up...?

adriana-alter:
Screen Shot 2020-03-02 at 11 34 48 AM

src/components/CustomListSearch.tsx
adriana-alter: Nvm, got it.

EdwinGuzman: Cool. It's in the `Lanes & Filters` menu for a library.

EdwinGuzman:
Looks good. I'm not sure if it's because of this PR but this is not happening on `master` (or the production instance) in the Reset Adobe form: 
Screen Shot 2020-03-02 at 11 46 50 AM

adriana-alter:
That's...new.  I can't imagine why it would be from this PR.  Is it on QA?

adriana-alter:
Okay, it seems like it IS from this PR.  Hang on...

adriana-alter:
Okay, I have no idea why that was happening--I tried commenting out all of the CSS added/changed by this branch, and that didn't help--but I fixed it.

PR #298, opened 2020-02-14
Disable button if tests can't be run

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2493

It might be a better idea to hide the button altogether; opinions?

Screen Shot 2020-02-14 at 1 54 23 PM

### DIFF
--- a/src/components/SelfTests.tsx
+++ b/src/components/SelfTests.tsx
@@ -70,7 +70,6 @@ export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
   render(): JSX.Element {
     const integration: ServiceData = this.state.mostRecent;
     const selfTestException = integration.self_test_results && integration.self_test_results.exception;
-    console.log(integration.self_test_results);
     const cannotRunTests = selfTestException && integration.self_test_results.disabled;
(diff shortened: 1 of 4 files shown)


src/components/SelfTests.tsx
```diff
@@ -70,6 +70,7 @@ export class SelfTests extends React.Component<SelfTestsProps, SelfTestsState> {
   render(): JSX.Element {
     const integration: ServiceData = this.state.mostRecent;
     const selfTestException = integration.self_test_results && integration.self_test_results.exception;
+    const cannotRunTests = selfTestException?.search(/[before you can run self tests]/) >= 0;
```
EdwinGuzman: Is there a better way to set this up so it's not doing a regex on a string? Perhaps make the server return a flag that says that service cannot run self-tests?

adriana-alter: I will investigate!  Just so I understand, though...what's the problem with doing it as a regex?

adriana-alter: Update: turns out the server IS already returning a flag (I just didn't see it before)!

EdwinGuzman: I don't like relying on strings to do checks... seems like sturdy.

PR #300, opened 2020-03-11
Language filter

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2537

Goes with https://github.com/NYPL-Simplified/server_core/pull/1168

Screen Shot 2020-03-11 at 1 36 23 PM

### DIFF
--- a/src/components/CustomListEditor.tsx
+++ b/src/components/CustomListEditor.tsx
@@ -306,12 +306,12 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
 
   getSearchQueries(sortBy: string, language?: string) {
     const entryPointSelected = this.state.entryPointSelected;
-    let query = language ? "" : "&languages=all";
+    let query = language ? "" : "&language=all";
     if (entryPointSelected && entryPointSelected !== "all") {
       query += `&entrypoint=${encodeURIComponent(entryPointSelected)}`;
     }
     sortBy && (query += `&order=${encodeURIComponent(sortBy)}`);
-    language && (query += `&languages=${[language]}`);
+    language && (query += `&language=${[language]}`);
     return query;
   }
 
--- a/src/components/CustomListEditor.tsx
+++ b/src/components/CustomListEditor.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { CollectionData as AdminCollectionData } from "../interfaces";
+import { LanguagesData, LibraryData, CollectionData as AdminCollectionData } from "../interfaces";
 import { CollectionData, BookData } from "opds-web-client/lib/interfaces";
 import TextWithEditMode from "./TextWithEditMode";
 import EditableInput from "./EditableInput";
@@ -9,7 +9,8 @@ import SearchIcon from "./icons/SearchIcon";
 import { Button, Panel, Form } from "library-simplified-reusable-components";
 
 export interface CustomListEditorProps extends React.Props<CustomListEditor> {
-  library: string;
+  languages: LanguagesData;
+  library: LibraryData;
   list?: CollectionData;
   listId?: string | number;
   listCollections?: AdminCollectionData[];
@@ -58,7 +59,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     const listId = this.props.listId;
     const listTitle = this.props.list && this.props.list.title ? this.props.list.title : "";
     const nextPageUrl = this.props.list && this.props.list.nextPageUrl;
-    const opdsFeedUrl = `${this.props.library}/lists/${listTitle}/crawlable`;
+    const opdsFeedUrl = `${this.props.library?.short_name}/lists/${listTitle}/crawlable`;
     const hasChanges = this.hasChanges();
     // The "save this list" button should be disabled if there are no changes
     // or if the list's title is empty.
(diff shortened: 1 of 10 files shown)


src/components/CustomListEditor.tsx
```diff
@@ -58,7 +59,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     const listId = this.props.listId;
     const listTitle = this.props.list && this.props.list.title ? this.props.list.title : "";
     const nextPageUrl = this.props.list && this.props.list.nextPageUrl;
-    const opdsFeedUrl = `${this.props.library}/lists/${listTitle}/crawlable`;
+    const opdsFeedUrl = `${this.props.library?.short_name}/lists/${listTitle}/crawlable`;
```
EdwinGuzman: Why is this the `short_name` now?

src/components/CustomListEditor.tsx
```diff
@@ -301,13 +304,14 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
-  getSearchQueries(sortBy: string) {
+  getSearchQueries(sortBy: string, language: string) {
     const entryPointSelected = this.state.entryPointSelected;
-    let query = "&language=all";
+    let query = "";
     if (entryPointSelected && entryPointSelected !== "all") {
       query += `&entrypoint=${encodeURIComponent(entryPointSelected)}`;
     }
     sortBy && (query += `&order=${encodeURIComponent(sortBy)}`);
+    language && (query += `&language=${[language]}`);
```
EdwinGuzman: Do you know if we don't need `all` to be the default anymore?

src/components/CustomListEditor.tsx
adriana-alter: Because `library` is now an object rather than a string.  The string was just the relevant library's short_name.  (We're passing the whole library object so that we can access the information about what languages it has.)

src/components/CustomListEditor.tsx
adriana-alter: Seems unnecessary to explicitly set it here; the select menu defaults to `all`.

src/components/CustomListEditor.tsx
EdwinGuzman: I see. Does this have repercussions on any other parts of any other apps?... I'll know the server code is already merged in, but I'll test opds-web-client and circulation-patron-web later.

src/components/CustomListSearch.tsx
```diff
@@ -31,7 +34,8 @@ export default class CustomListSearch extends React.Component<CustomListSearchPr
 
   submitSearch() {
     const searchTerms = encodeURIComponent((this.refs["searchTerms"] as HTMLInputElement).value);
-    this.props.search(searchTerms, this.state.sortBy);
+    const language = encodeURIComponent((this.refs["languages"] as HTMLSelectElement).value);
```
EdwinGuzman: Eventually, this should use the new ref api.

src/components/CustomListSearch.tsx
```diff
@@ -81,12 +85,35 @@ export default class CustomListSearch extends React.Component<CustomListSearchPr
...
+            {languages.map(x => <option value={x} aria-selected={false}>{this.getLanguageName(x)}</option>)}
+          </select>
+        </section>
+      </fieldset>
+    );
+    return fieldset;
+  }
+
+  getLanguageName(languageAbbreviation: string): string {
+    return this.props.languages && this.props.languages[languageAbbreviation].join("; ");
```
EdwinGuzman: Can you use the `?.` shortcut here? Not a big deal if you can't.

src/components/__tests__/CustomListEditor-test.tsx
```diff
@@ -427,15 +452,25 @@ describe("CustomListEditor", () => {
 
     let searchForm = wrapper.find("form");
     searchForm.simulate("submit");
-
+    // The default lanuage is "all"
```
EdwinGuzman: Small typo.

src/components/CustomListEditor.tsx
adriana-alter: I don't see how it could; the exact same value is getting passed along...

src/components/__tests__/CustomListEditor-test.tsx
adriana-alter: Good catch!!

src/components/CustomListEditor.tsx
EdwinGuzman: Okay, great.

src/components/CustomListSearch.tsx
adriana-alter: As I recall, I tried to, but there was an issue with following the `?` with brackets instead of with a dot...

src/components/CustomListEditor.tsx
EdwinGuzman: Maybe the other apps don't know they need to do `library.short_name` instead of just `library`.

adriana-alter: But `library` is only getting passed to `CustomListSearch`, which does know.

EdwinGuzman: I see. Didn't understand what was going on before.

PR #305, opened 2020-04-17
Save list

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2692, and also a related TextWithEditMode bug which turned out to be making it possible to save lists with blank names.


src/components/CustomListEditor.tsx
```diff
@@ -175,19 +176,23 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
...
         title,
         entries: nextProps.list.books,
-        collections: collections,
+        collections: collections
       });
     } else if ((!this.props.list || !this.props.listCollections) && nextProps.list && nextProps.listCollections) {
       this.setState({
         title: this.state.title,
         entries: this.state.entries,
-        collections: nextProps.listCollections,
```
EdwinGuzman, line 184: Technically, we do use dangling commas. This isn't an issue now, but it'll come back with prettier most likely.

src/components/__tests__/TextWithEditMode-test.tsx
```diff
@@ -164,4 +164,39 @@ describe("TextWithEditMode", () => {
 
     getValueStub.restore();
   });
+
+  it("optionally disables the save button if the text is blank", () => {
+    let wrapper = mount(
+      <TextWithEditMode
+        placeholder="editable thing"
+        aria-label="label"
+        disableIfBlank={true}
```
EdwinGuzman, line 173: You can just do `disableIfBlank` and that'll pass in `true`.

EdwinGuzman:
There is an issue that's new: you can save a list without a title.

On QA circulation:
1. go to the create a new list page
2. search for a book (I'm using "car")
3. add one item to the list - the "Save this list" button is disabled
4. if the "save list title" is clicked without a title, the "save this list" button is still disabled. the list needs a list title first and it can't be an empty string.

Local:
1. go to the create a new list page
2. search for a book ("car" again)
3. add one item to the list - the "Save this list" button is enabled. This is wrong since a list title wasn't added. **This is the new bug**.
4. this part is better since you cannot click on "save list title" without a title added.

EdwinGuzman:
The last update fixes the bug found... but here's another that doesn't appear in QA:

If you switch to a different list to edit, the "save this list" button is always enabled even though no changes were made (since you just switched to it). I think this also appears when clicking on "create new list" button - the "save this list" button is enabled initially.

adriana-alter:
This ties up with something else that I think might be a bug (but is on QA too): when you submit a search in the form for one list, and then switch to editing a different list, the search results are still there.  Do we actually want that?

EdwinGuzman:
That _might_ not be a bug. That's more of a UX issue because what if you want the same search results for different lists? So that's hard to say...

src/components/CustomListEditor.tsx
```diff
@@ -158,7 +159,7 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     // into account.
     if (!nextProps.list && !this.props.list) {
       // This is no current or previous list, so this is a new list.
-      this.setState({ title: "", entries: [], collections: [] });
+      this.setState({ title: (this.props.searchResults ? "" : this.state.title), entries: [], collections: []});
```
EdwinGuzman, line 162: I think this last fix is fine for now, but it'd be nice to reset the state each time the route changes to a new edit page instead of looking specifically for search results (as in, it should be possible to not make this rely on search results to show the right state).

EdwinGuzman:
Unfortunately, the last bug still appears - when switching to another edit page for the first time, the "save this list" button is disabled. But then when you switch to a second edit page or to creates new list page, the button is enabled automatically.

EdwinGuzman:
I also filed this ticket because the icons on the list page are now the wrong size - https://jira.nypl.org/browse/SIMPLY-2715

EdwinGuzman:
I haven't tested the last update but that might not be the best approach. Usually, depending on the router for these kinds of updates have proven to be unstable (in past projects). Did you try something else besides this?

adriana-alter:
Tried everything I could think of; this was the only thing that worked.  What do you think might be better?

EdwinGuzman:
I noticed I have one list that whenever I go to that list to edit, no matter from where, it will always have the "save this list" button enabled. Doesn't matter if there were any changes made, it'll always be enabled. Navigating to other lists to edit are disabled by default which is good. Do you have this issue with any one list?

adriana-alter:
Nope.  Is there something different about that list...?

EdwinGuzman:
Nope.

I don't think this is the best approach but the logic for this page is getting out of hand and I think a rewrite would be better long-term approach. For now, this is fine but you didn't remove the listener when the component unmounts, so the `setState` is still happening when you go to some other pages.

EdwinGuzman:
The last commit is good for this use case since the `CustomListEditor` component is only used once for this app. But let's say you have a component which is mounted multiple times on a page and you remove one. In that situation, this approach wouldn't work because when you unmount _one_ instance, you'll remove the listener for _all_ instances.

This is good to go but can you first review #306 and fix the merge conflicts? Then the `0.4.18`  can have multiple updates in that one release. Thanks!

adriana-alter:
How strongly do you feel about waiting to merge until #306 is ready (given that #306 isn't working for me, so there's going to be some more back-and-forth on it)?  I feel like I've been working on this forever; I'd really like to get it in before the end of the sprint, and just be done with it...

EdwinGuzman:
Let's give it until tomorrow morning to see if there are any updates to the #306 and if not, it'll go into the next release.

PR #306, opened 2020-04-24
Fix view detail link generation

tdilauro:
## JIRA Ticket

https://jira.nypl.org/browse/SIMPLY-2373

## What does this PR do?

Search results obtained while creating a new list were incorrectly constructed. This is because the link was always being constructed with the list title, but there is not yet a list when creating. The structure looked like this:
```
/<library-short-name>/lists/<list-title>/crawlable
```
or like this when creating (and there is not `list` yet:
```
/<library-short-name>/lists//crawlable
```

When creating a list, the the structure should look like:
```
/<library-short-name>/crawlable
```

## How should this be tested?

It would be good to add a test for this. I’m not a JS person, so it would be great if someone else could add a test or two for this.

I have performed manual live testing to convince myself that the links are generated correctly both when _editing_ and when _creating_ lists.

To recreate this error pre-fix:
- Enter List Manager and ensure that location URL ends in `/create'. If not, click the `Create New List` button and ensure that you have entered create mode.
- Perform a search that yields results.
- Click the `View Details` button for one of `Search Results` items.
- You should receive an error modal with text that begins like:
    ```
    Error
    Could not fetch data: https://<CM-instance>/<library>/lists//crawlable
    ```

### DIFF
--- a/src/components/CustomListEditor.tsx
+++ b/src/components/CustomListEditor.tsx
@@ -59,7 +59,8 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     const listId = this.props.listId;
     const listTitle = this.props.list && this.props.list.title ? this.props.list.title : "";
     const nextPageUrl = this.props.list && this.props.list.nextPageUrl;
-    const opdsFeedUrl = `${this.props.library?.short_name}/lists/${listTitle}/crawlable`;
+    const crawlable = `${listTitle ? `lists/${listTitle}/` : ""}crawlable`;
+    const opdsFeedUrl = `${this.props.library?.short_name}/${crawlable}`;
     const hasChanges = this.hasChanges();
     // The "save this list" button should be disabled if there are no changes
     // or if the list's title is empty.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ### v0.4.18
 #### Removed
 - Removed a reference to the unused `isActive` property of `opds-web-client`'s `router` context.
+#### Fixed
+- Fixed link construction for book `View Detail` button when creating a list in List Manager.
 
 ### v0.4.17
 #### Updated


src/components/CustomListEditor.tsx
```diff
@@ -59,7 +59,8 @@ export default class CustomListEditor extends React.Component<CustomListEditorPr
     const listId = this.props.listId;
     const listTitle = this.props.list && this.props.list.title ? this.props.list.title : "";
     const nextPageUrl = this.props.list && this.props.list.nextPageUrl;
-    const opdsFeedUrl = `${this.props.library?.short_name}/lists/${listTitle}/crawlable`;
+    const crawlable = (listTitle ? "lists/" + listTitle + "/" : "") + "crawlable";
```
EdwinGuzman, line 62: This should also be an interpolated string:
```
`${listTitle ? `lists/${listTitle}/` : ""}crawlable;`
```

EdwinGuzman:
Thanks for this. Can you give a way to reproduce the bug where search results were badly constructing while searching in a new list? I just tried it and obtain results when creating a new list, so I don't know what this is fixing.

The other things missing from this PR is an updated version in `package.json` as well as a description of the update in the CHANGELOG. Here's an example of a current PR that's in review:  #305 . It might be possible to get this change in along with Adriana's update, so you can update to `0.4.18` as well.

tdilauro:
> Thanks for this. Can you give a way to reproduce the bug where search results were badly constructing while searching in a new list? I just tried it and obtain results when creating a new list, so I don't know what this is fixing.

I updated the original PR text to include how to replicate pre-fix.

> The other things missing from this PR is an updated version in `package.json` as well as a description of the update in the CHANGELOG. Here's an example of a current PR that's in review: #305 . It might be possible to get this change in along with Adriana's update, so you can update to `0.4.18` as well.

I'll complete these updates and the other one you suggested.

EdwinGuzman:
Ah, there's a minor merge conflict from `master` you can fix.

tdilauro:
Thanks, @EdwinGuzman! The conflict is resolved.

adriana-alter:
I'm still getting an error message...
Screen Shot 2020-04-27 at 2 19 32 PM

tdilauro:
@adriana-alter Do you have any idea where that `cat?` in the URL is coming from. It is in the position that should reflect the library short name.

Can you share steps to reproduce this?

adriana-alter:
@tdilauro "Cat" was my search term on the list creator page.  I went to the list creator form, ran a search for "cat", and then clicked on the details link for the first result.  Tried it both before and after giving the new list a name.

tdilauro:
@adriana-alter I was mistaken in my early post (now edited). `cat?` is in the position in the URL that should reflect the list name. That would imply that the list has already been created.

Was your list named "cat?" (with the question mark), by any chance?

I am able to replicate the error you saw with the following steps:
- Go to List Manager and create a new list.
- Set the list name to "cat?".
- Save the list.
- Perform the search ("cat" or anything else).
- Click on the `View details` button.

The fix in this PR needs to be tested _before_ saving the list (see steps in [initial comment](https://github.com/NYPL-Simplified/circulation-web/pull/306#issue-408804118) above).

adriana-alter:
Yes, that's working now!  Do we already have a separate ticket for fixing the scenario in which you've already saved the list (a more likely use case)?

tdilauro:
@adriana-alter I'm not sure. The [error message above 
](https://github.com/NYPL-Simplified/circulation-web/pull/306#issuecomment-620151612) was triggered by a URL special character (`?`, in this case) in the list `title`.

I agree that it would be a good idea to eventually have a fix that either:
- encodes URL special characters in the list `title` or
- disallows those characters while editing or on the `Save list title` button click event.

I don't know `circulation-web` well enough to understand the implications of the former from the perspective of either functionality or UI consistency/metaphor, but it is a simple one-line fix, which I could include, assuming that someone with a better understanding of the app could test it properly.

EdwinGuzman:
`cat?` doesn't work for me but `cat!` is fine. The issue discussed should be it's own ticket since it does have implications on how users can build lists, so let's leave that off for now.

Here's the ticket: https://jira.nypl.org/browse/SIMPLY-2729

tdilauro:
> `cat?` doesn't work for me but `cat!` is fine.

The issue is that at least some URL syntax special characters can cause problems because of the way the title is inserted into the feed URL. `!` won't cause a problem because it is not one of these characters. `?`indicates the beginning of query params and, thus, causes problems when the URL is parsed.

> The issue discussed should be it's own ticket since it does have implications on how users can build lists, so let's leave that off for now.
> 
> Here's the ticket: https://jira.nypl.org/browse/SIMPLY-2729

tdilauro:
@adriana-alter If your tests were successful, I think this is ready to be merged.

I seem to have permission to merge, but I will leave that to you and @EdwinGuzman.

adriana-alter:
@tdilauro Thanks--merging now!

PR #308, opened 2020-06-26
Announcements

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2739 and its subtasks.

Goes with https://github.com/NYPL-Simplified/circulation/tree/adjust-announcements-config-setting

Requirements doc: https://docs.google.com/document/d/1ngvqr7oN-VHMdnygDD9uAeVpa45LYq-JUTT5zY5-gJ4/edit?ts=5ec7f043&pli=1#heading=h.t2bf91puc1w

Screen Shot 2020-06-26 at 2 55 50 PM

Note: while making sample announcements for the screenshot, I have just noticed a bug whereby the default end date does not update when you change the start date.  Will fix.  :woman_facepalming:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 ### v0.4.20
 #### Added
-- Implemented a new LibraryEditForm section so that admins can manage announcements.
+- Implemented a new LibraryEditForm section so that admins can manage library-specific announcements.
 
 ### v0.4.19
-#### Added
-- Implemented a new LibraryEditForm section so that admins can manage announcements.
+#### Fixed
+- Fixed a bug preventing scrolling through the available lists on the Lanes page.
 
 ### v0.4.18
 #### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.4.19
+#### Added
+- Implemented a new LibraryEditForm section so that admins can manage announcements.
+
 ### v0.4.18
 #### Removed
 - Removed a reference to the unused `isActive` property of `opds-web-client`'s `router` context.
--- a/src/components/Announcement.tsx
+++ b/src/components/Announcement.tsx
@@ -8,8 +8,9 @@ export interface AnnouncementProps {
   startDate?: string;
   endDate?: string;
   onChange?: () => void;
-  delete?: (content: string) => void;
-  edit?: (content: string) => void;
+  id?: number;
+  delete?: (id: number) => void;
+  edit?: (id: number) => Promise<void>;
 }
 
 export default class Announcement extends React.Component<AnnouncementProps, {}> {
@@ -20,11 +21,11 @@ export default class Announcement extends React.Component<AnnouncementProps, {}>
   }
   edit(e) {
     e.preventDefault();
-    this.props.edit(this.props.content);
+    this.props.edit(this.props.id);
   }
   delete(e) {
     e.preventDefault();
-    this.props.delete(this.props.content);
+    this.props.delete(this.props.id);
   }
   format(date) {
     let [year, month, day] = date.split("-");
--- a/src/components/Announcement.tsx
+++ b/src/components/Announcement.tsx
@@ -1,16 +1,13 @@
 import * as React from "react";
-import EditableInput from "./EditableInput";
-import AnnouncementForm from "./AnnouncementForm";
 import { Button } from "library-simplified-reusable-components";
 
 export interface AnnouncementProps {
-  content?: string;
-  start?: string;
-  finish?: string;
-  onChange?: () => void;
-  id?: number;
-  delete?: (id: number) => void;
-  edit?: (id: number) => Promise<void>;
+  content: string;
+  start: string;
+  finish: string;
+  id: string;
+  delete: (id: string) => void;
+  edit: (id: string) => Promise<void>;
 }
 
 export default class Announcement extends React.Component<AnnouncementProps, {}> {
@@ -19,22 +16,24 @@ export default class Announcement extends React.Component<AnnouncementProps, {}>
     this.delete = this.delete.bind(this);
     this.edit = this.edit.bind(this);
   }
-  edit(e) {
+  edit(e: Event) {
     e.preventDefault();
     this.props.edit(this.props.id);
   }
-  delete(e) {
+  delete(e: Event) {
     e.preventDefault();
     this.props.delete(this.props.id);
   }
-  format(date) {
+  format(date: string): string {
+    // The date is stored in the format year-month-day, because that's the format that the date picker
+    // in AnnouncementForm understands.  But month/day/year looks nicer for display purposes.
     if (date) {
       let [year, month, day] = date.split("-");
       return `${month}/${day}/${year}`;
     }
     return "";
   }
-  render() {
+  render(): JSX.Element {
     let announcement = (
       <section className="announcement-info">
         <section className="dates">
@@ -57,20 +56,16 @@ export default class Announcement extends React.Component<AnnouncementProps, {}>
         className="left-align"
       />
     );
-    let renderAnnouncement = () => {
-      return (
-        <div className="announcement">
-          { announcement }
-          <hr />
-          <section className="buttons">
-            { editButton }
-            { deleteButton }
-          </section>
-        </div>
-      );
-    };
+
     return (
-      renderAnnouncement()
+      <div className="announcement">
+        { announcement }
+        <hr />
+        <section className="buttons">
+          { editButton }
+          { deleteButton }
+        </section>
+      </div>
     );
   }
 }
--- a/src/components/Announcement.tsx
+++ b/src/components/Announcement.tsx
@@ -2,7 +2,6 @@ import * as React from "react";
 import EditableInput from "./EditableInput";
 import AnnouncementForm from "./AnnouncementForm";
 import { Button } from "library-simplified-reusable-components";
-// import { Droppable, Draggable } from "react-beautiful-dnd";
 
 export interface AnnouncementProps {
   content?: string;
@@ -36,7 +35,10 @@ export default class Announcement extends React.Component<AnnouncementProps, {}>
   render() {
     let announcement =
       <section className="announcement-info">
-        <span>{this.props.content}: {this.format(this.props.startDate)} to {this.format(this.props.endDate)}</span>
+        <section className="dates">
+          {this.format(this.props.startDate)} – {this.format(this.props.endDate)}
+        </section>
+        <span>{this.props.content}</span>
       </section>
     let editButton = (
       <Button
@@ -49,12 +51,14 @@ export default class Announcement extends React.Component<AnnouncementProps, {}>
       <Button
         content="Delete"
         onClick={(e) => this.delete(e)}
+        className="left-align"
       />
     )
     let renderAnnouncement = (provided?, snapshot?) => {
       return (
         <li className="announcement">
           { announcement }
+          <hr />
           <section className="buttons">
             { editButton }
             { deleteButton }
@@ -64,7 +68,6 @@ export default class Announcement extends React.Component<AnnouncementProps, {}>
     }
     return (
       renderAnnouncement()
-      // <Draggable draggableId={this.props.content} key={this.props.content}>{(provided, snapshot) => renderAnnouncement(provided, snapshot)}</Draggable>
     )
   }
 }
--- a/src/components/Announcement.tsx
+++ b/src/components/Announcement.tsx
@@ -2,12 +2,13 @@ import * as React from "react";
 import EditableInput from "./EditableInput";
 import AnnouncementForm from "./AnnouncementForm";
 import { Button } from "library-simplified-reusable-components";
-import { Droppable, Draggable } from "react-beautiful-dnd";
+// import { Droppable, Draggable } from "react-beautiful-dnd";
 
 export interface AnnouncementProps {
   content?: string;
   startDate?: string;
   endDate?: string;
+  position?: number;
   onChange?: () => void;
   editable?: boolean;
   delete?: (content: string) => void;
@@ -30,8 +31,11 @@ export default class Announcement extends React.Component<AnnouncementProps, {}>
     this.props.delete(this.props.content);
   }
   render() {
+    // <span className="position">#{this.props.position}</span>
     let announcement =
-      <span>{this.props.content}: {this.props.startDate} to {this.props.endDate}</span>;
+      <section className="announcement-info">
+        <span>{this.props.content}: {this.props.startDate} to {this.props.endDate}</span>
+      </section>
     let editButton = (
       <Button
         content="Edit"
@@ -45,7 +49,7 @@ export default class Announcement extends React.Component<AnnouncementProps, {}>
         onClick={(e) => this.delete(e)}
       />
     )
-    let renderAnnouncement = (provided, snapshot) => {
+    let renderAnnouncement = (provided?, snapshot?) => {
       return (
         <li className="announcement">
           { announcement }
@@ -57,7 +61,8 @@ export default class Announcement extends React.Component<AnnouncementProps, {}>
       )
     }
     return (
-      <Draggable draggableId={this.props.content} key={this.props.content}>{(provided, snapshot) => renderAnnouncement(provided, snapshot)}</Draggable>
+      renderAnnouncement()
+      // <Draggable draggableId={this.props.content} key={this.props.content}>{(provided, snapshot) => renderAnnouncement(provided, snapshot)}</Draggable>
     )
   }
 }
--- a/src/components/Announcement.tsx
+++ b/src/components/Announcement.tsx
@@ -27,7 +27,6 @@ export default class Announcement extends React.Component<AnnouncementProps, {}>
   }
   delete(e) {
     e.preventDefault();
-    console.log(this.props);
     this.props.delete(this.props.content);
   }
   render() {
--- a/src/components/Announcement.tsx
+++ b/src/components/Announcement.tsx
@@ -7,7 +7,7 @@ export interface AnnouncementProps {
   finish: string;
   id: string;
   delete: (id: string) => void;
-  edit: (id: string) => Promise<void>;
+  edit: (id: string) => void;
 }
 
 export default class Announcement extends React.Component<AnnouncementProps, {}> {
--- a/src/components/Announcement.tsx
+++ b/src/components/Announcement.tsx
@@ -5,8 +5,8 @@ import { Button } from "library-simplified-reusable-components";
 
 export interface AnnouncementProps {
   content?: string;
-  startDate?: string;
-  endDate?: string;
+  start?: string;
+  finish?: string;
   onChange?: () => void;
   id?: number;
   delete?: (id: number) => void;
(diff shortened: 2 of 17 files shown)


EdwinGuzman:
I haven't looked too closely at the code yet but here's some things I found while QA-ing:

1. More descriptive UI. The UI is fine but it can use more descriptions. I added some in _red_ in the following screenshot. Going through it with VoiceOver on, there's no indication that it's a list of _announcements_, just that it's a list. The same for the date inputs. It just reads "start" but is ambiguous (for those without eyesight). Still, I think being explicit is good in this case. The `(0/350)` doesn't indicate that that's how many characters a user has written out of the max. What happens to an announcement past the end date? If it just goes away in the UI, it might be good to add a description saying that it'll be removed. We don't want admins forgetting an end date, going back to the UI, and then not seeing their announcements thinking it's still active.

It might also be good to remind the user that just clicking on "submit" or "delete" doesn't actually go through with that process until they click on "submit" for the whole library form at the bottom of the page.

announcements

2. This next part leads into two separate bugs (I think).
- From the screenshot above, I have two announcements [A, B] in the list. I decide to edit the first one A and submit. Now it appears as the second announcement in the list as [B, A]. This is slightly confusing. It probably doesn't affect how those announcements are displayed in the client but I can see some thinking that now the order of how they appear has changed. Is that actually true? Are they stored sequentially so that the client can render them in some order?

- So now, I decide to edit announcement B and it goes back to the initial order [A, B]. I forgot to update something in announcement A so I go and edit it again. Now announcement B disappears:

Screen Shot 2020-06-29 at 11 45 59 AM

Now I click on submit or cancel. If I decide to edit announcement A again, the content of the announcement A disappears:
Screen Shot 2020-06-29 at 11 49 56 AM

This is the other bug I mentioned. It looks like I can replicate this without having to interact with both announcements. If you just click on "edit" and then "submit" or "cancel", and then try to "edit" again, the content disappears. It seems for any announcement, when you independently attempt to edit more than once, the content goes away.

src/components/Announcement.tsx
```diff
@@ -0,0 +1,71 @@
...
+    // in AnnouncementForm understands.  But month/day/year looks nicer for display purposes.
+    if (date) {
+      let [year, month, day] = date.split("-");
+      return `${month}/${day}/${year}`;
+    }
+    return "";
+  }
+  render(): JSX.Element {
+    let announcement = (
+      <section className="announcement-info">
```
EdwinGuzman, line 38: I think these should be `div`s rather than `section`s. I would use a `section` for the main `.announcement` class element (which is a div in this case) since it is semantically grouping all the announcement elements together. Further grouping dates or buttons with a section is not needed.

src/components/AnnouncementForm.tsx
```diff
@@ -0,0 +1,106 @@
...
+      if (!this.state.content || !this.state.start || !this.state.finish) {
+        return true;
+      } else if (this.state.content.length < 15 || this.state.content.length > 350) {
+        return true;
+      }
+      return false;
+    };
+    return (
+      <div className="announcement-form">
+        <EditableInput className={(this.state.content.length < 15 || this.state.content.length >= 350) && "wrong-length"} elementType="textarea" type="text" minLength={15} maxLength={350} value={this.state.content} label="Content (15-350 characters)" optionalText={false} onChange={(e) => this.updateContent(e)} description={`(${this.state.content.length}/350)`} />
```
EdwinGuzman, line 82: This line, and others like line 67, are too long and difficult to read.

EdwinGuzman: You have a chance to put `(this.state.content.length < 15 || this.state.content.length >= 350)` into a variable in this function instead of writing it out twice.

src/components/AnnouncementForm.tsx
```diff
@@ -0,0 +1,106 @@
...
+        <EditableInput
+          type="date"
+          max={this.state.finish}
+          value={this.state.start}
```
EdwinGuzman, line 86: This doesn't seem to work as intended: 
Screen Shot 2020-06-29 at 12 23 43 PM

src/components/AnnouncementForm.tsx
```diff
@@ -0,0 +1,106 @@
...
+  updateContent(content: string) {
+    this.setState({ content });
+  }
+  updateStartDate(start: string) {
+    this.setState({ start });
+  }
+  updateEndDate(finish: string) {
+    this.setState({ finish });
+  }
+  add(e: Event) {
```
EdwinGuzman, line 46: You should add a comment that this is adding the current announcement to the list of announcements in the parent component.

src/components/AnnouncementsSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+import Announcement from "./Announcement";
+
+export interface AnnouncementsSectionProps {
+  setting: SettingData;
+  value?: Array<any>;
+}
+
+export interface AnnouncementsSectionState {
+  currentAnnouncements: Array<AnnouncementData>;
+  editing?: any;
```
EdwinGuzman, line 13: This should be `AnnouncementData | null` or something similar.

src/components/AnnouncementsSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export default class AnnouncementsSection extends React.Component<AnnouncementsSectionProps, AnnouncementsSectionState> {
+  constructor(props: AnnouncementsSectionProps) {
+    super(props);
+    this.addAnnouncement = this.addAnnouncement.bind(this);
+    this.deleteAnnouncement = this.deleteAnnouncement.bind(this);
+    this.editAnnouncement = this.editAnnouncement.bind(this);
+    this.getValue = this.getValue.bind(this);
+    this.state = { currentAnnouncements: Array.isArray(this.props.value) ? this.props.value : [] };
```
EdwinGuzman, line 23: There should be a default null value for `editing`.

src/components/AnnouncementsSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+  addAnnouncement(announcement: AnnouncementData) {
+    let announcements = this.state.currentAnnouncements;
+    this.setState({ currentAnnouncements: announcements.concat(announcement) });
+  }
+  deleteAnnouncement(id: string) {
+    if (window.confirm("This will remove this announcement from your list. Are you sure you want to continue?")) {
+      this.setState({ currentAnnouncements: this.state.currentAnnouncements.filter(a => a.id !== id), editing: null});
+    }
+  }
+  async editAnnouncement(id: string) {
```
EdwinGuzman, line 34: Maybe all the other functions should also be async/await or is there a specific reason why it's only for this function?

src/components/AnnouncementsSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+    await this.setState({ editing: this.state.currentAnnouncements.find(a => a.id === id), currentAnnouncements: this.state.currentAnnouncements.filter(a => a.id !== id) });
+  }
+  renderAnnouncement(a: AnnouncementData): JSX.Element {
+    return (
+      <Announcement
+        key={a.content}
```
EdwinGuzman, line 40: Do you need a key here? Maybe that's why some of the content or announcement is disappearing. `content` changes when you edit it so it's not a good value for `key`.

src/components/AnnouncementsSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+        edit={this.editAnnouncement}
+      />
+    );
+  }
+  renderList(): JSX.Element {
+    return (
+      <ul className="announcements-ul">
+        {
+          Array.isArray(this.state.currentAnnouncements) && this.state.currentAnnouncements.map(a =>
+              <li key={a.content}>
```
EdwinGuzman, line 55: Same as above, use `id` instead.

src/components/AnnouncementsSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+          )
+        }
+      </ul>
+    );
+  }
+  render(): JSX.Element {
+    return (
+      <div className="announcements-section">
+        { this.state.currentAnnouncements.length > 0 && this.renderList() }
+        { this.state.currentAnnouncements.length < 3 && <AnnouncementForm add={this.addAnnouncement} content={this.state.editing?.content} start={this.state.editing?.start} finish={this.state.editing?.finish} /> }
```
EdwinGuzman, line 67: The previous line calls a function to render the DOM. I'd follow the same pattern and make this `AnnouncementForm` into a variable so it's easier to read.

src/components/InputList.tsx
```diff
@@ -66,6 +66,31 @@ export default class InputList extends React.Component<InputListProps, InputList
     // Hide the "Add" button if there are no options left
     let showButton = !this.state.options || this.state.options.length > 0;
     let hasListItems = this.state.listItems && this.state.listItems.length > 0;
+    let element: string | JSX.Element;
+    if (setting.format === "language-code") {
```
EdwinGuzman, line 70: Nice. Better than a double ternary op.

src/components/LibraryEditForm.tsx
```diff
@@ -129,6 +130,21 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
     return forms;
   }
 
+  renderAnnouncements(setting: SettingData, value) {
```
EdwinGuzman, line 133: You're not using `setting` but still passing something in when you call this function. Also, the value you pass in line `135` should be a variable or maybe this `setting` argument.

src/components/LibraryEditForm.tsx
```diff
@@ -129,6 +130,21 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
...
+            description: "announcements",
+            format: "date-range",
+            key: "announcements",
+            label: "Announcements",
+            type: "list"}}
+            value={value && JSON.parse(value)}
+            ref="announcements"
+      />
+    );
+    // value={[{content: "FIRST ANNOUNCEMENT", startDate: "2020-05-19", endDate: "2020-05-21", id: 1}, {content: "SECOND ANNOUNCEMENT", startDate: "2020-06-01", endDate: "2020-08-01", id: 2}]}
```
EdwinGuzman, line 145: If you want value descriptors like this, I'd put it above the function:
```
/**
 * description....
 * Example of `value` argument: .... 
 */
renderAccouncements(...
```

src/components/LibraryEditForm.tsx
```diff
@@ -177,12 +193,22 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
...
+        { fields.map(setting => render(setting))}
       </fieldset>
     );
   }
 
   submit(data: FormData) {
+    let announcements = (this.refs["announcements"] as any).getValue();
+    let announcementList = [];
+    announcements.forEach((a: AnnouncementData) => announcementList.push(a));
+    data?.set("announcements", JSON.stringify(announcementList));
```
EdwinGuzman, line 211: I don't like how now specific form content is being added to the generic `LibraryEditForm` but at a first glance I don't see another option. But it'd be good to think about making this more generic.

adriana-alter:
The latest commit is just to update some things that needed updating--it doesn't address the feedback or require re-reviewing yet.

src/components/LibraryEditForm.tsx
adriana-alter: Aaah, that was just for development purposes so that something would show up before the server code was in place.  Will remove.

src/components/AnnouncementsSection.tsx
adriana-alter: Yes, that was just a placeholder before the id-generating server code was in place--good catch!!

src/components/AnnouncementsSection.tsx
adriana-alter: This is the only one that needs it; the other ones work fine without it, but this only works with it.

EdwinGuzman: That doesn't sound right though. All the functions call `this.setState` which is asynchronous but functions don't typically need the `async` declaration. If you are finding that you need it for something in this case, then there might be an issue elsewhere, too.

src/components/AnnouncementsSection.tsx
```diff
@@ -0,0 +1,74 @@
...
+    let announcements = this.state.currentAnnouncements;
+    this.setState({ currentAnnouncements: announcements.concat(announcement) });
+  }
+  deleteAnnouncement(id: string) {
+    if (window.confirm("This will remove this announcement from your list. Are you sure you want to continue?")) {
+      this.setState({ currentAnnouncements: this.state.currentAnnouncements.filter(a => a.id !== id), editing: null});
+    }
+  }
+  async editAnnouncement(id: string) {
+    await this.setState({ editing: this.state.currentAnnouncements.find(a => a.id === id), currentAnnouncements: this.state.currentAnnouncements.filter(a => a.id !== id) });
```
EdwinGuzman, line 35: It'll be easier to read if the filter and find are placed in variables.

src/components/LibraryEditForm.tsx
EdwinGuzman: Gotcha, but if you do need it for an explanation, that's fine, but it should go in the comment area.

src/components/AnnouncementForm.tsx
adriana-alter: I can't replicate that...

src/components/Announcement.tsx
adriana-alter: Isn't it better to not nest `div`s inside `section`s?  Anyway, the dates/buttons have to be grouped for the CSS.  (We use that pattern in a lot of other places; I'd be willing to go through and figure out an alternative if you think it's an accessibility issue, but that seems like a job for another PR.)

EdwinGuzman: I'm not sure if nesting divs inside sections is an issue. Technically, having multiple nested sections isn't an issue either since the DOM behaves the same way. But semantically it's not grouping one thematic section together but rather small parts within the large thematic section together. It seems like overkill with all the sections. It can be reviewed in another PR if you want.

EdwinGuzman:
I should now be using the the `add-announcements-configuration-setting` branch in `circulation`, right? The other branch got merged in. But now with the last commit (changing `format` to `type`) I see this and it doesn't let me create new announcements:
Screen Shot 2020-07-01 at 9 52 32 PM

I did pull on both `circulation` and `circulation-web` PR branches.

src/components/AnnouncementForm.tsx
EdwinGuzman: If you tab through from the "content" textarea to the "start date" input, you can update the month, day, and year manually and enter "2023". You can also move the up and down arrows and get a negative date.

Trying this again returns an error and doesn't successfully submit the announcement unlike before. So this seems fixed now.

adriana-alter:
Yes, you should be using `add-announcements-configuration-setting`...but it's supposed to be `type`, not `format`.  Something must have accidentally gotten reverted; weird!!  Changing it back now.

adriana-alter:
I haven't addressed the descriptive UI issue or the order of announcements issue yet; next round!

EdwinGuzman:
> order of announcements issue

How big of a bug do you think this is? I _think_ if you don't get to it in this PR it might be okay with a descriptive note that announcements will appear based on their dates rather than order in the UI. But, it might be confusing regardless.

src/components/LibraryEditForm.tsx
```diff
@@ -132,17 +132,12 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
...
   renderAnnouncements(setting: SettingData, value) {
     return (
-      <AnnouncementsSection setting={{
-            description: "announcements",
-            format: "date-range",
-            key: "announcements",
-            label: "Announcements",
-            type: "list"}}
+      <AnnouncementsSection
+            setting={{...setting, ...{format: "date-range"}}}
```
EdwinGuzman, line 136: You can do `{{ ...setting, format: "date-range" }}`. The above is not wrong but the spread is typically used for an object with multiple props.

adriana-alter: The linter doesn't like that.

adriana-alter:
I reordered the announcements; seemed easier than coming up with more copy...!

EdwinGuzman:
Yes, the ordering works well, too, except if there are multiple announcements that have the same start date. In that case, the order will still be "the last one edited appears last on the list".

I'd still include the word "Date" in the label for the start and end date, a short description after "(0/350)", and a short title for the list of scheduled announcements.

The other bug where the content disappeared is fixed.

adriana-alter:
I think that ordering is okay for multiple announcements with the same start date.  Do you think ordering them by end date would be better?

EdwinGuzman:
The start date seems fine but if it turns out to be confusing, then it should be updated later.

adriana-alter:
Re: some of your points from your initial comment on this:

-- To my knowledge, outdated announcements don't automatically get deleted on the back end, therefore don't disappear from the UI.  So that should be okay.

-- I think it's okay not to have a message saying that the changes won't go through until the whole LibraryEditForm is submitted, because there are other things on that form (Languages, Geographic Areas) that follow that pattern, and they don't have a message.  Although...maybe those _should_ have a message??  What do you think?

EdwinGuzman:
That's interesting about the outdated announcements. Will they just stay in the backend indefinitely for now?

Yeah, it might be good to have some sort of copy to let users know that they need to make their updates on individual panels, submit them all individually if more than one panel is being updated, and then submit the whole form. But that should be a different story.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.4.19
+#### Added
+- Implemented a new LibraryEditForm section so that admins can manage announcements.
```
EdwinGuzman, line 5: NBD but maybe "can manage library-specific announcements.".

EdwinGuzman:
I think this is good to go but can you confirm that the backend and the setting object has the correct property? This works for me if I have `} else if (setting.format === "announcements") {` in `LibraryEditForm` but the PR has `setting.type`. The `circulation` branch `add-announcements-configuration-setting` does have the latest code.

adriana-alter:
It is definitely supposed to be `type`, not `format`; you shouldn't have `format` in `LibraryEditForm`.  Pull again?  

My local copy of `circulation` has `type`, but the PR, I have just discovered, has `format`.  Something weird has been going on with that branch such that I've changed it like three times, and somehow it doesn't take. I don't understand how this is possible.  But will try again...

EdwinGuzman:
What file in `circulation` should it have `type`?

CHANGELOG.md
```diff
@@ -1,8 +1,12 @@
 ## Changelog
 
+### v0.4.20
+#### Added
+- Implemented a new LibraryEditForm section so that admins can manage announcements.
+
```
EdwinGuzman, line 6: The message appears twice also under `0.4.19`. It should have the previous message about the scrolling bug fix on the Lanes page.

adriana-alter: Clicked the wrong thing 🤦‍♀️

adriana-alter:
`api/config` should have `type`.  Like so: 
Screen Shot 2020-07-16 at 4 35 36 PM

adriana-alter:
And `api/admin/controller/library_settings`.  And the relevant test files.
Screen Shot 2020-07-16 at 4 36 49 PM

adriana-alter:
Okay, the change to `add-announcements-configuration-setting` seems to have worked this time!  Pull and make sure it doesn't do the weird silently-reverting-the-change thing again?

EdwinGuzman:
Yes, the latest commit in the `circulation` branch is now allowing this to render correctly.

PR #310, opened 2020-07-22
Assign temporary ids

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2895, a bug I found during QA.  

The bug: if you have unsaved announcements (i.e. you have created them but not submitted LibraryEditForm), then trying to edit an announcement causes the unsaved ones to disappear.

Explanation: the logic in `editAnnouncement` (in `AnnouncementsSection`) filters by ID in order to figure out which announcements to keep displaying in the list.  But the ID is assigned by the server.  So if your announcements aren't saved yet, they don't have IDs.

Workaround: I assigned temporary IDs to announcements when they're created.  The temporary IDs get blanked out when you submit the form.  This way, editing works as expected, and the announcements still get the server-generated permanent IDs when you save them.

### DIFF
--- a/src/components/AnnouncementForm.tsx
+++ b/src/components/AnnouncementForm.tsx
@@ -21,6 +21,7 @@ export default class AnnouncementForm extends React.Component<AnnouncementFormPr
   constructor(props: AnnouncementFormProps) {
     super(props);
     this.updateStartDate = this.updateStartDate.bind(this);
+    this.updateEndDate = this.updateEndDate.bind(this);
     let [start, finish] = this.getDefaultDates();
     this.state = {content: this.props.content || "", start: this.props.start || start, finish: this.props.finish || finish};
   }
@@ -92,7 +93,6 @@ export default class AnnouncementForm extends React.Component<AnnouncementFormPr
       } else if (wrongLength) {
         return true;
       }
-      this.updateEndDate = this.updateEndDate.bind(this);
       return false;
     };
     return (
--- a/src/components/AnnouncementForm.tsx
+++ b/src/components/AnnouncementForm.tsx
@@ -21,9 +21,8 @@ export default class AnnouncementForm extends React.Component<AnnouncementFormPr
   constructor(props: AnnouncementFormProps) {
     super(props);
     this.updateStartDate = this.updateStartDate.bind(this);
-    this.updateEndDate = this.updateEndDate.bind(this);
     let [start, finish] = this.getDefaultDates();
-    this.state = {content: "", start: start, finish: finish};
+    this.state = {content: this.props.content || "", start: this.props.start || start, finish: this.props.finish || finish};
   }
   getDefaultDates(): string[] {
     // By default, the start date is today's date and the end date is two months from today.
@@ -76,9 +75,9 @@ export default class AnnouncementForm extends React.Component<AnnouncementFormPr
       this.setState({content: "", start: start, finish: finish});
     }
   }
+
   componentWillReceiveProps(newProps: AnnouncementFormProps) {
     // Switch from creating a new announcement to editing an existing one.
-    // if (newProps.content !== this.props.content) {
     if (newProps.content?.length > 0) {
       const { content, start, finish } = newProps;
       this.setState({ content: content, start: this.formatDate(start), finish: this.formatDate(finish) });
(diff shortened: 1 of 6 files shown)


src/components/AnnouncementForm.tsx
```diff
@@ -93,6 +92,7 @@ export default class AnnouncementForm extends React.Component<AnnouncementFormPr
       } else if (wrongLength) {
         return true;
       }
+      this.updateEndDate = this.updateEndDate.bind(this);
```
EdwinGuzman, line 95: This should be in the constructor instead of in this component.

src/components/AnnouncementsSection.tsx
```diff
@@ -24,14 +24,22 @@ export default class AnnouncementsSection extends React.Component<AnnouncementsS
   }
   addAnnouncement(announcement: AnnouncementData) {
     let announcements = this.state.currentAnnouncements;
+    // If the announcement hasn't been saved yet, it does not have an ID; the ID is assigned by the server code.  But editing
+    // announcements involves finding them by their ID, so we give the new announcement a temporary ID just so that we'll be able
+    // to edit it.  The temporary ID gets blanked out before the announcement goes to the server, so the permanent ID will still get
+    // assigned on the back end.
+    if (!announcement.id) {
+      let tempId = ["temp_1", "temp_2", "temp_3"].find(x => !announcements.map(a => a.id).includes(x));
```
EdwinGuzman, line 32: Is this only allowing up to three announcements to have a temporary id? I don't think a user creating 20 announcements seems likely, but I think it'd be better if you generate a random id.

src/components/AnnouncementsSection.tsx
```diff
@@ -95,6 +103,9 @@ export default class AnnouncementsSection extends React.Component<AnnouncementsS
     );
   }
   getValue(): Array<AnnouncementData> {
-    return this.state.currentAnnouncements;
+    // If there are any new announcements, blank out their temporary IDs so that the server wil assign them permanent ones.
+    return this.state.currentAnnouncements.map(a =>
+      a.id.match(/temp_/) ? ({content: a.content, start: a.start, finish: a.finish} as AnnouncementData) : a
```
EdwinGuzman, line 108: And you can still prepend `temp_` to the randomly generated ids to clear them out.

src/components/AnnouncementsSection.tsx
adriana-alter: You can only ever have three announcements at once.  Once you've hit three, the create form disappears.

EdwinGuzman: Right, I forgot about that. I just realized that this is not mentioned to the user in a description. It's not good to assume that they will know they can create a max of three announcements at a time. They might think it's a bug that the create form disappears.

You mentioned that past announcements in the db don't disappear but will they be returned by the server?

src/components/AnnouncementForm.tsx
adriana-alter: Not sure how that ran away--good catch!

src/components/AnnouncementsSection.tsx
adriana-alter: That's...a really good question.  I'll check.

adriana-alter: Okay, as far as I can tell, you cannot make any more announcements until you get rid of your old announcements--i.e. there is no way there will ever be more than 3 announcements associated with your library (regardless of whether they've expired).

EdwinGuzman: So this means that if I make three announcements for issues in July but want to make a new announcement in September I can't? Even if the older July announcements have been expired?

EdwinGuzman:
Approving. The fix itself and the update are fine. But, just having up to 3 max announcements regardless if they are expired seems like a larger issue, right?

adriana-alter:
I'm not sure why that's an issue, given that we established they _don't_ disappear from the UI when they expire.  If you want to make a new one, you can just delete one of the expired ones.  I mean, I guess in theory someone might want to have a record of all their past announcements...but maybe that should live somewhere different anyway (it seems weird to have things that can no longer be edited live in an edit form)...

PR #314, opened 2020-08-27
Implement date parsing fix

adriana-alter:
Band-aid fix [implemented by Jonathan Green](https://github.com/NYPL-Simplified/circulation-web/pull/313); more comprehensive solution under investigation.

@EdwinGuzman What do you think--publish new package version just for this, or do you know of an upcoming thing it could be bundled with?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.4.22
+#### Fixed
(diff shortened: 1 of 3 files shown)


EdwinGuzman:
It's fine to publish a new version for this since it's a bug that impedes using the CM. It would've been better to have the version and changelog updated in the other PR so others know how to format `circulation-web` PRs, but next time is fine for that.

PR #318, opened 2020-09-11
Check for unnamed items

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3022.

I can't reproduce the bug on local or QA, but my best guess, based on the error message, is that a metadata service object missing the `name` property has somehow gotten in there.  Assuming I'm right: this is a bandaid fix so that the page will load properly and we can see what's going on (I'm giving any unnamed items a default name in the format "Unnamed" + the type of item), but it points to a bigger, and weirder, bug: the `name` field is definitely required, so it should never have been possible to get into this situation in the first place, but...

(The `node-sass` update is not related, but turned out to be necessary; I tried to run the dev server, and got a fun surprise involving tons of `node-sass` error messages making everything crash...)


package.json
```diff
@@ -34,6 +34,7 @@
     "font-awesome": "^4.6.3",
     "isomorphic-fetch": "^2.2.1",
     "library-simplified-reusable-components": "1.3.16",
+    "node-sass": "^4.14.1",
```
EdwinGuzman, line 37: This is fine but should stay in `devDependencies`.

src/components/__tests__/SelfTestsCategory-test.tsx
```diff
@@ -124,4 +124,32 @@ describe("SelfTestsCategory", () => {
     selfTests = wrapper.find(SelfTests).at(0);
     expect(selfTests.prop("sortByCollection")).to.be.true;
   });
+
+  it("handles unnamed items", () => {
+    let nameless =   {
```
EdwinGuzman, line 129: Extra space.

EdwinGuzman (commented):
Great. The problem _was_ the missing name. I updated the Metadata Wrangler and gave it a name so now the problem is fixed in production. Like you said, this shouldn't happen since names are required but older systems don't have that required field fixed, so this patch is fine.

Can you add a comment to the `handleUnnamed` function to explain why it's doing what it's doing?

EdwinGuzman:
And CHANGELOG and version update.

EdwinGuzman (approved):

PR #321, opened 2020-09-17
Adding Prettier

EdwinGuzman:
No specific JIRA ticket for this but I can create one... Edit: [SIMPLY-3094](https://jira.nypl.org/browse/SIMPLY-3094)

It adds prettier and some eslint configs coming from what's in the Library Card App. I added a space in `src/components/SelfTestsCategory.tsx` just so that the pre-commit hook can pick it up because that's a file recently worked on. Let me know what you think. These are general configs but can be modified.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,52 @@
+{
+  "extends": [
+    "eslint:recommended",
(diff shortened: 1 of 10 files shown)


adriana-alter (approved):
Seems good to me!

PR #322, opened 2020-10-02
Trying to get rid of prettier errors...?

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2777...but apparently when I committed it, Prettier messed with a bunch of things, so it looks like there are way more changes than there actually are.  The relevant ones are just the `max_cache_age=0` stuff.

### DIFF
--- a/src/components/CatalogPage.tsx
+++ b/src/components/CatalogPage.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable */
 import * as React from "react";
 import * as ReactDOM from "react-dom";
 const OPDSCatalog = require("opds-web-client");
@@ -56,7 +57,6 @@ export default class CatalogPage extends React.Component<CatalogPageProps, {}> {
 
     const fetcher = new DataFetcher({ adapter });
     const actions = new ActionsCreator(fetcher);
-
     return (
       <>
         <ActionsProvider actions={actions} fetcher={fetcher}>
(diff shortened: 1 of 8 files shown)


src/components/Header.tsx
```diff
@@ -60,109 +65,125 @@ export class Header extends React.Component<HeaderProps, HeaderState> {
...
     let isLibraryManager = this.context.admin.isLibraryManager(currentLibrary);
     let isSystemAdmin = this.context.admin.isSystemAdmin();
-    let isSiteWide = (!this.context.library || !currentLibrary);
+    let isSiteWide = !this.context.library || !currentLibrary;
     let isSomeLibraryManager = this.context.admin.isLibraryManagerOfSomeLibrary();
 
     // Links that will be rendered in a NavItem Bootstrap component.
     const libraryNavItems = [
-      { label: "Catalog", href: "%2Fgroups" },
+      { label: "Catalog", href: "%2Fgroups?max_cache_age=0" },
```
EdwinGuzman, line 92: Is this adding the query to the browser URL as well? Is it just navigating to this URL or actually making the request to the endpoint with that query?

EdwinGuzman (approved):
The prettier updates are fine. I verified that the fetcher is making the request with that param.

src/components/CatalogPage.tsx
```diff
@@ -101,7 +101,9 @@ export default class CatalogPage extends React.Component<CatalogPageProps, {}> {
   }
 
   expandCollectionUrl(url: string): string {
-    return url ? document.location.origin + "/" + url : url;
+    return url
+      ? document.location.origin + "/" + url + "?max_cache_age=0"
```
leonardr, line 105: We're seeing a problem that's happening when `url` already contains a query string. We end up with two query strings. This needs to use a `&` instead of a `?` if `url` already contains a `?`.

vbessonov: I think it's better to use some library (for example, [build-url](https://www.npmjs.com/package/build-url)) rather than trying to build a URL by concatenating strings

PR #323, opened 2020-10-15
Display user's permissions level

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3192; shows user what they're currently logged in as.

Screen Shot 2020-10-14 at 2 01 05 PM

### DIFF
--- a/src/components/__tests__/Header-test.tsx
+++ b/src/components/__tests__/Header-test.tsx
@@ -263,14 +263,23 @@ describe("Header", () => {
       wrapper = mount(<Header />, { context: fullContext, childContextTypes });
 
       let toggle = wrapper.find(".account-dropdown-toggle").hostNodes();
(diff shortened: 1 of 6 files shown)


src/components/__tests__/Header-test.tsx
```diff
@@ -263,14 +263,23 @@ describe("Header", () => {
       wrapper = mount(<Header />, { context: fullContext, childContextTypes });
 
       let toggle = wrapper.find(".account-dropdown-toggle").hostNodes();
-      let dropdownLinks = wrapper.find("ul.dropdown-menu a");
+      let dropdownItems = wrapper.find("ul.dropdown-menu li");
+      let dropdownLinks = wrapper.find("ul.dropdown-menu li a");
```
EdwinGuzman, line 267: This is fine but I think you can also do `dropdownItems.find("a")`.

EdwinGuzman (approved):
Nice!

PR #326, opened 2020-11-05
Input list fix

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3266

I was getting `eslint` errors about string refs being deprecated, but this is time-sensitive, so I temporarily disabled `eslint` for the relevant file, and will file a separate ticket for updating the refs.

Most of the changes showing up in this PR are just Prettier formatting things.  The actual changes are line 46 of `InputList.tsx` (defining the `listItems` state property as `(props.value as string[]) || (props.setting.default as string[]) || []`), and the new test ("renders the default values") at the end of `InputList-test.tsx`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.4.30
+
+#### Fixed
+
+- Fixed a bug whereby the InputList wasn't rendering default values from the server
+
 ### v0.4.29
 
 #### Fixed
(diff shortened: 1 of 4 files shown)


EdwinGuzman (approved):
Looks good. How did you end up finding this bug? The prettier update can be annoying when reading PRs but the code looks more consistent and readable, and thanks for pointing up your specific update.

adriana-alter:
Viacheslav brought it to my attention--he's adding a new configuration setting to the server code, but the default values weren't showing up on the front end in development...

PR #327, opened 2020-11-16
More functional components

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3241

I broke two new (functional) subcomponents out of `CustomListSearch`: `CustomListSearchOptions` (the collapsible panel with the sorting options and language filter) and `CustomListSearchBox` (the search box and the entrypoints radio buttons).

A lot of the changes showing up in this PR are from pulling the new version of `master` into `customlists`.  The relevant stuff is just the CustomLists components (and their tests).

I am getting a whole bunch of unhelpful error messages when I try to commit (mostly complaints about snake case variable names; see screenshot), and had to use the `--no-verify` flag to make the commit go through; have you come across this?  I can't imagine it's from the changes in this branch...
Screen Shot 2020-11-16 at 11 23 41 AM


EdwinGuzman:
To remove the changes from `master`, you can merge `customlists` into this PR branch. Then those differences will not appear since all branches are up-to-date.

I haven't seen those specific changes but that's a good thing. I think you should fix them rather than use the `--no-verify` flag. I would disable the camelcase issue since a lot of data from the server is not in camelcase. It shows errors from `CustomListSearchBox` which you should be able to fix.

src/components/CustomListSearchBox.tsx
```diff
@@ -0,0 +1,61 @@
...
+          value={entryPoint}
+          onChange={() => props.changeEntryPoint(entryPoint)}
+        />
+      )
+    );
+
+    return entryPointsElms;
+  };
+
+  return render();
```
EdwinGuzman, line 60: Why not just return the dom instead of returning the function? This will also fix one of the linting issues in the screenshot you pasted, specifically, that you are using `getEntryPointsElms` in line 23 before defining it. So `getEntryPointsElms` should be moved above the render function or you can just return the DOM instead of defining a function.

src/components/CustomListSearchBox.tsx
```diff
@@ -0,0 +1,61 @@
+import * as React from "react";
+import SearchIcon from "./icons/SearchIcon";
```
EdwinGuzman, line 2: And you're not using this variable here, which is another lint error.

src/components/CustomListSearchOptions.tsx
```diff
@@ -0,0 +1,87 @@
...
+  updateSortBy: (sortBy: string) => void;
+  updateLanguage: (language: string) => void;
+  sortBy: string;
+  library?: LibraryData;
+  languages?: LanguagesData;
+}
+
+export default function CustomListSearchOptions(props?: CustomListSearchOptionsProps) {
+
+  let render = () => {
```
EdwinGuzman, line 16: Yeah, I'd go with the convention of doing `return (<Panel ..... );` instead of declaring a function for it.

EdwinGuzman:
Also, try turning on eslint/prettier in your IDE:
Screen Shot 2020-11-16 at 3 29 35 PM

EdwinGuzman:
Everything works as intended but I think my ES is messed up because I don't get any items back after searching. So just the comments above to fix.

src/components/CustomListSearchOptions.tsx
adriana-alter: How does that work for components with props, though?  All of the examples of this that I'm seeing online don't have props...

EdwinGuzman: It should work the same way. Check out this component: https://github.com/NYPL/nypl-library-card-app/blob/main/src/components/AccountFormFields/index.tsx#L16

Do you get an error when you try this pattern? I don't think the `export default` should get in the way, but if it does you can add `export default CustomListSearchOptions;` at the end of the file and not export this function directly.

EdwinGuzman:
The update looks good. just to verify, did you update this PR branch with the code from `master` or was it the `customlists` branch that has `master` updates? I'd feel better if you merge the latest updates into the working branches, not just to see the specific changes, but so that you don't accidentally merge in something that might cause you problems later. If you think it's okay and you'll fix any merge issues, then I can approve.

adriana-alter:
I've now pulled the `master` changes into `customlists`; not anticipating any major merge issues...hopefully 🙃

EdwinGuzman (approved):
Yeah, it's rough to keep track of a feature branch when there are active changes on the main branch. But it seems you'll be fine

PR #329, opened 2020-11-24
Now it's upset because componentWillReceiveProps has been deprecated

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-2614 (updating string refs in custom lists components to use the React16 refs API; I'm going to make a separate ticket for updating non-custom-lists string refs...)

### DIFF
--- a/src/components/CustomListEditor.tsx
+++ b/src/components/CustomListEditor.tsx
@@ -45,6 +45,8 @@ export default class CustomListEditor extends React.Component<
   CustomListEditorState
 > {
   static listener;
+  private listTitleRef = React.createRef<TextWithEditMode>();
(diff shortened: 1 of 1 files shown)


EdwinGuzman:
What exactly is upset about that?

EdwinGuzman (approved):
Looks good and a bit easier to read too. I'm surprised you also had to do the `UNSAFE_` prefix because I haven't had a linter complain about it (I've gotten warnings but not errors).

adriana-alter:
It was eslint--wouldn't let me commit.  (And adding the `UNSAFE_` prefix seemed better than just turning off the `no-deprecated` rule).

PR #332, opened 2020-12-22
Implement new refs API into...some components

adriana-alter:
I updated the refs in as many of the components as I could.  Conspicuous exceptions:

1) The CustomLists components.  This is because that work has already been done and merged into the `customlists` branch, so there didn't seem to be a point in doing it twice.

2) `EditableInput`, `ServiceEditForm`, `SitewideSettingEditForm`, `LibraryRegistrationForm`.  I ran into major problems indicating that the refs usage in these components isn't compatible with the new refs API.  I'm not convinced refs are even the best way to get the data in all of the places in which these components use them, though; I think that, in several cases, switching over to props/state would be a much safer and less convoluted approach anyway, particularly given that the alternative seems to entail trying to strong-arm the new refs into doing something that they're not really intended to do.  But that's going to involve considerable investigation and refactoring, so (after MUCH FRUSTRATION), I decided it was out of the scope of this branch and am making a separate ticket for it.

### DIFF
--- a/src/components/IndividualAdminEditForm.tsx
+++ b/src/components/IndividualAdminEditForm.tsx
@@ -36,6 +36,13 @@ export default class IndividualAdminEditForm extends React.Component<IndividualA
     settingUp: PropTypes.bool.isRequired,
     admin: PropTypes.object.isRequired as React.Validator<Admin>
   };
+  private emailRef = React.createRef<EditableInput>();
+  private passwordRef = React.createRef<EditableInput>();
+  private systemRef = React.createRef<EditableInput>();
+  private managerAllRef = React.createRef<EditableInput>();
+  private librarianAllRef = React.createRef<EditableInput>();
+  private libraryManagerRef = React.createRef<EditableInput>();
+  private librarianRef = React.createRef<EditableInput>();
 
   constructor(props) {
     super(props);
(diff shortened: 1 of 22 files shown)


src/components/IndividualAdminEditForm.tsx
```diff
@@ -55,7 +62,15 @@ export default class IndividualAdminEditForm extends React.Component<IndividualA
       this.setState({ admin: new Admin(nextProps.item.roles || []) });
     }
     if (nextProps.responseBody && !nextProps.fetchError) {
-      clearForm(this.refs);
```
EdwinGuzman, line 58: Where was  `this.refs` defined or coming from?

src/components/InputList.tsx
```diff
@@ -266,9 +267,9 @@ export default class InputList extends React.Component<
   updateNewItem() {
     let ref =
       this.props.setting.format === "language-code"
-        ? (this.refs["addListItem"] as any).refs["autocomplete"]
-        : (this.refs["addListItem"] as any);
-    this.setState({ ...this.state, ...{ newItem: ref.getValue() } });
+        ? this.addListItemRef.current.autocompleteRef.current
+        : this.addListItemRef.current;
+    this.setState({ ...this.state, ...{ newItem: (ref as any).getValue() } });
```
EdwinGuzman, line 272: This seems complicated enough that the `as any` is okay to do. Maybe later you can try a union of types.

EdwinGuzman (approved):
The code looks good. I'm having a hard time running the circulation project so I can't test this right now but once I get it running I'll manually test it.

src/components/IndividualAdminEditForm.tsx
adriana-alter: It's just from the old way of doing refs; it used to work as a shortcut for accessing all the refs (whereas now, apparently, you have to individually list them out in an array, as per line 65).

PR #334, opened 2021-02-11
Adjust permissions

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3485 and its subtasks.  Goes with the [circulation](https://github.com/NYPL-Simplified/circulation/pull/1556) and [core](https://github.com/NYPL-Simplified/server_core/pull/1239) `adjust-permissions` branches.

I handled this by assigning each setting a numerical `level` property on the back end, and then having the front end calculate a numerical permissions level for the currently logged-in admin.  System admin: 3; library manager: 2; librarian: 1.  Fields are hidden and/or made readonly if the setting's level <= the admin's level (i.e. a level 1 admin can't modify a level 3 setting).

The requirements say that librarians should be able to access the analytics configuration section, but only be able to create/modify Google analytics, not local analytics.  I did this by making it so that librarians can view but not edit existing local analytics configurations (the button on the list of analytics configurations will say "view" instead of "edit," even though it goes to the same place, and all the form fields will be disabled).  LMK if you think it would be better to just disable the "edit" button on the list instead.  (I'm not sure whether there's any utility in librarians being able to look at the existing local analytics...)

I have run into a really weird bug: the tests pass just fine in Travis, but when I try to run them in local, a bunch of them (including the first one in `index-test`) error out like so: `Uncaught UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "#<Object>"`  I'm still working on figuring this out, but it could take a while (I spent all day yesterday on it, without much luck), so I thought we might as well get the review process started in the meantime.


src/components/ConfigTabContainer.tsx
```diff
@@ -59,6 +54,10 @@ export default class ConfigTabContainer extends TabContainer<ConfigTabContainerP
     discovery: DiscoveryServices
   };
 
+  LIBRARIAN_TABS = ["libraries", "analytics"];
+  LIBRARY_MANAGER_TABS = ["libraries", "individualAdmins", "analytics"];
```
EdwinGuzman, line 58: This sets the order for the tabs but does the order matter? If not, maybe you can make `LIBRARIAN_TABS = ["libraries", "analytics"];` be the default and then `LIBRARY_MANAGER_TABS = [...LIBRARY_TABS, "individualAdmins"];` to show "inheritance" in what they can access.

src/components/ConfigTabContainer.tsx
```diff
@@ -70,32 +69,28 @@ export default class ConfigTabContainer extends TabContainer<ConfigTabContainerP
...
-            store={this.props.store}
-            csrfToken={this.props.csrfToken}
-            editOrCreate={this.props.editOrCreate}
-            identifier={this.props.identifier}
-            />
+          store={this.props.store}
+          csrfToken={this.props.csrfToken}
+          editOrCreate={this.props.editOrCreate}
+          identifier={this.props.identifier}
+          />
```
EdwinGuzman, line 81: Weird spacing?

src/components/EditableConfigList.tsx
```diff
@@ -283,11 +307,15 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
-   * true but the logic can be overridden by other classes
+   * true, as long as the user is a system admin, but the logic can be overridden by other classes
    * that inherit GenericEditableConfigList.
    */
   canDelete() {
-    return true;
+    return (this.getAdminLevel() === 3);
+  }
+
+  canEdit(item) {
```
EdwinGuzman, line 317: Where is the `item.level` coming from? I'm guessing that's also new in the server? Can you write a comment explaining that connection?

EdwinGuzman:
That's interesting... the tests run successfully for me locally.

src/components/IndividualAdminEditForm.tsx
```diff
@@ -109,19 +109,19 @@ export default class IndividualAdminEditForm extends React.Component<IndividualA
       <fieldset>
         <legend className="visuallyHidden">Admin information</legend>
         <EditableInput
-          elementType="input"
```
EdwinGuzman, line 112: Did prettier remove the indentation?

src/components/LibraryEditForm.tsx
```diff
@@ -74,6 +78,7 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
             value={this.props.item && this.props.item.name}
             description="The human-readable name of this library."
             error={this.props.error}
+            readOnly={this.props.adminLevel < 2}
```
EdwinGuzman, line 81: Can you use a constant variable instead of `2`? It's hard to know what "2" means in this context.

src/components/LibraryEditForm.tsx
```diff
@@ -43,6 +44,10 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
     }
   }
 
+  adminNotAuthorized(setting): boolean {
```
EdwinGuzman, line 47: Each setting `level` is also from the server, right?

src/components/ServiceEditForm.tsx
```diff
@@ -172,7 +177,8 @@ export default class ServiceEditForm<T extends ServicesData> extends React.Compo
         <EditableInput
           elementType="input"
           type="text"
-          disabled={this.props.disabled}
+          disabled={this.shouldDisable()}
+          readOnly={this.shouldDisable()}
```
EdwinGuzman, line 181: This isn't bad but you can make `this.shouldDisable()` into a variable in the `render` scope and use that instead of calling the function multiple times.

src/components/__tests__/EditableConfigList-test.tsx
```diff
@@ -448,4 +477,14 @@ describe("EditableConfigList", () => {
     expect(alert.text()).to.equal("Self-tests for the things have been moved to the troubleshooting page.");
     expect(alert.find("a").prop("href")).to.equal("/admin/web/troubleshooting/self-tests/thingServices");
   });
+
+  it("figures out what level of permissions the admin has", () => {
+    const libraryManager = new Admin([{ role: "manager", library: "nypl" }]);
+    const librarian = new Admin([{ role: "librarian", library: "nypl" }]);
+    expect(wrapper.instance().getAdminLevel()).to.equal(3);
```
EdwinGuzman, line 484: Does this mean you are considered to be a "system admin" by default?

EdwinGuzman:
> The requirements say that librarians should be able to access the analytics configuration section, but only be able to create/modify Google analytics, not local analytics. I did this by making it so that librarians can view but not edit existing local analytics configurations

As a system admin, I see all the navigation including the sitewide Dashboard, System Configuration, and Troubleshooting tabs. As a library manager, I see the sitewide Dashboard and System Configuration but don't see an analytics tab:
Screen Shot 2021-02-16 at 3 01 38 PM
As a librarian, I don't see the sitewide Dashboard or System Configuration tabs in order to be able to access the analytics tab:
Screen Shot 2021-02-16 at 3 02 08 PM

adriana-alter:
> As a system admin, I see all the navigation including the sitewide Dashboard, System Configuration, and Troubleshooting tabs. As a library manager, I see the sitewide Dashboard and System Configuration but don't see an analytics tab:
> Screen Shot 2021-02-16 at 3 01 38 PM
> As a librarian, I don't see the sitewide Dashboard or System Configuration tabs in order to be able to access the analytics tab:
> Screen Shot 2021-02-16 at 3 02 08 PM

Double-check that you pulled the `core` branch in addition to the `circulation` branch?  (This is definitely working for me...)

src/components/__tests__/EditableConfigList-test.tsx
adriana-alter: No, it just means that this test started out using a system admin for the wrapper's context (line 123).

EdwinGuzman:
Yeah, when I first tested this I used the related branches for `core` and `circulation`. I pull the latest constant changes from today but there's no difference for the second account I'm testing the permissions with.

To test, I'm going into the admin configuration and toggling the "Librarian" and "Library Manager" checkboxes for the second account.

adriana-alter:
I can't duplicate the problem at all.  Would you mind creating a new non-sys-admin account to try it with?  I don't know why that would matter, but that's the only difference between our set-ups that I can imagine; all of my admin accounts were created after this work was already in progress (because of having to set up my computer from scratch after it was fixed), so maybe there's different behavior with a preexisting admin account...? 

(FYI, I've now merged the circ and core branches into `develop`.)

Screen Shot 2021-02-17 at 12 36 47 PM

EdwinGuzman (approved):
Weirdness is gone and works well locally.

EdwinGuzman:
> (the button on the list of analytics configurations will say "view" instead of "edit," even though it goes to the same place, and all the form fields will be disabled). LMK if you think it would be better to just disable the "edit" button on the list instead. (I'm not sure whether there's any utility in librarians being able to look at the existing local analytics...)

It seems fine to view the analytics and not change it. Maybe they can bring up any issues with their manager.

PR #335, opened 2021-02-18
Remove individual admins tab for library managers

adriana-alter:
Unclear why I thought library managers were supposed to have access to the "individual admins" tab, but now that I look back over the requirements document with more caffeine in me, I see that this is not in fact the case 🙃


EdwinGuzman (approved):
This makes sense. Approving but wait for tests to pass before merging.

PR #336, opened 2021-02-19
Form submission bug

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3507 -- a bug that was resulting in a blank screen (sometimes!) on submission of a create form, with the following error message:
Screen Shot 2021-02-04 at 4 31 46 PM

This fix is admittedly kind of hacky.  But the underlying problem seems to be that there are still a few components that haven't been updated to use the new refs API (because they're going to require some fundamental restructuring, not just a quick refactor, and I haven't had a chance to deal with that yet)...so, once I've done that, it will automatically provide a real fix (once everything is updated, it should all go through the first branch of the `if/else` statement; the lines I've just had to modify wouldn't be reachable anymore)...and I'm moving it up my priority list...but in the meantime, we need there to NOT be a blank screen when an admin submits a create form. 

Prettier has seen fit to mess with a whole bunch of things in this file, so it looks like there are a lot more changes than there actually are.  The real change is just the addition of question marks on lines 36 and 37 (checking that `refs[key]` exists).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.4.34
+
+#### Fixed
+
+- Fixed bug resulting in blank screen upon form submission
+
 ### v0.4.33
 
 #### Fixed
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.4.33",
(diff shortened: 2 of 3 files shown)


EdwinGuzman (approved):
That makes sense and hope we can get to a refactor soon to fix these tech debt bugs.

PR #337, opened 2021-02-25
Make basic info settings sys-admin-only

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3556 and https://jira.nypl.org/browse/SIMPLY-3557; it turns out that librarians and library managers weren't supposed to have access to the fields in the "Basic Information" section after all.  

(The "Name" and "Short Name" fields are configured in this branch; the other three fields in this section are taken care of on the back end, in https://github.com/NYPL-Simplified/circulation/pull/1566 and https://github.com/NYPL-Simplified/server_core/pull/1241)

### DIFF
--- a/src/components/__tests__/LibraryEditForm-test.tsx
+++ b/src/components/__tests__/LibraryEditForm-test.tsx
@@ -12,7 +12,7 @@ import PairedMenus from "../PairedMenus";
 import InputList from "../InputList";
 import AnnouncementsSection from "../AnnouncementsSection";
 
-describe("LibraryEditForm", () => {
+describe.only("LibraryEditForm", () => {
   let wrapper;
   let save;
   let libraryData = {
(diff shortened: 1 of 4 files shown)


src/components/__tests__/LibraryEditForm-test.tsx
```diff
@@ -12,7 +12,7 @@ import PairedMenus from "../PairedMenus";
 import InputList from "../InputList";
 import AnnouncementsSection from "../AnnouncementsSection";
 
-describe("LibraryEditForm", () => {
+describe.only("LibraryEditForm", () => {
```
EdwinGuzman, line 15: Remove the `.only`.

EdwinGuzman:
Why can't "Name" and "Short Name" also be configured in the backend to keep it consistent?

adriana-alter:
"Name" and "Short Name" aren't created like the other fields; they're explicitly hard-coded by the front end (lines 74 to 99 of `LibraryEditForm.tsx`.  I don't know why it's like that (it was here before I was!), and it's very possibly worth looking into, but this isn't the branch for that.

EdwinGuzman:
Yeah, that's a bit odd.

EdwinGuzman (approved):

PR #338, opened 2021-03-02
Disabled fields not submitting

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3566

It turns out that the data from input fields with the `disabled` property doesn't get submitted; I switched over to using `readOnly` where applicable.

(The weird-looking commit history is because I accidentally branched off of the footer CSS bug that I'm also working on, rather than off of `master`; disregard!)

### DIFF
--- a/src/components/LibraryEditForm.tsx
+++ b/src/components/LibraryEditForm.tsx
@@ -61,7 +61,7 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
     }
 
     let categories = this.separateCategories(otherFields);
-    const requiresHigherThanLibrarian = 2;
+    const requiresSystemAdmin = 3;
     let basicInfoPanel = (
       <Panel
         id="library-basic-info"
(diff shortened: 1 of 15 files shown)


src/components/LibraryEditForm.tsx
```diff
@@ -202,11 +203,12 @@ export default class LibraryEditForm extends React.Component<LibraryEditFormProp
...
             setting={formatSetting(setting)}
             additionalData={this.props.additionalData}
-            disabled={this.props.disabled || this.adminNotAuthorized(setting)}
+            disabled={this.props.disabled}
             value={settingValue(setting)}
             default={findDefault(setting)}
             error={this.props.error}
-            readOnly={setting.readOnly}
+            disableButton={this.adminNotAuthorized(setting) || this.props.disabled}
+            readOnly={this.adminNotAuthorized(setting) || setting.readOnly}
```
EdwinGuzman, line 211: The `ProtocolFormField` above has different values passed into the `disabledButton` and `readOnly` props. Can you explain a bit why?

EdwinGuzman:
This works well as far as I can tell. I can update fields as a non-sys admin and sys admin and the values reflect when I refresh. The code and prop values are just a bit confusing.

EdwinGuzman (approved):

PR #339, opened 2021-03-04
Stop footer from floating

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3506

I initially noticed the floating footer problem in the context of the welcome page.
Before:
Screen Shot 2021-02-04 at 4 30 57 PM
After:
Screen Shot 2021-03-04 at 4 18 38 PM

But while I was working on it, I noticed that the footer was sometimes floaty on some of the other pages too, so this ended up being a more general fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.4.37
+
+#### Fixed
+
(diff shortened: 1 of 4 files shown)


EdwinGuzman (approved):
Looks good. The rule `body > div > div` isn't very maintainable but a good point to later scope css to specific components.

PR #344, opened 2021-04-26
Position: relative on config pages

adriana-alter:
(Hopefully) resolves https://jira.nypl.org/browse/SIMPLY-3723, the footer placement issue that Tim reported.  But I can't actually reproduce the problem he was having, so hard to say for sure...

### DIFF
--- a/src/stylesheets/footer.scss
+++ b/src/stylesheets/footer.scss
@@ -8,7 +8,7 @@ footer {
   z-index: 1;
   position: absolute;
   bottom: 0;
-  height: 56px; 
+  height: 56px;
   p {
     margin: 2px 0;
     a, a:visited, a:link {
@@ -19,4 +19,8 @@ footer {
       }
     }
   }
+}
+
+.config footer {
+  position: relative;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.4.40
+
+#### Fixed
+
+- Fixed another footer placement bug.
+
 ### v0.4.39
 
 #### Updated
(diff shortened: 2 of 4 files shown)


src/stylesheets/footer.scss
```diff
@@ -20,3 +20,7 @@ footer {
     }
   }
 }
+
+.config footer {
+  position: relative;
+}
```
EdwinGuzman, line 26: Tim pointed out that changing line 9 from `absolute` to `relative` fixed the issue. This fix is similar but it's going to only target the footer in `.config` pages. Is there any reason to not target all footers?

adriana-alter: I don't remember the details, but when I was originally trying to fix the footer issues, I discovered that changing line 9 fixed the `.config` pages but broke other pages...

EdwinGuzman: Can you quickly test to see if the other pages still get broken if you do make that change? After that, the change looks good.

tdilauro: I see what you mean, @adriana-alter. It seems like at least some of the "library" pages are set up differently from those on the "config" side. For example, on the `catalog` page, it seems that the goal is for the header and footer to remain fixed. But if the footer `position` is set to `relative`  then the footer scrolls, too.

adriana-alter: Exactly!  (I believe it's, at least in part, because some of the pages have their styling determined by the OPDS web catalog element, which lives in a different repo.). @EdwinGuzman, I'm figuring this is good to go?

EdwinGuzman (approved):

src/stylesheets/footer.scss
EdwinGuzman: Yep.

PR #345, opened 2021-05-04
List tests

adriana-alter:
Apparently we need to be able to give QA a list of the testable functionality for the admin interface; the best way to do that seems to be by giving them a list of our current tests.  I made a custom Mocha reporter so that, rather than someone having to put the list together manually, you can just run `npm run test-list` to generate a formatted, up-to-date list of tests, grouped by component, like so: 
Screen Shot 2021-05-04 at 1 25 46 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.4.41
+
+#### Added
+
+- Added a custom Mocha reporter so that a formatted list of suites and tests can be generated via the CLI.
+
 ### v0.4.40
 
 #### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "simplified-circulation-web",
-  "version": "0.4.38",
+  "version": "0.4.40",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "simplified-circulation-web",
-      "version": "0.4.38",
+      "version": "0.4.40",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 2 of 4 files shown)


EdwinGuzman:
This is... interesting. Do you know how QA will use this list? It doesn't seem like knowing or testing `BookDetailsContainer` may be useful outside a unit test context but it could be useful.

leonardr:
We want these integration tests for the admin interface. QA has written integration tests for the SimplyE app, which gives us pretty good integration coverage of the circulation manager that SimplyE is talking to, but we don't have anything for the admin interface.

This becomes a pressing need right now because we touched all the code (and all the unit tests!) during the Python 3 conversion. It's pretty easy to verify that we didn't change anything that broke SimplyE, but more difficult to verify that we didn't break the admin interface.

EdwinGuzman (approved):
Yeah, that makes sense. I'll approve since it's a pressing need but I do think writing to a file would be easier to keep around in the repo and to share.

leonardr:
I think some of these controllers don't map on to integration test cases, but a lot of them do -- "BookCoverEditor shows book title" and so on.

leonardr:
If you want to just give me a copy of this file to pass on to QA, that will remove the pressing need to get this into the repo ASAP.

adriana-alter:
I couldn't get the formatting to work properly when I tried to have it write to a file; suggestions welcome!  In general, though, I like the idea of giving QA a CLI command rather than (or at least in addition to) a file; that way they can always make sure they have the most up-to-date version of the list, instead of giving us the extra overhead of always remembering to give them an updated file whenever something changes...

PR #353, opened 2021-08-05
Simply 3564/upgrade react methods

oliviawongnyc:
## Description
- Prepended all deprecated React lifecycle methods (componentWillMount and componentWillReceiveProps) with UNSAFE_.
- Added rules to ignore to .eslintrc file to be able to make commits (will address these rules, and update the code properly in another ticket/PR).

## Motivation and Context
- Using deprecated components causes the console to be crowded with warnings.
- Easily being able to see which methods we are using are deprecated may make debugging faster.

## How Has This Been Tested?
- All tests still pass.

### DIFF
--- a/src/components/EditableConfigList.tsx
+++ b/src/components/EditableConfigList.tsx
@@ -132,7 +132,7 @@ export abstract class GenericEditableConfigList<
     const ExtraFormSection = this.ExtraFormSection;
     const itemToEdit = this.itemToEdit();
     return (
-      <div className={this.getAdditionalContentClassName()}>
+      <div className={this.getClassName()}>
         <h2>{headers["h2"]}</h2>
         {canListAllData && this.links && this.links["info"] && (
           <Alert bsStyle="info">{this.links["info"]}</Alert>
@@ -278,7 +278,7 @@ export abstract class GenericEditableConfigList<
     return { h2, h3 };
   }
 
-  getAdditionalContentClassName() {
+  getClassName(): string {
     const className = this.AdditionalContent ? "has-additional-content" : "";
     return className;
   }
--- a/src/components/EditableConfigList.tsx
+++ b/src/components/EditableConfigList.tsx
@@ -36,7 +36,10 @@ export interface EditableConfigListOwnProps {
   settingUp?: boolean;
 }
 
-export interface EditableConfigListProps<T> extends EditableConfigListStateProps<T>, EditableConfigListDispatchProps<T>, EditableConfigListOwnProps {}
+export interface EditableConfigListProps<T>
+  extends EditableConfigListStateProps<T>,
+    EditableConfigListDispatchProps<T>,
+    EditableConfigListOwnProps {}
 
 export interface EditFormProps<T, U> {
   item?: U;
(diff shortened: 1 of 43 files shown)


src/components/EditableConfigList.tsx
```diff
@@ -249,18 +273,20 @@ export abstract class GenericEditableConfigList<T, U, V extends EditableConfigLi
...
-    let h2 = `${this.getItemType()} configuration`;
-    let h3 = `Create a new ${this.itemTypeName}`;
+    const h2 = `${this.getItemType()} configuration`;
+    const h3 = `Create a new ${this.itemTypeName}`;
     return { h2, h3 };
   }
 
-  getClassName() {
-    let className = this.AdditionalContent ? "has-additional-content" : "";
+  getAdditionalContentClassName() {
```
EdwinGuzman, line 281: Did you update this?

src/components/IndividualAdmins.tsx
```diff
@@ -20,21 +27,25 @@ export class IndividualAdmins extends EditableConfigList<IndividualAdminsData, I
...
+      : "Individual admin configuration";
+    const h3 = this.props.settingUp
+      ? "Set up your system admin account"
+      : "Create a new individual admin";
     return { h2, h3 };
   }
 
-  getClassName() {
-    let className = this.props.settingUp ? "set-up" : "";
+  getSetUpClassName() {
```
EdwinGuzman, line 47: I noticed you changed this in `EditableConfigList.tsx` and renamed it `getAdditionalContentClassName`. This `IndividualAdmins` class extends the class exported by `EditableConfigList.tsx`. As far as I can tell, `getSetUpClassName` doesn't get called anywhere in this file. Since this class is extending an existing class, the function name needs to be the same as the parent's function.

In `EditableConfigList`, `getClassName` is defined and it is also invoked. In this file, it is only re-defined as it will be called in the `render` function from the parent. In other words, by renaming this to something the parent class doesn't expect, it won't ever be called. So either rename it to `getAdditionalContentClassName` or revert back to `getClassName`. Whichever you do, make sure you cover the other class components that inherit from the base class.

src/components/Lane.tsx
```diff
@@ -56,37 +63,44 @@ export default class Lane extends React.Component<LaneProps, LaneState> {
...
+  UNSAFE_componentWillReceiveProps(newProps) {
+    this.setState({
+      expanded: this.state.expanded,
+      visible: newProps.lane.visible,
+    });
   }
 
   toggleVisible(): void {
-    const change: boolean = !this.state.visible;
+    const change = !this.state.visible;
```
EdwinGuzman, line 103: Did the linter remove these?

EdwinGuzman:
Just have the other question about the linter removing explicit types.

src/components/Lane.tsx
oliviawongnyc: Yeah, that was the linter for some reason. I can try adding back!

EdwinGuzman (approved):
The update looks good and a quick spot-check worked fine. I originally asked about the type being removed in the `Lane.tsx` file. Not a big deal but if you can add back any types that were removed, that'd be great. Still approving as it's not a blocker.

PR #356, opened 2021-08-11
Simply 2102/pull through book cover url

oliviawongnyc:
## Description

- Added a value to the book cover URL EditableInput in the BookCoverEditor component.

## Motivation and Context

- Prior to this update, the book cover URL field started out blank for the user, so they weren't able to see the current URL.

## How Has This Been Tested?

- Modeled this change after another EditableInput that already had a value attribute.
- Checked that the field was being displayed in local host.
- Ensured all tests are still passing.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,11 @@
 ## Changelog
 
-### 08/10/2021
+### v0.5.3
 
-#### Updated (to be included in next version)
+#### Updated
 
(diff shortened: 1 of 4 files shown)


EdwinGuzman (approved):
The update and the changelog look good. We're having trouble testing this locally but once it's ready to be tested, make sure that updating the cover through the URL or the cover image does update the form as expected.

PR #364, opened 2021-10-04
Simply 3918/refactor custom lists sidebar

oliviawongnyc:
## Description

- Refactored `CustomListsSidebar` 
- Added new context specifically for the List Manager
- Added new tests and adjusted old tests for `CustomListsSidebar`
- Added tests for for the new context

## Motivation and Context

- Addresses [SIMPLY-3918|https://jira.nypl.org/browse/SIMPLY-3918]
- Brings `CustomListsSidebar` component up-to-date with current best practices

## How Has This Been Tested?

- Updated old tests and added new ones for new code, all 888 unit tests are passing
- Checked the UX in localhost to make sure side bar functionality still works

## Checklist:


src/components/CustomListsSidebar.tsx
```diff
@@ -111,5 +91,24 @@ export default class CustomListsSidebar extends React.Component<
...
+        className="btn create-button"
+        to={"/admin/web/lists/" + library + "/create"}
+      >
+        Create New List
+      </Link>
+
+      {lists && lists.length > 0 && (
+        <div>
+          {renderSortButtons()}
+          <ul>{lists.map((list) => renderListInfo(list))}</ul>
```
samanthaandrews, line 109: First off, just want to say this is really great work. It's already so much clearer and I like the addition of ListManagerContext.

I'm thinking that we should change these two lines into functional React components. For example, these two lines would change to look something like:

```
<SortButtons />
<ul>
    <ListInfo list={list} />
 </ul>
```
and then each of those components would be refactored into something like this:

```
const ListInfo = (list: CustomListData) => {
  const whatever = bla bla bla;      
      return (
        <div>
        ..... component rendering here
        </div>
        );
  }
```
The reason I advocate for this change is explained by this Kent C. Dodds [blog post](https://kentcdodds.com/blog/dont-call-a-react-function-component/) and has to do with the React lifecycle hooks, and it _could_ cause buggy state problems in the future even if it's not happening now. We want to be _rendering_ functional components instead of _calling_ them as functions.

src/components/CustomListsSidebar.tsx
```diff
@@ -2,75 +2,55 @@ import * as React from "react";
...
               disabled={false}
             />
           );
         })}
       </fieldset>
     );
-  }
+  };
+
+  const renderListInfo = (list: CustomListData) => {
```
samanthaandrews, line 51: we can destructure the `list` object here

`const { name, id, entry_count } = list`

and then within the return, we don't need to write `list.id`, `list.name`, etc.

they become just `id`, `name`, `entry_count`.

this is more of a nit (something that is more of a stylistic preference on my end) since it's slightly more DRY than the current implementation. i try to use object or array destructuring whenever possible.

EdwinGuzman:
I haven't checked the code yet but went through the UI.

Something I noticed:
- I have two lists. I switch between lists by clicking on "Edit" for each list.
- The "Save this list" button is enabled. I haven't made any changes so this shouldn't be enabled.
- I then make a change (either update the title or add/remove a book in the list) and click on "Save this list".
- Now the "Save this list" button is disabled, which is expected.

Maybe the init state should start off as disabled.

Also, and I think this might be outside the scope of this PR, but a message about adding at least one title in the list would be good somewhere -- so users know the minimum needed to create a list.

src/components/CustomLists.tsx
```diff
@@ -78,12 +78,6 @@ export class CustomLists extends React.Component<
   CustomListsProps,
   CustomListsState
 > {
-  context: { admin: Admin };
```
EdwinGuzman, line 81: Nit: since this file is no longer using `Admin`, you can remove the import on line 20.

src/components/CustomLists.tsx
```diff
@@ -69,7 +69,7 @@ export interface CustomListsProps
     CustomListsOwnProps {}
 
 export interface CustomListsState {
-  sort: string;
+  sort: "asc" | "desc";
```
EdwinGuzman, line 72: You can make this a `type` or `interface` and reuse it in `CustomListsSidebar`.

samanthaandrews (changes requested):
Let's change those components to render instead of calling as a function and this should be good to merge 👍

src/components/CustomListsSidebar.tsx
EdwinGuzman: Yeah, `renderListInfo` is a big enough function that it can be its own component. If you do make this update, note that then you won't need the `admin` context in this `CustomListsSidebar` component but rather only in the new component you create. You'll also have to pass in more props to this new component.

I think you'll find that this update reduces the props that `CustomListsSidebar` uses. So you'll end up doing some prop drilling (such as the `identifier` prop which is only used in `renderListInfo` which is what is being considered for its own component).

src/components/CustomListsSidebar.tsx
```diff
@@ -2,75 +2,55 @@ import * as React from "react";
...
-  sortOrder: string;
+  sortOrder: "asc" | "desc";
 }
 
-export default class CustomListsSidebar extends React.Component<
-  CustomListsSidebarProps,
-  {}
-> {
-  render(): JSX.Element {
+export default function CustomListsSidebar({
```
EdwinGuzman, line 19: Now that you know more about this component, write a short comment explanation for it!
Use the comment block syntax:

```
/**
 * This component ...
 */
export default function CustomListsSidebar({
```

src/components/__tests__/CustomListsSidebar-test.tsx
```diff
@@ -115,10 +121,27 @@ describe("CustomListsSidebar", () => {
...
 
   it("displays the delete button only to library managers", () => {
-    let deleteButtons = wrapper.find(".custom-list-buttons button");
+    const deleteButtons = wrapper.find(".custom-list-buttons button");
     expect(deleteButtons.length).to.equal(2);
-    wrapper.setProps({ isLibraryManager: false });
-    deleteButtons = wrapper.find(".custom-list-buttons button");
+  });
+
+  it("does not display the delete button to non-managers", () => {
```
EdwinGuzman, line 128: 👍

samanthaandrews (approved):
LGTM 👍

PR #365, opened 2021-10-07
Simply 3922/refactor custom list page

oliviawongnyc:
## Description

- Refactored CustomListPage
- Added editorStore + csrfToken to new ListManager context
- Added new tests and adjusted old tests

## Motivation and Context

- Addresses [SIMPLY-3918|https://jira.nypl.org/browse/SIMPLY-3918]
- Brings CustomListPage component up-to-date with current best practices

## How Has This Been Tested?

- Updated old tests and added new ones for new code, all 890 unit tests are passing
- Checked the UX in localhost to make sure side bar functionality still works

### DIFF
--- a/src/components/ContextProvider.tsx
+++ b/src/components/ContextProvider.tsx
@@ -6,7 +6,10 @@ import { PathFor } from "../interfaces";
 import { State } from "../reducers/index";
 import Admin from "../models/Admin";
 import PathForProvider from "opds-web-client/lib/components/context/PathForContext";
-import { ListManagerProvider } from "./ListManagerContext";
+import {
+  ListManagerProvider,
+  ListManagerProviderProps,
+} from "./ListManagerContext";
 
 export interface ContextProviderProps extends React.Props<{}> {
   csrfToken: string;
@@ -17,7 +20,7 @@ export interface ContextProviderProps extends React.Props<{}> {
     role: string;
     library?: string;
   }[];
-  listManagerProps?: any;
+  listManagerProps?: ListManagerProviderProps;
 }
 
 /** Provides a redux store, configuration options, and a function to create URLs
@@ -29,7 +32,7 @@ export default class ContextProvider extends React.Component<
   store: Store<State>;
   admin: Admin;
   pathFor: PathFor;
-  listManagerProps: any;
+  listManagerProps: ListManagerProviderProps;
 
   constructor(props) {
     super(props);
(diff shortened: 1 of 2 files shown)


src/components/ListManagerContext.tsx
```diff
@@ -1,16 +1,19 @@
...
 export const ListManagerContext = React.createContext(null);
 
 export const ListManagerProvider: React.FC<any> = ({
   children,
   roles,
   email,
+  csrfToken,
 }) => {
   const admin = new Admin(roles || [], email || null);
+  const editorStore = buildStore();
```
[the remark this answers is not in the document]
EdwinGuzman: Yeah, good point. There should only be one store and that will make this refactor trickier to implement.

EdwinGuzman:
This is based on @samanthaandrews ' comment about only creating one store in the app.

Here's what you can do (which I tested manually and tests pass and the UI seems okay): Declare the `ListManagerProvider` component inside of `ContextProvider`. This way, you can pass the already created store.

* remove `ListManagerProvider` from `src/index.tsx`
* in `ListManagerContext`, declare `editorStore` as an incoming prop and pass it to the `value` prop.
* in `ContextProvider`, build the props to pass to `ListManagerProvider` and render it in the return function.

I'll create a PR with these changes which I think will help.

EdwinGuzman:
Here: https://github.com/NYPL-Simplified/circulation-web/pull/366
You don't have to merge that PR. It's meant more for an option around building only one store. I just won't be around for a while so this may be easier to review and check.

src/components/__tests__/CustomLists-test.tsx
```diff
@@ -180,7 +189,12 @@ describe("CustomLists", () => {
       expect(error.length).to.equal(0);
 
       wrapper.setProps({
-        fetchError: { status: 500, response: "Error", url: "url" },
+        children: (
```
EdwinGuzman, line 192: This is still weird to look at but it makes sense. I guess it's just the limitation of Enzyme and the new context API.

src/components/ContextProvider.tsx
```diff
@@ -16,6 +17,7 @@ export interface ContextProviderProps extends React.Props<{}> {
     role: string;
     library?: string;
   }[];
+  listManagerProps?: any;
```
samanthaandrews, line 20: can we turn this into a more specific type?

src/components/ContextProvider.tsx
```diff
@@ -27,6 +29,7 @@ export default class ContextProvider extends React.Component<
   store: Store<State>;
   admin: Admin;
   pathFor: PathFor;
+  listManagerProps: any;
```
samanthaandrews, line 32: same here

PR #369, opened 2021-11-09
Simply 3939/refactor custom list entries editor

oliviawongnyc:
## Description

- Refactored the `CustomListEntriesEditor` component to bring it up-to-date with current best practices. 
- Added `CustomListSearchResults` and `CustomListEntries`. 
- Added/edited tests where needed.

## Motivation and Context

- Work towards List Manager refactor outlined in [TAD|https://docs.google.com/document/d/1QhlTh9FLUAWJC9BcmSDgu7m972xyaBoHh1wcgnorq2o/edit#]

## How Has This Been Tested?

- Clicked through local host in the browser
- Unit tests are all passing.

## Checklist:

- [x] I have updated the documentation accordingly. <-- added some in-line documentation for new functionality


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 #### Refactored
 
+- Refactored the `CustomListEntriesEditor` component to bring it up-to-date with current best practices. Added `CustomListSearchResults` and `CustomListEntries`. Added/edited tests where needed.
```
EdwinGuzman, line 7: Nit: newest updates go last in this list.

src/components/CustomListEditor.tsx
```diff
@@ -140,25 +301,64 @@ export default function CustomListEditor({
...
+     */
+    const totalEntriesServer = parseInt(entryCount, 10);
+    if (
+      entryCount &&
+      totalEntriesServer !==
+        totalEntriesServer - deletedListEntries.length + addedListEntries.length
+    )
+      return true;
+
     return false;
```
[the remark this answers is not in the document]
EdwinGuzman: Interestingly enough, clicking on "Cancel changes" in the state above returns the display back to how it was before but nothing else changes:
Screen Shot 2021-11-09 at 5 59 42 PM

EdwinGuzman (commented):
I'm halfway through the review and I'll continue tomorrow.

One thing I noticed while testing which _may_ be related to this PR but let me know if it isn't:
- I have two books in my list and one is "The Art of War".
- I now search for "car".
- I decide to remove "The Art of War" and add a car-related book to my list.
- "The Art of War" is moved to the search results list but maybe it shouldn't since it's technically not a search result entry. (I'm not sure if this is an existing functionality).
- If I refresh the page and search for "car" again, then "The Art of War" shows up!

I'm thinking:
1 - where _should_ deleted entries from a list go? Should they be moved back to the "search results" list even though the search keyword may not match the book?
2 - I'm concerned that searching for a keyword brings back the book into the search result. It's probably cached in elastic search and not the browser. I tested this by opening up a new incognito window and searching again for "car" and "The Art of War" _still_ shows up. We can go more into this later but first check if this happens in production.

CHANGELOG.md
oliviawongnyc: K!

src/components/CustomListEditor.tsx
```diff
@@ -52,42 +53,75 @@ export default function CustomListEditor({
...
-      data.append("deletedEntries", JSON.stringify(deletedListEntries));
-      const collections = draftCollections.map((collection) => collection.id);
-      data.append("collections", JSON.stringify(collections));
+    }
+  }, [list]);
 
-      editCustomList(data, listId && String(listId));
+  React.useEffect(() => {
+    /**
+     * If the addedListEntries changed as a result of the setAddedListEntries() above,
```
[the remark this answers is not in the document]
oliviawongnyc: Do you mean the comment should make mention of those other functions? Or the useEffect function itself should do something if `addedListEntries` changes, but `loadedMoreEntries` is false (like it would be if those other functions ran)?

src/components/CustomListEditor.tsx
```diff
@@ -96,19 +130,144 @@ export default function CustomListEditor({
...
+        newEntries.unshift(entry);
+        setDraftEntries(newEntries);
+      }
+    }
+    /**
+     * Create an array containing the added entries, without the most recent added entry.
+     */
+    const prevAdded = addedListEntries.filter((entry) => entry.id !== id);
+    const newAdded = prevAdded.concat(entry);
+    setAddedListEntries(newAdded);
```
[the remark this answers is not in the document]
oliviawongnyc: This line of code on 163 and the one described on line 167 below are relics of the old components (see [here](https://github.com/NYPL-Simplified/circulation-web/blob/47d1332eddba3a2536585ae942afb60b8008eafe/src/components/CustomListEntriesEditor.tsx#L509) and [here](https://github.com/NYPL-Simplified/circulation-web/blob/47d1332eddba3a2536585ae942afb60b8008eafe/src/components/CustomListEntriesEditor.tsx#L528)). I thought about why these steps are here quite a bit before deciding to keep them in because a) I figured it couldn't hurt to ensure duplicates were not put into the added or deleted arrays and b) by using the filter, we're setting state to a new array based on the previous state as opposed to modifying the state directly. Though, the latter could be achieved in other ways of course.

src/components/CustomListEditor.tsx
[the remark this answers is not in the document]
oliviawongnyc: Edwin! Quit finding issues with this display count, would ya?! 😂 

I believe I fixed this issue by adding the following if statement to the `deleteEntry` function:

```
 if (list.books.filter((book) => book.id === id).length) {
      const prevDeleted = deletedListEntries.filter((entry) => entry.id !== id);
      const newDeleted = [...prevDeleted, ...deletedEntry];
      setDeletedListEntries(newDeleted);
    }
```
Here, we are only updating the `deletedListEntries` state if the book has already been saved in the server. Otherwise, if it was just recently added (without saving), we can just delete from the visible entries list and not add it to the `deletedListEntries`. Then, it's like it never happened. I think it works. Does it make sense to you?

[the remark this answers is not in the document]
EdwinGuzman: Yes, the last commit seemed to have fixed both issues.

src/components/CustomListSearchResults.tsx
```diff
@@ -0,0 +1,173 @@
...
+import AddIcon from "./icons/AddIcon";
+import { Button } from "library-simplified-reusable-components";
+import CatalogLink from "opds-web-client/lib/components/CatalogLink";
+import { getMedium, getMediumSVG } from "opds-web-client/lib/utils/book";
+
+export interface Entry extends BookData {
+  medium?: string;
+}
+
+export interface CustomListEntriesEditorProps {
```
hokei, line 16: Strange interface name?

src/components/CustomListEditor.tsx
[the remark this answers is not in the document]
EdwinGuzman: That the comment should either say all the functions that change `addedListEntries` or not mention any specific functions. This makes it seem like it's only `setAddedListEntries` that updates or changes `addedListEntries`.

src/components/CustomListEditor.tsx
```diff
@@ -96,19 +130,147 @@ export default function CustomListEditor({
...
+  };
+
+  const deleteAll = () => {
+    /**
+     * Create a copy of the draftEntries.
+     */
+    const entriesToDelete = draftEntries.slice();
+
+    setDraftEntries([]);
+    setDeletedListEntries(deletedListEntries.concat(entriesToDelete));
```
EdwinGuzman, line 258: Not a big deal, but `.concat` returns a new array. I'm curious if you need to do `draftEntries.slice` call above. Does it not not work if you remove that and do `setDeletedListEntries(deletedListEntries.concat(draftEntries))`?

src/components/__tests__/CustomListEditor-test.tsx
```diff
@@ -67,6 +131,27 @@ describe("CustomListEditor", () => {
     search = stub();
     loadMoreSearchResults = stub();
     loadMoreEntries = stub();
+
+    childContextTypes = {
+      pathFor: PropTypes.func.isRequired,
+      router: PropTypes.object.isRequired,
+    };
+    fullContext = {
+      ...{
```
EdwinGuzman, line 140: I don't think you need this. Can you pass what's inside directly?

src/components/__tests__/CustomListEntriesEditor-test.tsx
```diff
@@ -149,945 +83,67 @@ describe("CustomListEntriesEditor", () => {
...
-      <CustomListEntriesEditor
-        searchResults={searchResultsData}
-        loadMoreSearchResults={loadMoreSearchResults}
-        loadMoreEntries={loadMoreEntries}
-        isFetchingMoreSearchResults={false}
-        isFetchingMoreCustomListEntries={false}
-      />,
-      { context: fullContext, childContextTypes }
-    );
+  // Skipping tests related to drag-and-drop functionality until the
```
EdwinGuzman, line 109: 👍

src/components/__tests__/CustomListEntriesEditor-test.tsx
```diff
@@ -149,945 +83,67 @@ describe("CustomListEntriesEditor", () => {
...
-    searchEntries = wrapper.find(".custom-list-search-results li");
-    expect(searchEntries.length).to.equal(newSearchResultsData.length);
+  it.skip("shows message in place of search results when dragging from list entries", () => {
+    // simulate starting a drag from entries
+    // find search results container and check that isDropDisabled is false
+    // expect message to equal "Drag books here to remove them from the list."
+    // if you drop anywhere on the page, the message goes away
+    // simulate dropping outside a droppable (no destination)
+    // find search results container and check that isDropDisabled is true
+    // message length should be 0
```
EdwinGuzman, line 147: 👍 for the pseudo-code comments. Should make it easier to remember what to test later.

EdwinGuzman (approved):
Some nitpicky things but I think it's fine to move on. Lots of big changes and something that stuck out to me is that the drag-and-drop package is now being imported and used in multiple files. It's not an issue but it can be confusing why it's called in multiple components and test files. Also, some of the names are confusing such as `CustomListEditor`, `CustomListEntries`, `CustomListEntriesEditor`.

src/components/CustomListEditor.tsx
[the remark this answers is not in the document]
oliviawongnyc: Oh, I see. Will clarify in comment.

src/components/CustomListEditor.tsx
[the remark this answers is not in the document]
oliviawongnyc: I removed the extra filter and all seems to be working fine still.

src/components/CustomListEditor.tsx
```diff
@@ -96,19 +130,144 @@ export default function CustomListEditor({
...
+      }
+    }
+    /**
+     * Create an array containing the added entries, without the most recent added entry.
+     */
+    const prevAdded = addedListEntries.filter((entry) => entry.id !== id);
+    const newAdded = prevAdded.concat(entry);
+    setAddedListEntries(newAdded);
+    /**
+     * Set deleted entries to array containing all except the one being added.
```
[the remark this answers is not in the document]
oliviawongnyc: Same as above.

src/components/CustomListSearchResults.tsx
oliviawongnyc: Fixed!

src/components/CustomListEditor.tsx
oliviawongnyc: I think this works! Will amend.

src/components/__tests__/CustomListEditor-test.tsx
oliviawongnyc: Like this or something else?

```
    wrapper = Enzyme.mount(
      <CustomListEditor
        languages={languages}
        library={library}
        list={listData}
        listCollections={listCollections}
        listId="1"
        editCustomList={editCustomList}
        search={search}
        loadMoreSearchResults={loadMoreSearchResults}
        loadMoreEntries={loadMoreEntries}
        isFetchingMoreSearchResults={false}
        isFetchingMoreCustomListEntries={false}
        searchResults={searchResultsData}
      />,
      {
        context: {
          ...{
            pathFor: stub().returns("url"),
            router: {
              createHref: stub(),
              push: stub(),
              isActive: stub(),
              replace: stub(),
              go: stub(),
              goBack: stub(),
              goForward: stub(),
              setRouteLeaveHook: stub(),
            },
          },
        },
        pathFor: PropTypes.func.isRequired,
        router: PropTypes.object.isRequired,
      }
    );
```

oliviawongnyc: If so, this somehow results in an error while testing:

```
  1) CustomListEditor
       "before each" hook for "renders Header and Body components":
     TypeError: this.context.pathFor is not a function
```

EdwinGuzman: No, I meant more like: 
```
// before
const fullContext = {
    ...{
      pathFor: stub().returns("url"),
      router: {
        createHref: stub(),
        push: stub(),
        isActive: stub(),
        replace: stub(),
        go: stub(),
        goBack: stub(),
        goForward: stub(),
        setRouteLeaveHook: stub(),
      },
    },
  };
// after
const fullContext = {
    pathFor: stub().returns("url"),
    router: {
      createHref: stub(),
      push: stub(),
      isActive: stub(),
      replace: stub(),
      go: stub(),
      goBack: stub(),
      goForward: stub(),
      setRouteLeaveHook: stub(),
    },
  };
```

Your example is off because you also have `pathFor` and `router` outside of the `context` key, no?

```
{
        context: {
          ...{
            pathFor: stub().returns("url"),
            router: {
              createHref: stub(),
              push: stub(),
              isActive: stub(),
              replace: stub(),
              go: stub(),
              goBack: stub(),
              goForward: stub(),
              setRouteLeaveHook: stub(),
            },
          },
        },
        pathFor: PropTypes.func.isRequired,
        router: PropTypes.object.isRequired,
      }
```

PR #371, opened 2021-11-18
Simply 3938/refactor custom list search updated

oliviawongnyc:
## Description

- Refactored the `CustomListSearch` component to bring it up-to-date with current best practices. 
- Added `CustomListSearchFormContent` and `CustomListSearchAdvancedOptions` to make all components smaller/more manageable.
- Added/edited tests where needed.

## Motivation and Context

- This refactor is part of the larger List Manager refactor ([TAD](https://docs.google.com/document/d/1QhlTh9FLUAWJC9BcmSDgu7m972xyaBoHh1wcgnorq2o/edit)), which is being undertaken to make the code more up-to-date, simpler, and easier to read and maintain.

## How Has This Been Tested?

- I have adjusted and added tests where necessary, and all unit tests are passing.
- I tested search functionality on localhost.

## Checklist:

- [ x ] I have updated the documentation accordingly.
- [ x ] All new and existing tests passed.


tslint.json
```diff
@@ -1,55 +1,37 @@
...
-        "whitespace": [
-            true,
-            "check-branch",
-            "check-decl",
-            "check-operator",
-            "check-separator",
-            "check-type"
-        ]
-    }
+  "extends": ["tslint-react-a11y"],
```
EdwinGuzman, line 2: From 4 to 2 spaces? 👍

src/components/CustomListSearch.tsx
```diff
@@ -1,191 +1,96 @@
...
-  }
+export default function CustomListSearch({
+  search,
+  entryPoints,
+  startingTitle,
+  library,
+  languages,
+}: CustomListSearchProps): JSX.Element {
+  const [searchTerms, setSearchTerms] = React.useState(
+    startingTitle ? startingTitle : ""
```
EdwinGuzman, line 24: Set `startingTitle` on line 19 to `""`. That way, it's the default and you can do `React.useState(startingTitle)` here.

src/components/CustomListSearch.tsx
```diff
@@ -1,191 +1,96 @@
...
+  );
+  const [sortBy, setSortBy] = React.useState(null);
+  const [selectedLanguage, setSelectedLanguage] = React.useState("all");
+  const [selectedEntryPoint, setSelectedEntryPoint] = React.useState<string>(
+    "All"
```
EdwinGuzman, line 29: You might be able to set this to a specific type since you all the possible entrypoint values.

EdwinGuzman: By that I mean an enum or type instead of "All". Maybe something like `Entrypoints.All`. This might not make much sense here so don't worry if it doesn't make too much sense.

src/components/CustomListSearchAdvancedOptions.tsx
```diff
@@ -0,0 +1,100 @@
...
+    <Panel
+      id="advanced-search-options"
+      key="advanced-search-options"
+      style="instruction"
+      headerText="Advanced search options"
+      content={[
+        <fieldset key="sortBy" className="well search-options">
+          <legend>Sort by:</legend>
+          <ul>
+            {Object.keys(sortByOptions).map((x) => (
```
samanthaandrews, line 51: Let's rename `x` to something more descriptive

samanthaandrews (changes requested):
Just one small thing, otherwise this looks great.

src/components/CustomListSearchAdvancedOptions.tsx
```diff
@@ -0,0 +1,100 @@
...
+    Author: "author",
+  };
+
+  const languageFields =
+    library?.settings &&
+    Object.keys(library.settings).filter((x) => x.match(/_collections/));
+
+  const languagesToMap = [].concat(
+    ...languageFields?.map((x) => library.settings[x])
+  );
```
EdwinGuzman, line 35: `.map` returns a new array, so I don't think you need the `[].concat` to begin with. Maybe just `languageFields?.map(....)` is enough. If not, maybe `[...languageFields?.map(...)]`.

src/components/CustomListSearchAdvancedOptions.tsx
```diff
@@ -0,0 +1,100 @@
...
+            <i>
+              Selecting "Title" or "Author" will automatically filter out less
+              relevant results.
+            </i>
+          </p>
+        </fieldset>,
+        <fieldset key="languages" className="well search-options">
+          <legend>Filter by language:</legend>
+          <section>
+            {/* eslint-disable-next-line jsx-a11y/no-onchange */}
```
EdwinGuzman, line 80: Is this line here because of the `aria-selected` attribute on line 85?

src/components/CustomListSearchFormContent.tsx
```diff
@@ -0,0 +1,89 @@
...
+}: CustomListSearchFormContentProps): JSX.Element {
+  return (
+    <>
+      <fieldset key="search">
+        <legend className="visuallyHidden">Search for titles</legend>
+        {entryPoints && entryPoints.length && (
+          <div className="entry-points">
+            <span>Select the entry point to search for:</span>
+            <div className="entry-points-selection">
+              {!entryPoints.includes("All") && (
```
EdwinGuzman, line 42: This was confusing before and a comment would make it clear that we _always_ want to displays the "All" option even if it's not selected on the backend.

src/components/__tests__/CustomListSearchAdvancedOptions-test.tsx
```diff
@@ -0,0 +1,90 @@
...
+    expect(options.at(0).prop("value")).to.equal("all");
+    expect(options.at(0).text()).to.equal("All");
+    expect(options.at(1).prop("value")).to.equal("eng");
+    expect(options.at(1).text()).to.equal("English");
+    expect(options.at(2).prop("value")).to.equal("fre");
+    expect(options.at(2).text()).to.equal("French");
+    expect(options.at(3).prop("value")).to.equal("spa");
+    expect(options.at(3).text()).to.equal("Spanish; Castilian");
+  });
+});
```
EdwinGuzman, line 90: I think this is a great start but you should also test the two `set` functions that are passed instead of what gets displayed.
If the Sinon stub is not enough, there can be a variable `language` that is set to an empty string and you can pass a function for `setSelectedLanguage` that updates the `language` variable.

src/components/__tests__/CustomListSearchFormContent-test.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const searchTerms = "";
+  const entryPoints = ["Book", "Audio"];
+  const sortBy = null;
+  const selectedLanguage = "all";
+  const selectedEntryPoint = "All";
+  beforeEach(() => {
+    setSortBy = stub();
+    setSearchTerms = stub();
+    setSelectedEntryPoint = stub();
+    setSelectedLanguage = stub();
```
EdwinGuzman, line 37: I think you can/should also test these `set` functions that are passed. Even if they don't update a local variable, that they are called appropriately.

src/components/CustomListSearchAdvancedOptions.tsx
oliviawongnyc: I would have thought the same thing, but I tested it out and for this particular data, which is nested, the options without concat lead to a 2-dimensional array, and the code requires a flat array.

src/components/CustomListSearchAdvancedOptions.tsx
oliviawongnyc: No, this line is here because of an outdated rule that says to use onBlur instead of onChange for select tags. [More info here](https://books.google.com/books?id=5fo8EAAAQBAJ&pg=PT383&lpg=PT383&dq=disable+jsx-a11y/no-onchange&source=bl&ots=gRnZmXpyAu&sig=ACfU3U3KhHe0NaEKwVD19l-OYnx4kGNdpQ&hl=en&sa=X&ved=2ahUKEwjsw7_o98L0AhUMZzABHeCcBkMQ6AF6BAgWEAM#v=onepage&q=disable%20jsx-a11y%2Fno-onchange&f=false), if you're curious.

src/components/CustomListSearchFormContent.tsx
oliviawongnyc: Done!

src/components/__tests__/CustomListSearchAdvancedOptions-test.tsx
oliviawongnyc: I added functions to test that the set functions are called onChange. Does this work?

src/components/CustomListSearchAdvancedOptions.tsx
EdwinGuzman: What about just `const languagesToMap = languageFields?.map(....);` ?

EdwinGuzman (approved):
Updated tests look good. I'd err on the side of more tests, even for prop functions, until you can move to RTL.

src/components/CustomListSearch.tsx
```diff
@@ -16,18 +17,18 @@ export interface CustomListSearchProps {
...
   const [sortBy, setSortBy] = React.useState(null);
   const [selectedLanguage, setSelectedLanguage] = React.useState("all");
-  const [selectedEntryPoint, setSelectedEntryPoint] = React.useState<string>(
-    "All"
-  );
+  const [selectedEntryPoint, setSelectedEntryPoint] = React.useState<
+    EntryPoint
+  >("All");
+
+  console.log("entryPoints -->", entryPoints);
```
samanthaandrews: Stray console.log here

samanthaandrews: nvm ignore me looking at outdated code!!

PR #373, opened 2021-12-09
Simply 3969/clear draft bug

oliviawongnyc:
## Description

- Fixed bug wherein if a user added or deleted books from a list, then navigated to a new list without saving, the deletedListEntries and addedListEntries would not be reset to empty.

## Motivation and Context

- This issue would cause the displayed number of books in a list to be wrong.

## How Has This Been Tested?

- Added a new test to CustomListEditor-test to check.
- All existing tests still pass.

## Checklist:

### DIFF
--- a/src/components/CustomListEditor.tsx
+++ b/src/components/CustomListEditor.tsx
@@ -9,7 +9,6 @@ import CustomListEditorHeader from "./CustomListEditorHeader";
 import CustomListEditorBody from "./CustomListEditorBody";
 import { Entry } from "./CustomListBuilder";
 import { getMedium } from "opds-web-client/lib/utils/book";
-import { EntryPoint } from "./CustomListSearch";
 export interface CustomListEditorProps {
   languages: LanguagesData;
   library: LibraryData;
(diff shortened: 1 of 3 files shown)


src/components/CustomListEditor.tsx
```diff
@@ -68,13 +67,21 @@ export default function CustomListEditor({
       setDraftTitle(list.title);
       setDraftEntries(list.books);
       setDraftCollections(listCollections);
+      /**
+       * Whenever the user navigates to a new list,
+       * we want to set the deleted and added buckets back to empty.
+       */
+      setDeletedListEntries([]);
+      setAddedListEntries([]);
```
EdwinGuzman, line 75: This might be better in the return function for this `useEffect`. Since that is the "cleanup" function that is called when the component unmounts, it should clear out these two arrays when the user moves to a different page. Unless if you tried this already and it didn't work?

EdwinGuzman: I just noticed that this is for navigating to a new _list_ and though I read a new _page_. The above suggestion probably won't work in this case since the component is always mounted.

EdwinGuzman (approved):

PR #374, opened 2021-12-16
Simply 2729/special characters in list title

oliviawongnyc:
## Description

- Fixed issue where the "View Details" button did not work in the List Manager.

## Motivation and Context

- When creating a list, if a user clicked "View Details" on any search result or entry, they would get an error. 
- Though seemingly unrelated, I believe this change fixes [SIMPLY-2729](https://jira.nypl.org/browse/SIMPLY-2729).

## How Has This Been Tested?

- I tested this locally with open access collections and non-open access collections.
- All unit tests pass.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### Bugfix
 
-- Fixed issue where the "View Details" button did not work in the List Manager.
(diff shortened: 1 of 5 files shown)


EdwinGuzman (approved):
Hopefully this can be tested soon to find out if it works well in production.

PR #375, opened 2021-12-30
Simply 3974/update save requests

oliviawongnyc:
## Description

- Changed List Manager to save on every action (delete, add), rather than in bulk using the save button ([SIMPLY-3974](https://jira.nypl.org/browse/SIMPLY-3974)).
- This PR includes the code changes only – the tests have been commented out and will be worked on next in order to keep the PR size manageable (note: this branch is being merged into a feature branch that will eventually be merged into main only when tests are complete and passing).

## Motivation and Context

- Simplifies the code so there is no longer a need for so many draft arrays with complicated logic to keep them all up-to-date with every action, which often results in bugs.
- Simplifies the UX so that the user no longer has to actively save.
- See [here](https://docs.google.com/document/d/1d434yFoCjDk8j08iH1rcM8353XOtDeqPTbgxg6Ng9Kk/edit) and [here](https://docs.google.com/document/d/1otB6i34U1WojGCaB0mBgaJNYSVN122cpjjl6msPvIqo/edit) for more details.

## How Has This Been Tested?

- This has been tested manually on localhost, but I have not yet fixed the unit tests (will do that in subsequent PRs).

## Checklist:

### DIFF
--- a/src/components/CustomListEditor.tsx
+++ b/src/components/CustomListEditor.tsx
@@ -69,6 +69,14 @@ export default function CustomListEditor({
 
   const [showSaveError, setShowSaveError] = React.useState<boolean>(false);
 
+  const [responseBodyState, setResponseBodyState] = React.useState<string>("");
+
+  // console.log("responseBodyState -->", responseBodyState);
+
+  React.useEffect(() => {
+    setResponseBodyState(responseBody);
+  }, [responseBody]);
+
   React.useEffect(() => {
     if (list) {
       setDraftTitle(list.title);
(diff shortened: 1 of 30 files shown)


src/components/CustomListEditor.tsx
```diff
@@ -49,88 +49,42 @@ export default function CustomListEditor({
...
-        list.books.forEach((book) => {
-          newAddedListEntries.forEach((entry, j) => {
-            if (book.id === entry.id) {
-              newAddedListEntries.splice(j, 1);
-            }
-          });
-        });
-        setAddedListEntries(newAddedListEntries);
-      }
+      setDraftEntries([]);
```
hokei, line 83: If the list is saving now on every action, what is the purpose of a draft?

src/components/CustomListEditor.tsx
```diff
@@ -49,88 +49,42 @@ export default function CustomListEditor({
...
       setDraftTitle(list.title);
-      setDraftEntries(list.books);
       setDraftCollections(listCollections);
-      /**
-       * Whenever the user navigates to a new list,
-       * we want to set the deleted and added buckets back to empty.
-       */
-      setDeletedListEntries([]);
-      setAddedListEntries([]);
+      setDraftEntries(list.books);
```
EdwinGuzman, line 76: Curious, does the order matter or was there another reason this was moved to the end?

src/components/CustomListInfo.tsx
```diff
@@ -23,29 +23,41 @@ export default function CustomListInfo({
...
 
   return (
     <li key={id} className={isActive ? "active" : ""}>
       <div className="custom-list-info">
         <p>{name}</p>
-        <p>Books in list: {entry_count}</p>
+        {/*
+         * If the number of books in a list is stored in context,
+         * use that number. Otherwise, use entry_count.
+         */}
```
EdwinGuzman, line 54: In what situation is the number of books _not_ stored in context that you have to resort to `entry_count`?

EdwinGuzman:
Does this also cover the list title changing and saving or is that a separate update?

EdwinGuzman (approved):
I'm going to approve so you can move forward but here are some things to consider later on like we discussed:
- the name of the list can be added to the context so you can update it on the sidebar
- RTL tests for this section of the app

src/components/CustomListInfo.tsx
oliviawongnyc: It's only in context if it's a list that is/was being worked on in the current user session.

PR #383, opened 2022-01-25
Simply 3989/configure jest

oliviawongnyc:
## Description

- Installed Jest and added its config file.

## Motivation and Context

- We plan to migrate tests from Enzyme to React Testing Library, and React Testing Library recommends Jest to use as its test runner.

## How Has This Been Tested?

- Existing Enzyme/Mocha tests still run and pass.
- Added an example Jest test which runs and passes.

## Checklist:


jest.config.js
```diff
@@ -0,0 +1,191 @@
...
+  // setupFilesAfterEnv: [],
+
+  // The number of seconds after which a test is considered as slow and reported as such in the results.
+  // slowTestThreshold: 5,
+
+  // A list of paths to snapshot serializer modules Jest should use for snapshot testing
+  // snapshotSerializers: [],
+
+  // The test environment that will be used for testing
+  testEnvironment: "jsdom",
```
EdwinGuzman, line 136: This makes me think that you might not need to set up jsdom in the setup file you need to add in `setupFilesAfterEnv`. But test it with and without the jsdom config in the setup file.

EdwinGuzman (approved):
So this won't be a new release but only merging into `main`, right?

jest.config.js
oliviawongnyc: Will do!

PR #384, opened 2022-01-26
Simply 3945/bug fix

oliviawongnyc:
## Description

- Made save error in List Manager disappear if user navigates to another list.
- Added a call to `/custom_lists` whenever a user switches to a new list.

## Motivation and Context

- The save error was erroneously appearing on a list with a title because it wasn't being turned off when the user switched lists.
- In the situation where a user would create a new list, add books to it, navigate to another list, and then navigate back to the new list, the entry_count would read as 0 (even though there were books in the list). Making a new call to custom_lists when the user switches lists ensures the update entry_count comes through to the frontend.

## How Has This Been Tested?

- This change has been tested locally.
- All unit tests pass.

## Checklist:


hokei:
Are there tests that would fail if this regresses?  It seems like a pretty subtle thing that won't be manually checked every time we make changes.

oliviawongnyc:
> Are there tests that would fail if this regresses? It seems like a pretty subtle thing that won't be manually checked every time we make changes.

Moving tests for this component to RTL in the coming days, and will make sure there is a test that covers this! @hokei

src/components/CustomLists.tsx
```diff
@@ -205,6 +205,7 @@ export class CustomLists extends React.Component<
       nextProps.fetchCustomListDetails &&
       nextProps.identifier !== this.props.identifier
     ) {
+      nextProps.fetchCustomLists();
```
EdwinGuzman, line 208: So one of the bugs was fetching the new or the same list when you moved away?

EdwinGuzman:
Regarding the unit jest test, does this need to be merged in now to QA or can it wait until the unit tests are written?

src/components/CustomLists.tsx
oliviawongnyc: We were already fetching the list at /custom_lists/[list-id/, but we weren't fetching /custom_lists/. For some reason, the correct, updated entry_count is sent to the frontend when you request all lists, but not when you request the single list, as we were doing before. So now we are hitting both endpoints when a user switches lists.

oliviawongnyc:
> Regarding the unit jest test, does this need to be merged in now to QA or can it wait until the unit tests are written?

Sure! Makes sense. @EdwinGuzman

